PR #8, opened 2017-05-07
Emily fallback

Emily-Jiang:


Emily-Jiang:
Unfortunately multiple branches still carry forward the previous commits.

Emily-Jiang:
Please review as I would like to merge in soon.

api/src/main/java/org/eclipse/microprofile/fault/tolerance/cdi/CircuitBreaker.java
```diff
@@ -0,0 +1,69 @@
...
+     * @return The delay unit after the circuit is open
+     */
+
+    TimeUnit delayUnit() default TimeUnit.SECONDS;
+
+    /**
+     *
+     * @return The failure threshold to open the circuit
+     */
+    long failThreshold() default 2;
```
jGauravGupta: failThresholdPercentage should be introduce to activate circuit breaker based on failure percentage instead of failure count.

Emily-Jiang: @jGauravGupta can you explain the usage and use case? I see one the first failure, it will be 100%, so the circuit will open automatically then.

Emily-Jiang:
close this one as I opened a separate pr for each api.

api/src/main/java/org/eclipse/microprofile/fault/tolerance/cdi/CircuitBreaker.java
jGauravGupta: if 50%+ (A%+) of requests in 10 (B) seconds are failures then open the circuit.

Emily-Jiang: This does not work. If the first execution failed, the percentage will be 100%. Do you expect the circuit will be open straightaway?

jGauravGupta: So `requestVolumeThreshold` should also be part of circuit breaker metadata, which is the minimum number of requests in a rolling window.
default value could be 10 or 20 request. Circuit activation logic should run after request reach this this threshold. What do you think ?

In case of `failThreshold` , suppose if `failThreshold = 10`. On Monday requests fail 9 time and after 2-3 days request fail 1 more time then Is it right to Open Circuit ?

Emily-Jiang: I would not want to change the default value to be a large number for no apparent reason. From my experience, normally a small number of retry are used. By the way, default is an initial number and it should be overwritten. 
As for your failThreshold, for a long running microservice, the answer will be yes to your question.

Issue #4, opened 2017-05-05
Create Fallback functionality


johnament:
When an invocation fails, provide the ability to fallback to another invocation that may be more reliable not necessarily the full need.

Emily-Jiang:
Please have a look at the above commit a479883 to see whether this meets your expectation.

mkouba:
The spec should be more clear on the lifecycle and requirements of a `FallbackHandler` implementation, i.e. a new instance of handler is created for each fallback invocation?

johnament:
I don't see the interface mentioned in the spec.

mkouba:
It's part of the API, see `org.eclipse.microprofile.fault.tolerance.inject.Fallback.handler()`. Not sure what the spec actually means in MP.

johnament:
Oh, I think I know what you mean.  Are you expecting that the handler is a CDI bean?

mkouba:
Well, I'm not expecting anythig... just saying that it's undefined ;-). The truth is that using POJO is a bit limiting and using CDI bean might be more powerful... but I have no idea how to define this. Hm, maybe `@Fallback` could be a qualifier and the first matching `FallbackHandler` bean could be used? I'm not sure.

Emily-Jiang:
@mkouba just to confirm: The fallback class should be unmanaged. A new instance should be created for each fallback. Will document this in the javadoc. The usage:
@Fallback(handler=MyFallback.class) or we could even simplify as
@Fallback(MayFallback.class)

In this way, it is much easier to implement and also I have not seen the use cases where need to inject the fallback and want the same instance of the fallback is used. The fallback class is pretty much stateless. 
As for your idea of the first matching FallbackHandler bean could be used, it will easily run into non-portable behaviour or random behaviour though, as we cannot control which one will be used for fallback on.

johnament:
I would prefer to see these as CDI managed.  If there's no scope then they're not beans, but if they're `@Dependent` then a new one per invocation.

mkouba:
Well, I was thinking about something like this:
```java
class MyService {
  @Inject
  OtherService otherService;
  
  @Timeout(3000)
  @Fallback // If handler not set then lookup bean of type FallbackHandler<Long> and qualifier @FallbackFor(MyService.class)
  Long getAmount() {
    return otherService.getAmount() * 2;
  }

  // The fallback bean should be dependent
  // We can detect unsatisfied and ambiguous handlers
  @Produces
[...]

PR #10, opened 2017-05-16
#4 add fallback

Emily-Jiang:

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/fault/tolerance/cdi/Fallback.java
+++ b/api/src/main/java/org/eclipse/microprofile/fault/tolerance/cdi/Fallback.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2017 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.eclipse.microprofile.fault.tolerance.cdi;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Define the Fallback annotation to specify the fallback callable, BiConsumer or BiFuncation
+ * @author <a href="mailto:emijiang@uk.ibm.com">Emily Jiang</a>
+ * @author John Ament
+ *
+ */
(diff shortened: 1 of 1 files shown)


kenfinnigan:
Anyway we can make this type safe so it isn't a String?

Maybe Method instead?

Emily-Jiang:
As for the return type, only primitive type, String, Class, annotation, enumeration are permitted. Reflection will be used to work out the method. By the way, Hystrix uses the similar approach.

kenfinnigan:
Maybe a different approach then?

Fallback being a handler [1] that is then specified in the annotation [2]. Brings type safety to the problem without much effort, and fallbacks can be re-used where it makes sense.

[1] - https://github.com/kenfinnigan/wildfly-swarm-helloworld/blob/master/chaining/hola/src/main/java/org/wildfly/swarm/helloworld/chaining/hola/OlaFallbackHandler.java#L29
[2] - https://github.com/kenfinnigan/wildfly-swarm-helloworld/blob/master/chaining/hola/src/main/java/org/wildfly/swarm/helloworld/chaining/hola/OlaService.java#L40

Emily-Jiang:
Not sure the handler approach is suitable for the annotation. Here is an example of Spring boot with Hystrix:
@RestController
public class NodeAController {
		 
		 @GetMapping("NodeA")
		 @HystrixCommand(fallbackMethod="fallback" )
		 public String makeCallToNodeB() throws Exception{
		 		 return new RestTemplate()
		 		 		 		 .getForEntity(new URI("http://localhost:8081/NodeB"), String.class)
		 		 		 		 .getBody();
		 }
		 
		 public String fallback(){
		 		 return "Red alert! This String is returned by the fallback method";
		 }
}
It is very similar to the current approach.

kenfinnigan:
Can you elaborate on why you think the handler approach isn't suitable, as I thought the examples I provided showed it was?

Isn't it best to take advantage of type safety wherever we can?

I'd also prefer to not just copy how Spring Boot does things without thinking about alternatives

Emily-Jiang:
I did not copying from anywhere and it is quite easy and natural thing to come up. After your comments, I looked further on Hystrix and try to understand your use case. Since Spring boot uses Hystrix, I thought to mention it here about its annotation support. 

As for this fallback annotation, in this case, the return type can either be a string or a class. If you looked at the 1st commit, it was a Class and then I thought it is much easier to use if the return type is String. If we all think returning a class is better. We can spec the fallback method in the class needs to have the exactly same method signature as the main service, which is workable. 

I might have misunderstood your handler idea. The return type cannot be anything more concrete than a class though.

Emily-Jiang:
Maybe using CDI event is a better approach. When max retries reached, circuit open, no others things to do, execution failure should trigger a Fallback event to fire, so that the applications can observe the event and fallback. A PR to follow to demonstrate this idea.

johnament:
When we spoke about fallbacks in the past, they were another implementation of the given class, which matched method signatures.  The string value would only be used if the method name was different.

Emily-Jiang:
See https://github.com/eclipse/microprofile-fault-tolerance/pull/23

Emily-Jiang:
Usage:
`public class MyFTService  {
 
    /**
     * @Todo test the asynchronous behaviour
     * 
     */
    
    public void serviceA() {
        
        serviceB();
    }
    
    @Retry
    @Fallback(handler=MyFallBackHandler.class)
    public void serviceB() {
        //remote call
    }
    
}

public class MyFallBackHandler implements FallbackHandler
{

    @Override
    public void handle(ExecutionContext context)
    {
         ...
        System.out.println(context.getMethod());
        System.out.println(context.getParameters());

    }

}

`

Emily-Jiang:
see #24

PR #20, opened 2017-06-01
microprofile-fault-tolerance-18 Add ElementType.TYPE to annotations @Target

### LINKED ISSUE #18: Add ElementType.TYPE to annotations @Target
Not having these prevents creation of CDI interceptors in implementations.

antoinesd:
fixes #18

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/fault/tolerance/inject/CircuitBreaker.java
+++ b/api/src/main/java/org/eclipse/microprofile/fault/tolerance/inject/CircuitBreaker.java
@@ -32,7 +32,7 @@ import java.time.temporal.ChronoUnit;
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
-@Target({ ElementType.METHOD })
+@Target({ ElementType.METHOD, ElementType.TYPE })
 public @interface CircuitBreaker {
 
     /**
@@ -53,10 +53,10 @@ public @interface CircuitBreaker {
      */
 
     ChronoUnit delayUnit() default ChronoUnit.MILLIS;
-    
+
 
     /**
-     * The number of consecutive requests in a rolling window 
+     * The number of consecutive requests in a rolling window
      * that will trip the circuit.
      * @return the number of the consecutive requests in a rolling window
      *
@@ -64,7 +64,7 @@ public @interface CircuitBreaker {
     long requestVolumeThreshold() default 20;
     /**
      * The failure threshold to trigger the circuit to open.
-     * e.g. if the requestVolumeThreshold is 20 and failureRation is .50, 
+     * e.g. if the requestVolumeThreshold is 20 and failureRation is .50,
      * more than 10 failures in 20 consecutive requests will trigger
      * the circuit to open.
      * @return The failure threshold to open the circuit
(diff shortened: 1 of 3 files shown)


Emily-Jiang:
We implemented the apis and it seems to be ok.

antoinesd:
So you had to go thru all the SPI to define your interceptor? Or did you do a non-CDI implementation?

kenfinnigan:
Is it a problem to add Type support for the annotations?

We need to allow for implementations to achieve the same result in different ways.

kenfinnigan:
@antoinesd did you sign the commit with -s?

It says the IP validation failed

Emily-Jiang:
@antoinesd we created an interceptor binding in the implementation and add the interceptor via the cdi spi. 
@kenfinnigan originally I did not add the Type because I think it may not be realistic to apply the policy on the class, which implies all methods within the class. If you guys think there is use case, we can add it.

antoinesd:
@Emily-Jiang that's what I wanted to avoid: defining an API to recreate it in your implementation, could make you feel that you did wrong in the first place.
I agree that, applying these policy on all the class may be rare, but why preventing it?
@kenfinnigan I lost the auto signing from my IDE when updating gpg. It should be ok now

Emily-Jiang:
@antoinesd I see what you meant. We created one interceptor binding, which is not an issue. If you create an interceptor for each, it will be cumbersome. By the way, the annotations can be used together. If we allow the annotations on both class and method level, we need to define which one take precedence. I would say the method level takes precedence. Trying to keep it simpilier in the first release:o.

Emily-Jiang:
@antoinesd @kenfinnigan Before we merge this one, can we reach agreement on whether to support this use cases:
e.g.
`
@Retry
public class myBean{

public void myService() {
myMethodA();
}
@CircuitBreaker
public void myMetholdA() {
//calling a remote service
}
@Retry(maxRetries=2)
@CircuitBreaker
public void myMethodB() {

}

public void myMethodC() {

}
}
`
Basically, all methods will have Retry policy applied. In the scenario of myService -> serviceA, both myService and ServiceA have the policy applied. Each method might have other policy applied and might override the policy. There are a number of combinations. Shall we get the method level support out first and gather feedback on how useful to apply on the class level as it can be very complicated? However, if this is a showstopper for you, we can discuss further to see how to best proceed. Thoughts?

antoinesd:
Yes. By doing this we stick to interceptor behaviour and it's the intuitive approach.

Emily-Jiang:
@antoinesd For the sake of argument, I don't think the interceptor binding requiring Type is a valid argument here. According to the interceptor spec:
"An interceptor binding type is a Java annotation defined as Retention(RUNTIME). Typically an interceptor binding is defined as Target({TYPE, METHOD, CONSTRUCTOR}) or any subset of valid target types.". The interceptor Binding is not required to define TYPE. It is purely valid to define the method level interceptors, which means Target({METHOD}).

Emily-Jiang:
Let's allow the annotation to be supported on the class-level and document the user cases. If we find it is confusing, we can always go back to Method only, as I don't think Target(Type) should play a significant blockage if we remove it.

PR #22, opened 2017-06-01
#4 add event model to enable the exception recovery e.g. fallback

Emily-Jiang:

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/fault/tolerance/inject/ExecutionContext.java
+++ b/api/src/main/java/org/eclipse/microprofile/fault/tolerance/inject/ExecutionContext.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright (c) 2017 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


kenfinnigan:
Do you have an example of how this would be used by a developer?

Emily-Jiang:
Here is the code snippet:
`public class TestFallback
{

    
    public void serviceA() {
        serviceB();
    }
    // An event will be fired if the circuit is open or max retry exceeded
    @CircuitBreaker
    public void serviceB() {
        
        //call a remote service 
    }
    //observe the event
    public void fallbackOnServiceBFailure(@Observes @FailureMethod("serviceB") ExecutionFailureEvent events){
        
        //fallback method body
    }
}`

Emily-Jiang:
Had a further thought on this. I think this event model does not address the fallback aspect on Hystrix and failsafe as that fallback approach is kind of treating fallback as part of mainservice invocation. This event model separates main service from fallback. I think it is still useful for logging and metrix gathering purpose. Thoughts?

kenfinnigan:
After looking at the usage, it definitely looks like something that would be an implementation detail as opposed to developer created.

Right now I'm not sure what's provided would be beneficial for metrics/logging. It would require more investigation

Emily-Jiang:
I think the event model is useful so that the status of circuit breaker etc can be logged and observed at a minimum. It is completely optional for the app developers.

kenfinnigan:
I don't think there's a use case for a developer needing this ability, so I would consider this an implementation detail rather than something that's part of an exposed API

heiko-braun:
I think fallback methods might require access to the invocation context. I.e what about possible parameters passed into `serviceA()` and `serviceB()`?

heiko-braun:
Ah, I see it's part of `ExecutionFailureEvent`.

PR #23, opened 2017-06-02
Fallback 

Emily-Jiang:


api/src/main/java/org/eclipse/microprofile/fault/tolerance/inject/FallbackHandler.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ */
+public interface FallbackHandler
+{
+    /**
+     * Handle the previous calling failure and then call alternative methods
+     * or perform any alternative operations.
+     * @param context the execution context 
+     */
+    void handle(ExecutionContext context);
```
kenfinnigan: Should this return T and then `FallbackHandler` becomes `FallbackHandler<T>`?

Otherwise, how is a response returned from a fallback?

kenfinnigan:
For now I would prefer to remove `ExecutionFailureEvent` and `FailureMethod` to keep it simple

api/src/main/java/org/eclipse/microprofile/fault/tolerance/inject/ExecutionContext.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+public interface ExecutionContext {
+
+     /**
+     * Returns the method being executed
+     * 
+     * @return the method
+     */
+    public Method getMethod();
```
heiko-braun: Is it safe to pass the Method instance into user level code? Would a method name (String) suffice as well?

api/src/main/java/org/eclipse/microprofile/fault/tolerance/inject/ExecutionFailureEvent.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+package org.eclipse.microprofile.fault.tolerance.inject;
+
+/**
+ * The event payload on execution failure.
+ * 
+ * @author <a href="mailto:emijiang@uk.ibm.com">Emily Jiang</a>
+ */
+
+public interface ExecutionFailureEvent {
```
heiko-braun: What are the use cases for exposing `retryExceeded` and `circuitOpen` here?

api/src/main/java/org/eclipse/microprofile/fault/tolerance/inject/FallbackHandler.java
Emily-Jiang: +1 @kenfinnigan !

api/src/main/java/org/eclipse/microprofile/fault/tolerance/inject/ExecutionContext.java
Emily-Jiang: What is the issue with passing method? Method name is not sufficient as the Method can be used to find out the declaring class etc.

api/src/main/java/org/eclipse/microprofile/fault/tolerance/inject/ExecutionFailureEvent.java
Emily-Jiang: A user can observe these events and log the info etc.

Emily-Jiang:
I'll create another PR to just commit the fallback and leave the event model for further discussion.

Emily-Jiang:
See #24

api/src/main/java/org/eclipse/microprofile/fault/tolerance/inject/ExecutionContext.java
heiko-braun: That's exactly what I am concerned about. Why would user code need access all the declaring class or even be able to re-invoke that method? Wouldn't it be safer to prevent access of this kind?

heiko-braun:
@Emily-Jiang let's continue on the other PR

api/src/main/java/org/eclipse/microprofile/fault/tolerance/inject/ExecutionContext.java
Emily-Jiang: The fallback is completely in user's hand. They should be able to do anything they want. If they want to revoke the previous method, they have to get hold of the instance. If they choose to do that with a lot of effort, there might be a reason for it and they should know what they are doing.

Issue #13, opened 2017-05-17
Create timeout annotation


Emily-Jiang:

mkouba:
I think the `TimeOut` annotation would be easier to read/use if you rename:
* the annotation to `Timeout`
* `TimeOut.timeOut()` to `TimeOut.value()`
* `TimeOut.timeOutUnit()` to `TimeOut.unit()`

E.g. 
```java
@Timeout(500) // 500 ms
void foo() { }
@Timeout(value=5, unit=ChronoUnit.SECONDS) // 5 seconds
void bar() { }
```

kenfinnigan:
+1

Emily-Jiang:
@mkouba sounds good. Please provide a PR.

PR #31, opened 2017-06-13
Make TimeOut annotation easier to read and use

mkouba:
- rename annotation to Timeout
- rename timeOut() to value()
- rename timeOutUnit() to unit()

See also https://github.com/eclipse/microprofile-fault-tolerance/issues/13#issuecomment-308125929


kenfinnigan:
@mkouba have you signed an ECA?

Issue #27, opened 2017-06-07
BulkHead


Emily-Jiang:
Provide a way to configure the maximum number of threads accessing a resource.

johnament:
This is a duplicate of #5

PR #47, opened 2017-06-30
#27 add bulkhead

Emily-Jiang:


johnament:
We need to define scoping within bulkhead.  E.g. if I have a request scoped component, am I counting the invocations across all threads (to different instances) of that component? Or is it only when its the same instance across multiple threads?

Emily-Jiang:
I think we should spec counting the multiple threads per instance. There is no bulkhead in failsafe. Not sure how Hystrix counts the threads. I thought it might just count per instance not per class. Perhaps @antoinesd can confirm.

Emily-Jiang:
I'm going to merge in. Please open issue if you have more thoughts on this.

johnament:
I'll start to work on the spec tomorrow.

PR #57, opened 2017-07-04
Clarify the behavior of FallbackHandler when the return types don't m…

### LINKED ISSUE #51: Fallbacks should fail deployment when return types don't match
The following TCK assertion is bad

```
    public void testFallbackFailure() {
        try {
            fallbackClient.serviceB();
            Assert.fail("serviceB should throw an IllegalArgumentException in testFallbackFailure");
        }
        catch(RuntimeException ex) {
            Assert.assertTrue(ex instanceof IllegalArgumentException, "serviceB should throw an IllegalArgumentException in testFallbackFailure");
        }
        Assert.assertEquals(fallbackClient.getCounterForInvokingCountService(), 5, "The max number of execution should be 5");

    }
```

IMHO, we should fail deployment if the fallback handler doesn't match the return type of the method.  Not throw an illegal argument exception.

johnament:
…atch.

This closes #51

Signed-off-by: John D. Ament <john.d.ament@gmail.com>

### DIFF
--- a/spec/src/main/asciidoc/fallback.asciidoc
+++ b/spec/src/main/asciidoc/fallback.asciidoc
@@ -16,38 +16,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 // Contributors:
-// Emily Jiang
+// John D. Ament
 
-= Microprofile Fault Tolerance
-:author: Emily Jiang
-:email: emijiang@uk.ibm.com
-:revnumber: 0.1
-:revdate: 2017-05-07
-:revremark: Specification
-:version-label!:
-:sectanchors:
-:doctype: book
-:license: Apache License v2.0
-:source-highlighter: coderay
-:toc: left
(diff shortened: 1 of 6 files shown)


spec/src/main/asciidoc/fallback.asciidoc
```diff
@@ -0,0 +1,33 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// Contributors:
+// John D. Ament
+
+[[fallback]]
+== Fallback
+Fallbacks are invoked once a Retry or CircuitBreaker has failed enough times.
```
Emily-Jiang: This is not automatically invoked. I'll sat Fallback can be configured to be invoked once ...

PR #61, opened 2017-07-11
#microprofile-fault-tolerance-40 Introduce an annotation to configure default operation handler

### LINKED ISSUE #40: Introduce an annotation to configure default operation handler
API should introduce an annotation to configure default (i.e non in fallback) operation handler.
This should be done for multiple reason:

* To provide same level of configuration for default operation than we have for fallback
* To allow user to create handler class having implementation specific feature
* Mark method for Fault Tolerance handling with default behaviour (default timeout, fallback, etc...)

This annotation would work like `@Fallback`, we only may have to rename `FallbackHandler` interface to `OperationHandler`

I suggest the following names for this annotation:

* `@FaultToleranceOperation`
* `@FaultToleranceHandler`
* `@FaultToleranceConfiguration` (if we'd like to had impl specific config in a Map member)

I chose the term 'operation' to avoid 'command' which is too  much "Hystrix-ish", but of course another term could be chosen here.

antoinesd:
fixes #40

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/faulttolerance/FaultToleranceDefault.java
+++ b/api/src/main/java/org/eclipse/microprofile/faulttolerance/FaultToleranceDefault.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) 2017 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/eclipse/microprofile/faulttolerance/FaultToleranceDefault.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.METHOD, ElementType.TYPE })
+@Inherited
+public @interface FaultToleranceDefault {
+
+    static final class DEFAULT {}
+
+    Class<?> value() default DEFAULT.class;
```
antoinesd: Not sure to understand you remark. This annotation proposal has 2 goal:

* activate Fault tolerance operation with default behaviour (let to impl decision) when other FT annotations don't make sense on it. It then differentiates a method without any annotation that shouldn't be "fault toleranced" from a method that we want "fault toleranced"

* Optionally provide a class specific to impl allowing overriding standard fault tolerance code. You can see it as providing specific config for the impl.

So it's not a suffix it's a way to override or send specific configuration to the impl.

Emily-Jiang:
After today's hangout, we decided to close this and revisit at a later date if such a user case arises. Thanks @antoinesd !

PR #67, opened 2017-07-13
Add class scope timeout tck tests

neilgsyoung:


tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/timeout/clientserver/DefaultTimeoutClient.java
```diff
@@ -0,0 +1,65 @@
...
+        catch (InterruptedException e) {
+            //expected
+        }
+        return null;
+    }
+    
+    /**
+     * serviceB specifies a an overriding Timeout longer than the default, at 2 seconds
+     */
+    @Timeout(value = 2000)
```
Emily-Jiang: you can just do @Timeout(2000)

PR #69, opened 2017-07-13
#4 provide a way to fallback to a method on the same instance as the …

Emily-Jiang:
…original method

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
+++ b/api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
@@ -29,6 +29,15 @@ import java.lang.annotation.Target;
  * The fallback annotation to define the fallback handler class so that
  * a failure can be handled properly.
  * 
+ * <li>If value is specified, use {@link FallbackHandler#handle()} on the specified handler to execute the fallback.
+ * <li>If fallbackMethod is specified, invoke the method specified by the fallbackMethod on the same class.
+ * <li>If both are specified, try to fallback to fallbackMethod first, 
+ * if it fails then call {@link FallbackHandler#handle()}
+ * 
+ * 
+ * 
+ *  
+ * 
  * @author <a href="mailto:emijiang@uk.ibm.com">Emily Jiang</a>
  * 
  */
@@ -40,8 +49,19 @@ public @interface Fallback {
 
     /**
      * Specify the fallback class to be used. An new instance of the fallback class
-     * is returned. The instance is unmanaged.
+     * is returned. The instance is unmanaged. The type parameter of the fallback class must be assignable to the
+     * return type of the annotated method. Otherwise, the {@link javax.enterprise.inject.spi.DeploymentException} must be thrown.
+     * 
      * @return the fallback class
      */
     Class<? extends FallbackHandler<?>> value();
+    /**
+    * Specify the method name to be fallbacked to. This method belongs
+    * to the same class as the method to fallback.
+    * The method must be either zero argument or have the exactly same arguments as the method being annotated.
+    * The method return type must be assignable to the return type of the method annotated. 
+    * Otherwise, the {@link javax.enterprise.inject.spi.DeploymentException} must be thrown.
+    * @return the local method to fallback to
+    */
+    String fallbackMethod() default "";
 }
--- a/api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
+++ b/api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
@@ -44,4 +44,10 @@ public @interface Fallback {
      * @return the fallback class
      */
     Class<? extends FallbackHandler<?>> value();
+    /**
+     * Specify the method name to be fallbacked to. This method belongs
+     * to the same class as the method to fallback.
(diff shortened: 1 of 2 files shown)


johnament:
The second commit here seems to have nothing to do with `Fallback`.

api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
```diff
@@ -44,4 +44,10 @@
      * @return the fallback class
      */
     Class<? extends FallbackHandler<?>> value();
+    /**
+     * Specify the method name to be fallbacked to. This method belongs
+     * to the same class as the method to fallback.
+     * @return the local method to fallback to
+     */
+    String fallbackMethod() default "";
```
johnament: Its confusing to use both `FallbackHandler` and `fallbackMethod` as the usage would imply to use that method on the handler implementation.

While I see it demonstrated quite often to point to another method on the same class, I don't see any real world scenarios where this would be the case.  I would prefer to use these two in conjunction:

- if FallbackHandler is specified, then use `fallbackMethod` on the handler to execute the fallback
- If no handler specified, then its assumed to be this class
- If handler is specified but no method name specified, then assume there is a method on the handler with the same name, args and return type.

Emily-Jiang: If FallbackHandler is specified, it will have only one method handle(ExecutionContext context). There will be any method called the name specified by fallbackMethod.
In light of your concern, we can say :
-  If FallbackHandler is specified, then use handle on the handler to execute the fallback 
-  If no handler specified, then invoke the method specified by the fallbackMethod on the same class
- If both specified, try to fallback to fallbackMethod first, if it fails then call FallbackHandler.handle. or we just thrown DeploymentException?
By the way, for validating fallbackMethod, its argument could be either no arg or exactly same as the original method.

Emily-Jiang:
The 2nd commit was not supposed to be on this PR. will fix it.

api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
```diff
@@ -40,8 +49,19 @@
...
+     * is returned. The instance is unmanaged. The type parameter of the fallback class must be assignable to the
+     * return type of the annotated method. Otherwise, the {@link javax.enterprise.inject.spi.DeploymentException} must be thrown.
+     * 
      * @return the fallback class
      */
     Class<? extends FallbackHandler<?>> value();
+    /**
+    * Specify the method name to be fallbacked to. This method belongs
+    * to the same class as the method to fallback.
+    * The method must be either zero argument or have the exactly same arguments as the method being annotated.
```
johnament: Why would zero argument be useful?  Should we also support `ExecutionContext` as a parameter?

api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
```diff
@@ -40,8 +49,19 @@
...
+    * The method return type must be assignable to the return type of the method annotated. 
+    * Otherwise, the {@link javax.enterprise.inject.spi.DeploymentException} must be thrown.
```
johnament: Now that I'm seeing more of these, I think we need to create a `FaultToleranceDeploymentException` that extends `DeploymentException`

Emily-Jiang: +1 on creating FaultToleranceDeploymentException

api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
Emily-Jiang: It is useful if the parameters are not needed and this simplies the implementation of the fallback. I think ExecutionContext is not useful in this situation as it requires unwrapping of the parameters. Application developers would rather use the exactly same parameters or zero arg to do the fallback as it is on the same class.

Emily-Jiang:
@johnament please see my comments and respond asap.

api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
```diff
@@ -40,8 +49,19 @@
 
     /**
      * Specify the fallback class to be used. An new instance of the fallback class
-     * is returned. The instance is unmanaged.
+     * is returned. The instance is unmanaged. The type parameter of the fallback class must be assignable to the
+     * return type of the annotated method. Otherwise, the {@link javax.enterprise.inject.spi.DeploymentException} must be thrown.
+     * 
      * @return the fallback class
      */
     Class<? extends FallbackHandler<?>> value();
```
antoinesd: You have to introduce a dummy Noop class to define a default value for `value` in order to allow user to not specify value for `value`

Emily-Jiang: +1. As discussed on the phone, please direct do a PR to get this rectified.

Emily-Jiang:
After talking with @johnament and @antoinesd , I'll merge this PR and then Antoine will create a PR to add a noop default for the fallback class.

PR #80, opened 2017-07-20
#4 address review comments on fallback

Emily-Jiang:


Emily-Jiang:
Fix the fallback default based on Anotine's review comments.

Issue #111, opened 2017-08-23
Fallback should not be placed on class level


Emily-Jiang:
Fallback annotation is to provide an fallback method for method invocation. It does not make sense to be on class level. There won't be a universal method for the whole class methods to fallback to.

PR #112, opened 2017-08-23
#111 make sure Fallback can only be on method level

Emily-Jiang:


Emily-Jiang:
I discussed with @antoinesd and Graurav and we agreed this is a good idea.

PR #113, opened 2017-08-24
Suggesting alignment of maven artifact groups

OndrejM:
According to the [Contributing guidelines in the Wiki](https://wiki.eclipse.org/MicroProfile/ContributingGuidelines#Deliverable_group_and_artifact_naming_convention) and also according to the practice used with MP Config, I suggest that maven group of all artifacts is changed to `org.eclipse.microprofile.fault-tolerance`, even for the parent. Using `fault-tolerance` seems better to me than `fault.tolerance`, because it implies that both words together form a single name of a feature, rather than a nested feaure.

I also fixed the version of parent to 1.0-SNAPSHOT to be in line with the version of other parents to make it easier to do a release with maven release plugin later.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -19,12 +19,12 @@
 
     <parent>
         <!-- This is just for now and will not work if the API has a separate release cycle than the rest. -->
-        <groupId>org.eclipse.microprofile.fault.tolerance.parent</groupId>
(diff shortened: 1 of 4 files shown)


Emily-Jiang:
Thanks Ondrej! You are ahead of me. I meant to fix the group id tomorrow:o.

Issue #108, opened 2017-08-23
Explore how we should integrate MP config spec


antoinesd:
Right now we don't use config spec while some elements in the spec are related to configuration (i.e. environment variable). We should check that we are not doing config spec job with those IMO.

Emily-Jiang:
https://groups.google.com/forum/#!topic/microprofile/30hr2ldFAYs

PR #118, opened 2017-08-30
#108 add hard dependency on config

Emily-Jiang:


Emily-Jiang:
I discussed this on the hangout today with Antoine and Grauve today. I'm going to merge in tomorrow if no other objections.

PR #138, opened 2017-09-11
Fix wrong assumption in Bulkhead Future tests

hutchig:
Signed-off-by: Gordon Hutchison <Gordon.Hutchison@gmail.com>
The tests were throwing errors if an application level Future had methods other than the
gets called - this was wrong.


hutchig:
Emily commit 88b1b36, which is now part of the PR, takes account of your feedback.

johnament:
Can you explain a bit why these tests were removed?

hutchig:
Apologies, I missed this question. Here is the answer: I removed these tests as I wrote them thinking that the cancel methods on the customers Future class would/should not be called, once I realized that with was my mistaken interpretation, I was anxious that they would sit their causing problems for other folks' implementations. I am working on some more complex CompletableFuture based replacements but did not get them finished in time for 1.0 and becoming aware that we were trying to chill for the 1.0 release, with the knowledge that the bad tests were sitting there and would cause trouble I deleted them and created the PR.

PR #147, opened 2017-09-12
Update the default maxDuration on Retry

Emily-Jiang:


Emily-Jiang:
thank you @johnament !

Issue #165, opened 2017-09-28
tck on a jee container fail with NPE


McFoggy:
While implementing FT as a standard JEE CDI extension, I am facing issues running the TCK.  
Given the attached simple demo project (tck+arquillian+wildfly), running `mvn test` ends with:

```
Results :

Failed tests:
  RetryTest>Arquillian.run:164->testClassLevelRetryMaxDuration:159 NullPointer
  RetryTest>Arquillian.run:164->testClassLevelRetryMaxDurationSeconds:184 NullPointer
  RetryTest>Arquillian.run:164->testClassLevelRetryMaxRetries:138 NullPointer
  RetryTest>Arquillian.run:164->testRetryMaxDuration:91 NullPointer
  RetryTest>Arquillian.run:164->testRetryMaxDurationSeconds:106 NullPointer
  RetryTest>Arquillian.run:164->testRetryMaxRetries:76 NullPointer
  RetryTest>Arquillian.run:164->testRetryWithDelay:119 NullPointer

Tests run: 7, Failures: 7, Errors: 0, Skipped: 0
```

Not being an expert of [testng](http://testng.org), I have read on [SO](https://stackoverflow.com/questions/10230710/beforeclass-annotations-invoking-methods-twice-when-using-arquillian-on-remote#answer-17036231) that injection occured both client and server side for deployed tests. Also [testng](http://testng.org) documentation states that injection is handled directly by testng or guice, which might collide with CDI+Arquillian.

How can the tck be run using a jee container?

**Example maven project:** [jee-mpft.zip](https://github.com/eclipse/microprofile-fault-tolerance/files/1340135/jee-mpft.zip)

hutchig:
[...]

PR #166, opened 2017-09-28
[Do not merge] open tck for extensions, use junit instead of testng

McFoggy:
As expressed in #165 I want to be able to run mpft tck (arquillian based) on a jee standard platform.

This PR introduces 2 changes allowing the tck to be run at least on wildfly:
- junit instead of testng
- extension point to ShrinkWrap web archives under tests in the tck

The extension point would allow third parties to enhance the archives under tests with additionnal stuff including, but not limited to, their implementation _(CDI extensions, interceptors, ...)_.

This PR is provided as a discussion entry point for the 2 topics above. The junit move, as provided in the PR, is only a transitional step. I didn't wanted upfront to reimplement fully the tests (without knowing if you would accept that) using junit so I created the `org.testng.Assert` facade class.

With this PR, you can run a demo project running tck on jee (wildfly) using arquillian: https://github.com/McFoggy/mpft-jee-tck.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
     <artifactId>microprofile-fault-tolerance-parent</artifactId>
     <version>1.1-SNAPSHOT</version>
     <packaging>pom</packaging>
-    
+
     <name>MicroProfile Fault Tolerance</name>
     <description>Eclipse MicroProfile Fault Tolerance Feature</description>
 
@@ -275,6 +275,7 @@
                 </executions>
                 <configuration>
                     <excludes>
+                        <exclude>src/main/resources/META-INF/services/org.eclipse.microprofile.fault.tolerance.tck.TckArchiveProvider</exclude>
                         <exclude>.travis.yml.*</exclude>
                         <exclude>bnd.bnd</exclude>
                         <exclude>*.log</exclude>
@@ -287,7 +288,7 @@
 
         </plugins>
     </build>
-    
+
     <profiles>
         <profile>
             <id>release</id>
--- a/tck/pom.xml
+++ b/tck/pom.xml
@@ -78,14 +78,14 @@
         </dependency>
 
         <dependency>
-            <groupId>org.testng</groupId>
-            <artifactId>testng</artifactId>
-            <version>6.9.9</version>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.12</version>
             <scope>compile</scope>
         </dependency>
         <dependency>
-            <groupId>org.jboss.arquillian.testng</groupId>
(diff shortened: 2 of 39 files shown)


hutchig:
The extension mechanism added adds to every tests deploy method:
+   return TckAdditions.decorate(war);
where decorate uses a ServiceLoader to populate a pipeline of shrinkwrap ers

+    public static WebArchive decorate(WebArchive web) {
+        ServiceLoader<TckArchiveProvider> providers = ServiceLoader.load(TckArchiveProvider.class);
+        providers.forEach(p -> web.addAsLibrary(p.additional()));
+        return web;
+    }

where an example additional is:

+    public Archive<?> additional() {
+        JavaArchive archive = ShrinkWrap.create(JavaArchive.class);
+        archive.addClass(Assert.class);
+        return archive;
+    }

What comes to mind is that Arquillian already has a mechanism for Loadable extension
to get their shrinkwrap hands on the archives: 

For example:

public class ArquillianLoadableExtension implements LoadableExtension {
	public void register(ExtensionBuilder extensionBuilder) {
		extensionBuilder.service(ApplicationArchiveProcessor.class, ArchiveProcessor.class);
	}
}

public class ArchiveProcessor implements ApplicationArchiveProcessor {
   public void process(Archive<?> applicationArchive, TestClass testClass) {
		if (applicationArchive instanceof WebArchive) {
			File hamcrest = new File("../lib/hamcrest-all-1.3.jar");
			((WebArchive) applicationArchive).addAsLibraries(hamcrest);
		}
	}
}

The extension loader:
https://github.com/arquillian/arquillian-core/blob/master/core/impl-base/src/main/java/org/jboss/arquillian/core/impl/loadable/LoadableExtensionLoader.java
is already quite mature and has been used by multiple projects as noted towards the end of:
https://github.com/eclipse/microprofile-config/blob/master/tck/running_the_tck.asciidoc
( as noted by @struberg ) so I would be hesitant of us coming up with a new mechanism unless it
does something better. Perhaps there is a technical advantage I have missed or perhaps we need to do a better job of discussing how to do it the 'other' way?

As far as the use of junit rather than testng goes, that was basically a historical thing coming from
everyone just copying what was there originally in the first few tests in MP Config which I THINK came from  @struberg too. :-) In this context, they are about equivalent, and junit is slightly more ubiquitous so I would +1 it  - it would have to be polled though to see what the consensus it :-)

McFoggy:
@hutchig I wasn't aware of those already existing Arquillian extension mechanism. Of course for me it is a +1 to use such existing mechanisms.

McFoggy:
closing as #165 is finally resolved by moving to latest version of arquillian universe.

Issue #174, opened 2017-10-08
Wrong exceptions used in TCK


johnament:
In the various javadocs for fault tolerance, configuration issues are defined by `FaultToleranceDefinitionException`.  However, the TCK in multiple places expects `javax.enterprise.inject.spi.DefinitionException` to be thrown.

The two exceptions are not compatible with one another, the fault tolerance exception does not extend the CDI one.  

Either, `FaultToleranceDefinitionException` extends `DefinitionException` or the TCK needs to be updated to reflect the proper exception being thrown.

Emily-Jiang:
will update TCK to fine tune this. Since the implementation is to use CDI, All validation exception will eventually wrap under `javax.enterprise.inject.spi.DefinitionException.`

johnament:
I just took a look at how Weld is using `DefinitionException`, under no situation is it being thrown as a part of a deployment validation problem, so I believe the only valid outcome is to look for `DeploymentException`.  You can see how we're doing this properly in the config project - [MissingValueOnInstanceInjectionTest.java#L40](https://github.com/eclipse/microprofile-config/blob/master/tck/src/main/java/org/eclipse/microprofile/config/tck/broken/MissingValueOnInstanceInjectionTest.java#L40)

Emily-Jiang:
@johnament On the call, we agreed the current tests are ok and the DefinitionException is the right. I'm going to close this one. Please let me know if you have any objections.

smillidge:
The test should actually just catch Exception. We run our tests with a remote Arquillian Server profile and these tests assume an embedded profile.

smillidge:
@johnament the config tests have the same problem they assume an embedded profile.

Emily-Jiang:
I believe this issue has been solved via the hangout and the new version of the Arquilliant upgrade.

PR #175, opened 2017-10-11
#174 update expected exception in tck to better align with spec

Emily-Jiang:

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/illegalConfig/IncompatibleFallbackMethodTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/illegalConfig/IncompatibleFallbackMethodTest.java
@@ -19,9 +19,9 @@
  *******************************************************************************/
 package org.eclipse.microprofile.fault.tolerance.tck.illegalConfig;
 
-import javax.enterprise.inject.spi.DefinitionException;
 import javax.inject.Inject;
 
+import org.eclipse.microprofile.faulttolerance.exceptions.FaultToleranceDefinitionException;
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.container.test.api.ShouldThrowException;
 import org.jboss.arquillian.testng.Arquillian;
@@ -37,7 +37,7 @@ public class IncompatibleFallbackMethodTest extends Arquillian {
     FallbackMethodClient fallbackMethodClient;
 
     @Deployment
-    @ShouldThrowException(DefinitionException.class)
+    @ShouldThrowException(FaultToleranceDefinitionException.class)
     public static WebArchive deployAnotherApp() {
         JavaArchive testJar = ShrinkWrap
                 .create(JavaArchive.class, "ftInvalid.jar")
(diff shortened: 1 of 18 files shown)


Pandrex247:
I think these tests need to be made more permissive, as they're assuming that an implementation isn't passing the exception to something else.
These tests are mandating that the final exception returned be `FaultToleranceDefinitionException`, which seems extremely strict - it's not checking any of the causes of the received exception at all.

Emily-Jiang:
No. It does not mandate the final exception is `FaultToleranceDefinitionException`. It asserts this exception is in the chain. In your tck runner, you can check this exception on the chain. 

Have that said, as I mentioned on the call with @johnament and @antoinesd, it is better to catch CDI `DefinitionException`. In this way, we are certain all app servers throws the same top level exception as well instead of wrapping on a different exception. Please see #174 for more details.

I will close this PR if no objections.

Pandrex247:
Hmm, if that's the case I'll have to debug Arquillian to see what it's getting back.
Checking our log file, we're throwing a deployment exception with a cause of `FaultToleranceDefinitionException`, yet Arquillian claims we're failing these tests.

Pandrex247:
So the reason it's failing for us is because we're testing using a remote Arquillian container, which is using Rest to send the deploy command and receive the result. Since it's using Rest, it's just getting a failure response, not an exception, and so reading the response code and throwing a generic container exception to mark that something went wrong (Jersey's `ContainerException` in our case).

So the argument for making the test more permissive would be to accommodate remote containers using Rest.

smillidge:
These deployment test don't work on remote arquillian profiles the Exception should be relaxed. As @Pandrex247 states we test using a remote arquillian profile which deploys via REST to a standalone server.

Emily-Jiang:
Can you explain a bit more @smillidge ? You can code up the arquillian runner to search for a particular exception on the chain not the final exception, which is most of us doing.

smillidge:
I'm not an arquillian expert but our Arquillian plugin uses the administration REST interface to deploy to a remote Payara Server. As the application fails to deploy (passing the test) we don't receive a full exception chain just a REST response saying the application failed to deploy.

Emily-Jiang:
@smillidge you should be able to view the exception and find the expected exception from the stack. @hutchig has done a PR to arquillian for such a functionality.

Emily-Jiang:
also @johnament and @McFoggy did further improvement in Arquillian to make it much easier for negative tests. @McFoggy will consume the latest version of Arquillian soon.

smillidge:
I don't understand Arquillian enough to know which side the exception check is done on in the remote case. If it is on the server side then perhaps the exception can be extracted from the stack. If on the client side I don't see how this can be done given it is a REST call. I suspect it needs more investigation.

McFoggy:
@Emily-Jiang let's be clear ; there are 2 issues for me behind the use of `@ShouldThrowException`:
- there was the problem that Arquillian didn't allow such a Deployment to be _testable_ (making arquillian extension not to be called); which was fixed & merged via @johnament PR arquillian/arquillian-core#135. For this I think that adapting the TCK using my [branch](https://github.com/McFoggy/microprofile-fault-tolerance/tree/issue-200) especially latest commit [83bffc54](https://github.com/McFoggy/microprofile-fault-tolerance/commit/83bffc54053addae283107a972229b73f8131890) will correct that behavior
- and there is the issue mentionned above by @Pandrex247 & @smillidge regarding the retrieval/encapsulation of deployment issues for remote containers. For wildfly, I opened yesterday wildfly/wildfly-arquillian#116 but other container adapters have the same problems.

The second bullet point is not easy to solve and depends on the adaptation of a lot of arquillian container adapters.

There might be other ways to check the deployment failure in Arquillian (including making use of the extension/plugin capabilities), but I don't know the library enough to show the good path.

Emily-Jiang:
no longer needed.

Issue #157, opened 2017-09-18
RetryTimeoutTest seems to contradict spec


johnament:
`RetryTimeoutTest` seems to contradict the [retry.adoc](https://github.com/eclipse/microprofile-fault-tolerance/blob/master/spec/src/main/asciidoc/retry.asciidoc) spec.  Specifically, it indicates that the `TimeoutException` is what triggers the retry.  However, the comments in the test indicate that a `TimeoutException` is expected.

If a `TimeoutException` is expected, that can only mean that upon failure, a retry will throw the underlying exception that caused it to fail.  This behavior is not specified though.

Emily-Jiang:
The TimeoutException will trigger retry if it meets the retryOn condition. After the maximum retries were performed, the exception will pass back to the client. I'll improve the spec to clarify this.

Emily-Jiang:
I think #176 should clarify this.

Emily-Jiang:
Please let me know if you think otherwise.

Issue #159, opened 2017-09-18
CircuitBreakerClientWithRetry has confusing retryOn values


johnament:
This retry on does make much sense: `@Retry(retryOn = {RuntimeException.class, TimeoutException.class}, maxRetries = 7, maxDuration = 20000)`

`TimeoutException` extends from `RuntimeException`, and the spec seems to indicate we always have to treat `TimeoutException` as a retry so this seems contradictory

Emily-Jiang:
That is not what the spec tries to mandate. The spec just says one example. We can reword the spec to clarify the ambiguity.

PR #176, opened 2017-10-11
#159 clarify spec

Emily-Jiang:


johnament:
Based on discussion today, the text should be updated to better reflect that it has to be explicitly added:

> If `@Retry` is used with `@Timeout`, a retry will only be triggered if `TimeoutException` or one of its super classes are defined in the `retryOn` attribute of the `@Retry`

Emily-Jiang:
As per our hangout discussion, I changed based on as per John's comment.

Emily-Jiang:
close #157 #159

Issue #188, opened 2017-11-07
Clarification on CircuitBreaker failureRatio & requestVolumeThreshold


McFoggy:
By implementing the spec in a [portable JEE7 compliant way](https://github.com/McFoggy/portable-mpft-jee) I am puzzling with some CircuitBreaker tests & CircuitBreaker javadoc.

The javadoc looks quite clear to me, I read it like the following:
- _requestVolumeThreshold_: the size of a window of consecutive calls to take into account
- _failureRatio_: a percentage of failures within a window that trigger the circuit to open

But if we look at [CircuitBreaker#testCircuitDefaultSuccessThreshold](https://github.com/eclipse/microprofile-fault-tolerance/blob/1.0/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/CircuitBreakerTest.java#L182), we see that this test defines a window of 4 consecutive calls for which 3 of them are expected to open the circuit (0.75 ratio).

If we look at the test & expected calls, starting from step 6, calls are: `SFFFFC`  (S: succesful call, F: failed call, C: call).
The TestNG test expects the circuit to open only after the 4th failed call _(ie after failed call 10)_.  
But if we consider the sliding window of 4 consecutive calls, then after the sequence `SFFF` _(ie after failed call 9)_, the 0.75 failure ratio is hit and should open the circuit.

If we really respect the tck tests, then for the same service as the one used in the above test sequences like `FFFSFFFSFFFS` which are constantly at 0.75 error rate never open the circuit.

Should we change the test to fail at step 9?

McFoggy:
Also the interpretation of `requestVolumeThreshold()` has to be clarified.

Do we have (1) to wait for a minimum amount of `requestVolumeThreshold()` calls before starting evaluating circuit state, or (2) shall we do our best to fail asap.

[...]

Issue #136, opened 2017-09-10
Inconsistent usage of threshold and ratio


johnament:
In order to configure successes, you use a `successThreshold()` and `requestVolumeThreshold()` to specify how many requests need to be successful.

In order to configure failures, you use `failureRatio()` and `requestVolumeThreshold()`

Its assumed that the `successThreshold` is a count of successes, where as for failures you need to calculate this number.  As a user, this is confusing as in one case I need to calculate a threshold where as in the other I need to just list out counts.  It would be better to only use one or the other.

Emily-Jiang:
In the next release, we can make them symmetric. I don't plan for any API changes at this stage.

McFoggy:
as expressed in #188 for me it is the expectations of the TCK tests that are wrong.

Currently I interpret `1.0` as:
- the change from `SEMI_OPEN` to `CLOSE` depends on `successThreshold()` consecutive success calls
- a change to `OPEN` depends on a `failureRatio()` failures over a window call of `requestVolumeThreshold()`

but I might be wrong in my reading of the API/specs.

Nevertheless, it can valuable to adapt the `SEMI-OPEN` to `CLOSE` step to use a ratio instead of a consecutive number of successful calls.

Emily-Jiang:
@McFoggy your interpretation was correct. will work on this issue shortly.

Emily-Jiang:
maybe we can introduce successRatio to specify the success ratio.

Emily-Jiang:
We discussed this issue in today's hangout. If we introduce the symmetric methods, it will create confusion and conflict if a developer specify both successratio and successThreshold. Which one to win? Besides, these parameters are configurable, it is unclear which value will have the final say. The current methods are sufficient to make the circuit breaker configurable and usable. It could be nice to have the symmetric behaviour but the disadvantage overweights the benefit.

PR #191, opened 2017-11-08
make failureRatio work into requestVolumeThreshold consecutive window calls

McFoggy:
> :warning: __by correcting the behavior of tests this commit introduces changes compared to initial TCK version__ :warning: 

as discussed in #188 & #136 the status change of a circuit into OPEN state should work in a window of `requestVolumeThreshold` calls and should match `failureRatio`. There's no need to wait for `requestVolumeThreshold` failed calls.

I separated the PR in 2 commits. The first one adresses really the issue, the second one clarifies & unifies a bit some expectations on the test file I worked on so by being more consistent it is easier to read/understand the tests.

I can of course squash/rebase both commits or remove the latest one if required.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/CircuitBreakerTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/CircuitBreakerTest.java
@@ -89,23 +89,26 @@ public class CircuitBreakerTest extends Arquillian {
      */
     @Test
     public void testCircuitClosedThenOpen() {
-        for (int i = 0; i < 7; i++) {
+        for (int i = 1; i < 8; i++) {
 
             try {
                 clientForCBWithDelay.serviceA();
 
-                if (i < 4) {
+                if (i < 5) {
                     Assert.fail("serviceA should throw an Exception in testCircuitClosedThenOpen on iteration " + i);
                 }
             }
             catch (CircuitBreakerOpenException cboe) {
-                // Expected on iteration 4
-                if (i < 4) {
+                // Expected on iteration 5
+                if (i < 5) {
                     Assert.fail("serviceA should throw a RuntimeException in testCircuitClosedThenOpen on iteration " + i);
                 }
             }
             catch (RuntimeException ex) {
                 // Expected
+                if (!contains(new int[]{1, 2, 3, 4}, i)) {
+                    Assert.fail("serviceA should not throw a RuntimeException on iteration " + i);
+                }
             }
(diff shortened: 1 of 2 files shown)


McFoggy:
@Emily-Jiang sorry you lost me...

You agreed both in #188 and #136 that we should evaluate a ratio `failureRatio()` of failed events in a window of `requestVolumeThreshold()` successive calls.  
If that is the case, as I understood so far, then at least [CircuitBreakerTest#testCircuitDefaultSuccessThreshold](https://github.com/eclipse/microprofile-fault-tolerance/blob/master/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/CircuitBreakerTest.java#L199) is wrong because the circuit should open after step 9 and not after step 10.
If it is wrong then it should be corrected to reflect what is going to be clarified in the spec, isn't it?

I would expect the following to occure for this test:
````

time
  |
  |    1       F        -> [1 item in window]
  |    2       F        -> [2 items in window]
  |    3       F        -> [3 items in window]
  |    4       F        -> [4 items in window] makes circuit to open
  |    5       C        -> [4 items in window] circuit open -> CircuitBreakerOpenException
  |
  ...wait...            sufficiant time so that considered window is empty
  |
  |    6       S        -> [1 item in window]
  |    7       F        -> [2 items in window]
  |    8       F        -> [3 items in window]
  |    9       F        -> [4 items in window] should open the circuit
  |   10       F    !!! SHOULD be CircuitBreakerOpenException !!!
  |   11       C
  |
  V

F failed call to service
S successful call to service
C tentative to call the service 
````

Emily-Jiang:
My earlier comment "The expectation is "wait for a minimum amount of requestVolumeThreshold() calls before starting evaluating circuit state". I will clarify this in the spec.". I meant to say the rolling window has to be full and then evaluate the circuit. Sorry. I might have misunderstand you. This is a behaviour change. I disagree "There's no need to wait for requestVolumeThreshold failed calls.". Hence my comments earlier.

McFoggy:
> I meant to say the rolling window has to be full and then evaluate the circuit.
Which I perfectly understood. 
But it means that in the above example you consider that the sequence `SFFF` from steps 6 to 9 does not fill the 4 size length expected window? why?

Emily-Jiang:
>Which I perfectly understood.
But it means that in the above example you consider that the sequence SFFF from steps 6 to 9 does not fill the 4 size length expected window? why?

I got you eventually. SFFF will trigger the circuit to open on the 5th invocation. I'll merge your tests soon. Thanks @McFoggy for your contribution!

Issue #183, opened 2017-10-28
Clarify whether method uses declaring class or current class


johnament:
Suppose I have the following:

```java
public class ServiceA {
    @Retry
    public String doIt() { .. }
}
public class ServiceB extends ServiceA {
    @Retry
    public String again() { .. }
}
public class ServiceC extends ServiceA {
     @Override
     public String doIt() { .. }
}
```

Given that these are all fault tolerant methods, and I'm using instances of `ServiceB` and `ServiceC`, what configuration should be used for the method `doIt` in these cases?  Do I use `ServiceA` as the configuration key or the actual impls?

Emily-Jiang:
For ServiceB instance, doIt() is from serviceA, which has retry.. For ServiceC instance, doIt is from ServiceC, which has no retry.

McFoggy:
And if we have
````java
@Retry
public class ServiceD extends ServiceA {
}
````

Which configuration of `@Retry` has to be used when method `doIt()` is called on an instance of ServiceD?
1. the one from ServiceD class 
2. or the one from original method ServiceA

I would hope (1) is expected, is that right?

McFoggy:
Sure, I will.

Emily-Jiang:
@McFoggy @johnament please read my PR to see whether it is clearer.

johnament:
@Emily-Jiang if you assign reviewers to a PR it'll show up in inboxes.  Which PR are you referring to?

McFoggy:
@johnament I guess it is #195 which relates to this issue

PR #192, opened 2017-11-08
add tests for Retry annotation retrieval

McFoggy:
add tests to check the correctness of the annotations picked up when
evaluating @Retry annotation values.
tests include: class inheritance, method override, annotation removal.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/visibility/retry/RetryServiceMethodSuppressLevel.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/visibility/retry/RetryServiceMethodSuppressLevel.java
@@ -1,0 +1,37 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2016-2017 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+package org.eclipse.microprofile.fault.tolerance.tck.visibility.retry;
+
+import java.io.IOException;
+import java.sql.Connection;
+
+import javax.enterprise.context.RequestScoped;
+
+@RequestScoped
+@RS(RetryServiceType.METHOD_SUPPRESS)
+public class RetryServiceMethodSuppressLevel extends BaseRetryService {
+    @Override
+    public Connection service() throws IOException {
+        // no annotation specified only one call expected
+        return super.service();
+    }
+}
+
+
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/visibility/retry/BaseRetryOnClassAndMethodService.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/visibility/retry/BaseRetryOnClassAndMethodService.java
@@ -1,0 +1,50 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2016-2017 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+package org.eclipse.microprofile.fault.tolerance.tck.visibility.retry;
+
+import java.io.IOException;
+import java.sql.Connection;
+
+import javax.enterprise.context.RequestScoped;
+
+import org.eclipse.microprofile.faulttolerance.Retry;
+
+@RequestScoped
+@RS(RetryServiceType.BASE_ROCM)
+@Retry
+public class BaseRetryOnClassAndMethodService implements RetryService {
+    private int nbCalls = 0;
+
+    @Override
+    @Retry(maxRetries = 4)
+    public Connection service() throws IOException {
+        nbCalls++;
+        return delegate();
+    }
+
+    protected Connection delegate() throws IOException {
+        throw new IOException("cannot access delegate service");
+    }
+
+    @Override
+    public int getNumberOfServiceCalls() {
+        return nbCalls;
+    }
+}
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/visibility/retry/BaseRetryOnClassService.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/visibility/retry/BaseRetryOnClassService.java
@@ -27,9 +27,9 @@ import javax.enterprise.context.RequestScoped;
 import org.eclipse.microprofile.faulttolerance.Retry;
 
 @RequestScoped
-@RS(RetryServiceType.BASE)
+@RS(RetryServiceType.BASE_ROC)
 @Retry
-public class BaseRetryService implements RetryService {
+public class BaseRetryOnClassService implements RetryService {
     private int nbCalls = 0;
 
     @Override
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/visibility/retry/BaseRetryOnMethodService.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/visibility/retry/BaseRetryOnMethodService.java
@@ -1,0 +1,49 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2016-2017 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+package org.eclipse.microprofile.fault.tolerance.tck.visibility.retry;
+
+import java.io.IOException;
+import java.sql.Connection;
+
+import javax.enterprise.context.RequestScoped;
+
+import org.eclipse.microprofile.faulttolerance.Retry;
+
+@RequestScoped
+@RS(RetryServiceType.BASE_ROM)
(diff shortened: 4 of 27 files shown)


Emily-Jiang:
Thanks @McFoggy for your contribution! Can you create an issue for this PR please?

McFoggy:
There is one already existing #183 that is referenced in the commit.

neilgsyoung:
These are really useful additional tests. They need a couple of minor changes to run successfully:

1/  The tests need to call the appropriate RetryService in RetryVisibilityTest. They currently all call the baseService.

`    @Test
    public void serviceOverrideClassLevelUsesClassLevelAnnotation() {
        int nbExpectedRetries = 4;  // see RetryServiceOverrideClassLevel class annotation

        checkServiceCall(nbExpectedRetries, this.serviceOverrideClassLevel, // <-replace baseService 
"serviceOverrideClassLevelUsesClassLevelAnnotation");
    }

    @Test
    public void serviceOverrideMethodLevelUsesMethodLevelAnnotation() {
        int nbExpectedRetries = 4;   // see RetryServiceOverrideMethodLevel#service() method annotation

        checkServiceCall(nbExpectedRetries, this.serviceOverrideMethodLevel,  // <-replace baseService
"serviceOverrideMethodLevelUsesMethodLevelAnnotation");
    }
    
    @Test
    public void serviceRetryRemovedAtMethodLevel() {
        int nbExpectedRetries = 0;  // see RetryServiceMethodSuppressLevel#service() method with no annotation

        checkServiceCall(nbExpectedRetries, this.serviceSuppressMethodLevel,  // <-replace baseService
"serviceRetryRemovedAtMethodLevel");
    }`

2/ The RetryServiceMethodSuppressLevel class needs reworking. It will currently inherit the default Retry policy from the BaseRetryService and the service will be executed 4 times. 
One way to rework this is to...
`public class RetryServiceMethodSuppressLevel extends AlternateRetryServiceOverrideMethodLevel`

....where AlternateRetryServiceOverrideMethodLevel, looks like this...

`@RequestScoped
@RS(RetryServiceType.METHOD_OVERRIDE)
public class AlternateRetryServiceOverrideMethodLevel extends AlternateBaseRetryService {
    @Override
    @Retry(maxRetries = 4)
    public Connection service() throws IOException {
        return super.service();
    }
}`

and AlternateBaseRetryService looks like this...

`@RequestScoped
@RS(RetryServiceType.BASE)
public class AlternateBaseRetryService implements RetryService {
    private int nbCalls = 0;

    @Override
    public Connection service() throws IOException {
        nbCalls++;
        return delegate();
    }

    protected Connection delegate() throws IOException {
        throw new IOException("cannot access delegate service");
    }

    @Override
    public int getNumberOfServiceCalls() {
        return nbCalls;
    }
}`

McFoggy:
@neilgsyoung can you please correct the formatting of your comment?

neilgsyoung:
That hopefully looks better.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/visibility/retry/RetryServiceMethodSuppressLevel.java
```diff
@@ -0,0 +1,37 @@
...
+package org.eclipse.microprofile.fault.tolerance.tck.visibility.retry;
+
+import java.io.IOException;
+import java.sql.Connection;
+
+import javax.enterprise.context.RequestScoped;
+
+@RequestScoped
+@RS(RetryServiceType.METHOD_SUPPRESS)
+public class RetryServiceMethodSuppressLevel extends BaseRetryService {
```
johnament: Don't forget that since the FT annotations are all `@Inherited` the original annotations will be retained.  E.g. calling `RetryServiceMethodSuppressLevel.class.getMethod("service").getAnnotation(Retry.class)` will still return the object.

I don't think your test for this will pass for anyone.

McFoggy: @johnament this came from #183 and the answer of from @Emily-Jiang to your issue.

I have no pb in keeping it or removing it

> I don't think your test for this will pass for anyone.

unless implementations use `getDeclaredAnnotation()` instead of `getAnnotation()`

Honestly, I see interests in both way of handling this annotation retrieval.
It is a matter of decision & balance between the strict conformity & the facility offer to disable a feature by overriding a method.
Moreover, looking at the javadoc of `@Inherited`, no strict description is done when it comes to method annotation lookup.

If there were votes available on an issue in github we could go with a democratic way ; instead I'd expect project leads to take a decision.

johnament: Well, the project leads would be John C or Kevin.  I'm not in agreement with Emily's POV, since these are `@Inherited`.  I would recommend that we remove `@Inherited` if this is the behavior expected.

Emily-Jiang: Since the subclass inherits the annotations, the method will have the annotations in RetryServiceMethodSuppressLevel. Actually, think it further. It might be better not to make them inherited, because we cannot easily get rid of the annotations. If they are not inherited, we can always add them. We can discuss this on tomorrow's (15th Nov) call as well.

johnament: Isn't that what I had just recommended, removing `@Inherited`?

Emily-Jiang: Yes, John. I think I am not in disagreement with you on this. Since this is behaviour changes, we should have further discussion on tomorrow's call.

McFoggy:
@neilgsyoung let's wait for the final conclusions on how to handle the annotations to take into account your second remark (or a reworked version).

also 
- I added a simple test that  demo what happens at J2SE level for our `Retry` annotation retrieval
- I prepared also new scenarios to be tested

McFoggy:
FYI I started a boring work which goal is to understand existing behaviors of annotation retrieval in established & already specified areas (EJBs & JTA). 

You'll find the project there: https://github.com/McFoggy/jee-tx-resolution.

I'll continue to enhance it to add some missing coverage and to run it on different containers ; but honestly it already surprised me in how EJB spec and JTA spec were behaving for some corner cases (at least under wildfly).  
It will deserve to have an overview (in a table probably) of the demoed behaviors and a README.

McFoggy:
@antoinesd as discussed today I added missing tests.

I have rebased and will squash on demand.

My impl does not pass the tests because currently I was looking for the annotations by hand first looking for annotation on method then on class and redo the same process on parent classes.  
From the understanding I had of the discussions on inheritance you had with @Emily-Jiang, I implemented the tests on the TCK in this PR more to enable a pure delegation to `java.lang.reflect.Method.getAnnotation()` & `java.lang.class#getAnnotation()`

Emily-Jiang:
@McFoggy we hit some issues when running the tests. Did you fix the issue @neilgsyoung suggested?

McFoggy:
> Did you fix the issue @neilgsyoung suggested?

@Emily-Jiang yes I think I did, but I can have make mistakes.

> we hit some issues when running the tests.

The tests are there to express an order of retrieval after the discussions that occured. 
In my opinion the question is not that Liberty or any other implementation does or does not pass the tests, the question is : are the tests reflecting the retrieval pattern that is expected in the spec.

McFoggy:
I am even not sure about the expectations I have put in the tests.  
Let's take the example of `RetryVisibilityTest#serviceOverrideClassLevelUsesClassLevelAnnotation`.  
We have a common service

````java
@Retry            // use defaults maxRetries = 3
BaseService {
     void service() {...}
}
````
which expresses: _define a retry policy of 3 for all exposed methods of this service and of sub implementations_ 

Then we have a specialization:
````java
@Retry(maxRetries = 4)
Specialization1Service extends BaseService {
}
````
which I assume means: _for the Specialization1 use a retry policy of 4_

Thus comes the question:
- do we expect the number of retry to be really 4 or 3 when calling _`service()`_ method on an instance of `Specialization1Service`.

If the implementation uses the annotation found starting from the method location, then obviously it will be 3 ; is that expected? If the implementation uses first the target object of the call then it will be 4.
The tests are written to expose coverage of such cases. 
If we think it is 3 then let's put 3 in the test expectation, if we think it is 4 then let's put 4 in the test expectation. The PR provided the test coverage, let's review it now to discuss the expectations.  
They are not always easy to define...

Emily-Jiang:
> Thus comes the question:

    do we expect the number of retry to be really 4 or 3 when calling service() method on an instance of Specialization1Service.

If the implementation uses the annotation found starting from the method location, then obviously it will be 3 ; is that expected? If the implementation uses first the target object of the call then it will be 4.
The tests are written to expose coverage of such cases.
If we think it is 3 then let's put 3 in the test expectation, if we think it is 4 then let's put 4 in the test expectation. The PR provided the test coverage, let's review it now to discuss the expectations.
They are not always easy to define...

From my understanding, it should be 4.

McFoggy:
> From my understanding, it should be 4.

the tests are written with this philosophy in mind or at least it was the intent...

McFoggy:
@antoinesd @Emily-Jiang @johnament @neilgsyoung any comment on the expectations of the tests?

neilgsyoung:
I need to investigate further but my initial run of the new visibility tests against OpenLiberty shows 15 passing and 5 failing tests...

Results :

Failed tests:
  RetryVisibilityTest>Arquillian.run:138->serviceBaseROCMNoRedefinition:215->checkServiceCall:248 in RetryVisibilityTest#serviceBaseROCMNoRedefinition service() should have been called exactly 4 times expected [4] but found [5]
  RetryVisibilityTest>Arquillian.run:138->serviceBaseROCMOverridedClassLevelNoMethodOverride:222->checkServiceCall:248 in RetryVisibilityTest#serviceBaseROCMOverridedClassLevelNoMethodOverride service() should have been called exactly 6 times expected [6] but found [5]
  RetryVisibilityTest>Arquillian.run:138->serviceBaseROMNoRedefinition:201->checkServiceCall:248 in RetryVisibilityTest#serviceBaseROMNoRedefinition service() should have been called exactly 1 times expected [1] but found [4]
  RetryVisibilityTest>Arquillian.run:138->serviceBaseROMOverridedClassLevelMethodOverride:187->checkServiceCall:248 in RetryVisibilityTest#serviceBaseROMOverridedClassLevelMethodOverride service() should have been called exactly 5 times expected [5] but found [4]
  RetryVisibilityTest>Arquillian.run:138->serviceRetryRemovedAtMethodLevel:159->checkServiceCall:248 in RetryVisibilityTest#serviceRetryRemovedAtMethodLevel service() should have been called exactly 1 times expected [1] but found [4]

Tests run: 20, Failures: 5, Errors: 0, Skipped: 0

Emily-Jiang:
@McFoggy can we go through these failing tests in the next hangout?

neilgsyoung:
Working through the failures that I see with OpenLiberty,

1. serviceBaseROCMNoRedefinition
Parent class: No annotation at class level, maxRetries=4 on method
Child class: No annotation at class level, no overriding method
TestExpectation: 3 retries
OpenLiberty TCK: 4 retries

Comment: The annotation on the service method in the parent class should be honoured, the service should be retried 4 times (5 executions)

2. serviceBaseROCMOverridedClassLevelNoMethodOverride
Parent class: No annotation at class level, maxRetries=4 on method
Child class: maxRetries=5 at class level, no overriding method
TestExpectation: 5 retries
OpenLiberty TCK: 4 retries

Comment: The annotation on the service method in the parent class should be honoured, the service should be retried 4 times (5 executions)

3. serviceBaseROMNoRedefinition
Parent class: No annotation at class level, default @Retry annotation on method
Child class: No annotation at class level, no overriding method
TestExpectation: 0 retries
OpenLiberty TCK: 3 retries

Comment: From org.eclipse.microprofile.faulttolerance.Retry, the default maxRetries value is 3,
the test should be changed to expect 4 executions of service().


4. serviceBaseROMOverridedClassLevelMethodOverride
Parent class: No annotation at class level, default @Retry annotation on method
Child class: maxRetries=4 at class level, no overriding method
TestExpectation: 4 retries
OpenLiberty TCK: 3 retries

Comment: The annotation on the service method in the parent class should be honoured.
From org.eclipse.microprofile.faulttolerance.Retry, the default maxRetries value is 3,
the test should be changed to expect 4 executions of service().

5. serviceRetryRemovedAtMethodLevel
Parent class: Default @Retry annotation at class level, no annotation on method
Child class: No annotation at class level, no annotation on overriding method
TestExpectation: 0 retries
OpenLiberty TCK: 3 retries

Comment: From org.eclipse.microprofile.faulttolerance.Retry, the default maxRetries value is 3,
the test should be changed to expect 4 executions of service().

neilgsyoung:
I'll use #213 to address the expected test results I described above.

Issue #180, opened 2017-10-27
Unclear on behavior of iteration count


johnament:
I'm trying to debug my way through this test -> [BulkheadAsynchRetryTest.java#L165](https://github.com/eclipse/microprofile-fault-tolerance/blob/master/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadAsynchRetryTest.java#L165)

What strikes me is that it launches 1000 iterations, seemingly for unknown reasons, to verify that a `BulkheadException` is thrown.

I don't understand this test at all unfortunately.

At what point are you expecting the BHE to be thrown?  I don't see anything about this test that would be triggering a BHE.

hutchig:
I wrote this test and will have a look. Apologies if it is causing a problem.

hutchig:
1000  is just a number that is obviously 'too many' for the bulkhead * the retries. 
So we have a loop attempting to firing 1000 things into an asych bulkhead of size 5 and waitingTaskQue 5 and maxRetries of 10. 
```
for (int i = 0; i < 1000; i++) {
                results[i] =  BulkheadRapidRetry55MethodAsynchBean::test(new Checker(1 * 1000, td));
            }

public class BulkheadRapidRetry55MethodAsynchBean implements BulkheadTestBackend {
    @Override
    @Bulkhead(waitingTaskQueue = 5, value = 5)
    @Asynchronous
    @Retry(retryOn =
    { BulkheadException.class }, delay = 1, delayUnit = ChronoUnit.MICROS, maxRetries = 10, maxDuration=999999)
    public Future test(BackendTestDelegate action) throws InterruptedException {
        Utils.log("in business method of bean " + this.getClass().getName());
        return action.perform();
    }
```
The queue will fill up and throw a bulkhead exception, this will be retried 10 times, to again fail, rinse and repeat until the bulkhead exception percolates back to the client on 'RetryOverload'. 
Hence the name of the test "testBulkheadMethodAsynchronous55RetryOverload". 

Have you a reasoning on how the test is incorrect? (Happy to learn if I am wrong as I can't see the error currently.)

hutchig:
[...]

PR #193, opened 2017-11-13
Remove @Asynchronous annotation from sync tests

Pandrex247:
Some of the synchronous tests are annotated with `@Asynchronous`, making them, well, not synchronous.

These tests also already submit themselves to an executor to simulate multiple simultaneous calls, so with this annotation present the tests are being submitted to an executor, to be then be submitted to a second executor.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadSynchRetryTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadSynchRetryTest.java
@@ -237,7 +237,7 @@ public class BulkheadSynchRetryTest extends Arquillian {
     public void testNoRetriesBulkhead() {
         int threads = 30;
         int maxSimultaneousWorkers = 5;
-        int expectedTasks = 10;
+        int expectedTasks = 5;
         TestData td = new TestData(new CountDownLatch(expectedTasks));
         threads(threads, zeroRetryBean, maxSimultaneousWorkers, expectedTasks, td);
         td.check();
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadSynchRetryTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadSynchRetryTest.java
@@ -29,6 +29,7 @@ import javax.inject.Inject;
 import org.eclipse.microprofile.fault.tolerance.tck.bulkhead.clientserver.BackendTestDelegate;
 import org.eclipse.microprofile.fault.tolerance.tck.bulkhead.clientserver.Bulkhead55ClassSynchronousRetryBean;
 import org.eclipse.microprofile.fault.tolerance.tck.bulkhead.clientserver.Bulkhead55MethodSynchronousRetryBean;
+import org.eclipse.microprofile.fault.tolerance.tck.bulkhead.clientserver.BulkheadRapidRetry50MethodSynchBean;
 import org.eclipse.microprofile.fault.tolerance.tck.bulkhead.clientserver.BulkheadRapidRetry550MethodSynchBean;
 import org.eclipse.microprofile.fault.tolerance.tck.bulkhead.clientserver.BulkheadRapidRetry55ClassSynchBean;
 import org.eclipse.microprofile.fault.tolerance.tck.bulkhead.clientserver.BulkheadRapidRetry55MethodSynchBean;
@@ -82,7 +83,10 @@ public class BulkheadSynchRetryTest extends Arquillian {
     private BulkheadRapidRetry55MethodSynchBean rrMethodBean;
 
     @Inject
-    private BulkheadRapidRetry550MethodSynchBean zeroRetryBean;
+    private BulkheadRapidRetry50MethodSynchBean zeroRetryBean;
+    
+    @Inject
+    private BulkheadRapidRetry550MethodSynchBean zeroRetryWaitingQueueBean;
 
     /**
      * This is the Arquillian deploy method that controls the contents of the
@@ -240,6 +244,21 @@ public class BulkheadSynchRetryTest extends Arquillian {
         int expectedTasks = 5;
         TestData td = new TestData(new CountDownLatch(expectedTasks));
         threads(threads, zeroRetryBean, maxSimultaneousWorkers, expectedTasks, td);
+        td.check();
+    }
+    
+    /**
+     * Test that that the waitingTaskQueue parameter is ignored due to the absence of 
+     * the Asynchronous annotation. Only 5 tasks should go through, as the waiting
+     * queue size should be ignored.
+     */
+    @Test()
+    public void testIgnoreWaitingTaskQueueBulkhead() {
+        int threads = 30;
+        int maxSimultaneousWorkers = 5;
+        int expectedTasks = 5;
+        TestData td = new TestData(new CountDownLatch(expectedTasks));
+        threads(threads, zeroRetryWaitingQueueBean, maxSimultaneousWorkers, expectedTasks, td);
         td.check();
     }
 
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/BulkheadRapidRetry55MethodSynchBean.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/BulkheadRapidRetry55MethodSynchBean.java
@@ -30,7 +30,7 @@ import org.eclipse.microprofile.faulttolerance.Retry;
 import org.eclipse.microprofile.faulttolerance.exceptions.BulkheadException;
 
 /**
- * A simple method level Asynchronous @Bulkhead bean that has a retry option.
+ * A simple method level synchronous @Bulkhead bean that has a retry option.
  *
  * @author Gordon Hutchison
  */
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/BulkheadRapidRetry55MethodSynchBean.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/BulkheadRapidRetry55MethodSynchBean.java
@@ -25,7 +25,6 @@ import java.util.concurrent.Future;
 import javax.enterprise.context.ApplicationScoped;
 
 import org.eclipse.microprofile.fault.tolerance.tck.bulkhead.Utils;
-import org.eclipse.microprofile.faulttolerance.Asynchronous;
 import org.eclipse.microprofile.faulttolerance.Bulkhead;
 import org.eclipse.microprofile.faulttolerance.Retry;
 import org.eclipse.microprofile.faulttolerance.exceptions.BulkheadException;
@@ -41,7 +40,6 @@ public class BulkheadRapidRetry55MethodSynchBean implements BulkheadTestBackend 
     @Override
     @ApplicationScoped
     @Bulkhead(waitingTaskQueue = 5, value = 5)
-    @Asynchronous
     @Retry(retryOn =
     { BulkheadException.class }, delay = 1, delayUnit = ChronoUnit.MILLIS, maxRetries = 10, maxDuration=999999)
     public Future test(BackendTestDelegate action) throws InterruptedException {
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/Bulkhead55ClassSynchronousRetryBean.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/Bulkhead55ClassSynchronousRetryBean.java
@@ -25,7 +25,6 @@ import java.util.concurrent.Future;
 import javax.enterprise.context.ApplicationScoped;
 
 import org.eclipse.microprofile.fault.tolerance.tck.bulkhead.Utils;
-import org.eclipse.microprofile.faulttolerance.Asynchronous;
 import org.eclipse.microprofile.faulttolerance.Bulkhead;
 import org.eclipse.microprofile.faulttolerance.Retry;
 import org.eclipse.microprofile.faulttolerance.exceptions.BulkheadException;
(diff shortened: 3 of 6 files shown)


Pandrex247:
This actually breaks one of the synchronous tests, `testNoRetriesBulkhead`, because this test actually relies upon asynchronous behaviour.

The test expects there to be 10 executions, as the bean that this test uses as a bulkhead of 5, with a waiting task queue of 5. The test therefore is expecting 5 threads to execute immediately, and then 5 more to execute once those 5 have finished.

1) The test is meant to be checking synchronous behaviour, not asynchronous
2) The spec (to me anyway) heavily implies that if the `@Asynchronous` annotation is not present, then the `waitingTaskQueue` parameter for `@Bulkhead` should be ignored: 

> When `@Bulkhead` is used with `@Asynchronous`, the thread pool isolation approach will be used. If `@Bulkhead` is used without `@Asynchronous`, the semaphore isolation approach will be used.

Pandrex247:
In my latest commit I've removed the `waitingTaskQueue` parameter from the bean being used in the faulty test (`testNoRetriesBulkhead`), in an effort to keep the test simple and clear.

What should be the expected behaviour for if you have this parameter specified without the `@Asynchronous` annotation? Should it be ignored, or should a validation exception be thrown?
If the former, should we create a similar test that puts this parameter back in and checks that it gets ignored? Or should I just put it back in to this test and alter the name / description a little?

Emily-Jiang:
waitingTaskQueue is ignored without @Asynchronous annotation. It was specified in the javadoc of Bulkhead.java.

Emily-Jiang:
+1 on adding a test to make sure it is ignored without @Asynchronous.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/BulkheadRapidRetry55MethodSynchBean.java
```diff
@@ -25,13 +25,12 @@
 import javax.enterprise.context.ApplicationScoped;
 
 import org.eclipse.microprofile.fault.tolerance.tck.bulkhead.Utils;
-import org.eclipse.microprofile.faulttolerance.Asynchronous;
 import org.eclipse.microprofile.faulttolerance.Bulkhead;
 import org.eclipse.microprofile.faulttolerance.Retry;
 import org.eclipse.microprofile.faulttolerance.exceptions.BulkheadException;
 
 /**
- * A simple method level Asynchronous @Bulkhead bean that has a retry option.
```
Emily-Jiang: Why did you change the purpose of this test?

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadSynchRetryTest.java
```diff
@@ -237,11 +241,26 @@ public void testBulkheadRetryClassSynchronous55() {
     public void testNoRetriesBulkhead() {
         int threads = 30;
         int maxSimultaneousWorkers = 5;
-        int expectedTasks = 10;
+        int expectedTasks = 5;
```
Emily-Jiang: Please explain why you changed this number?

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/BulkheadRapidRetry55MethodSynchBean.java
Pandrex247: None of the tests I've edited are used by any asynchronous test - they're used exclusively by _BulkheadSynchRetryTest_. I changed the class descriptions because they're wrong - they've presumably been copied over from other Asynchronous tests without adjusting, or were refactored incorrectly.

Pandrex247: It's even in the class name :)

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadSynchRetryTest.java
Pandrex247: I changed this number because it's wrong. As I explained in my second comment, the number of 10 is **only** correct if it's working under asynchronous (thread pool style) bulkhead.

Having removed the `@Asynchronous` annotation, this test should only expect 5 threads to return: It fires off 30 threads to a bulkhead with a value of 5, so 5 should execute, and 25 should immediately return with bulkhead errors. Given that there's no retry on this method (_maxRetries=0_), none of those 25 errors should try again.

Previously, the method was annotated with a bulkhead of 5, a waiting task queue of 5, and the `@Asynchronous` annotation, so the previous number of 10 was the correct number to pass the test, but the test itself was wrong - it shouldn't be testing asynchronous behaviour. It would fire off 30 threads, 5 would execute, 5 more would wait, and 20 would return bulkhead errors.

As requested, I put in an extra test to check that without the `@Asynchronous` annotation the _waitingTaskQueue_ parameter was ignored. The git diff though shows it a little funny, so I'll try to explain:

1. I removed the `@Asynchronous` annotation and _waitingTaskQueue_ parameter from the original class to fix the test and remove the spurious parameter respectively - `BulkheadRapidRetry550MethodSynchBean.java`
2. I then changed the name of this class to match the naming convention already being used, from _550_ to _50_, as it no longer had 3 values (_value_, _waitingTaskQueue_, and _maxRetries_) - it's now called `BulkheadRapidRetry50MethodSynchBean.java` because it only has 2 values (_value=5_ and _maxRetries=0_)
3. I added in my new test which is almost identical to the original, just with `@Asynchronous` removed, so as to test that _waitingTaskQueue_ is ignored when the `@Asynchronous` annotation isn't present - it should behave exactly the same as the original test (now `BulkheadRapidRetry50MethodSynchBean.java`), returning 5.

Pandrex247:
As requested, I put in an extra test to check that without the `@Asynchronous` annotation the _waitingTaskQueue_ parameter was ignored. The git diff though shows it a little funny, so I'll try to explain:

1. I removed the `@Asynchronous` annotation and _waitingTaskQueue_ parameter from the original class to fix the test and remove the spurious parameter respectively - `BulkheadRapidRetry550MethodSynchBean.java`
2. I then changed the name of this class to match the naming convention already being used, from _550_ to _50_, as it no longer had 3 values (_value_, _waitingTaskQueue_, and _maxRetries_) - it's now called `BulkheadRapidRetry50MethodSynchBean.java` because it only has 2 values (_value=5_ and _maxRetries=0_)
3. I added in my new test which is almost identical to the original, just with `@Asynchronous` removed, so as to test that _waitingTaskQueue_ is ignored when the `@Asynchronous` annotation isn't present - it should behave exactly the same as the original test (now `BulkheadRapidRetry50MethodSynchBean.java`), returning 5.

McFoggy:
for me it is a +1 for this PR. 

Synchronous tests should not be using methods annotated `@Asynchronous` as `@Asynchronous + @Bulkhead` handling differs from just `@Bulkhead` usage.

hutchig:
Some of the tests intentionally use an Asynchronous annotation as a (not ideal but portable) way of creating parallelism across JEE and JSE (e.g. Hammock)  before going on to test something else. This problem was discussed a little in issue #180 at the end.
I would not be safe to simply remove Asynch from all the tests that have Sync in the names as at least some are intended to be there and do what they do. 
Very early on, I do remember when I first encountered the Bulkhead queue I had assumed it would apply to both Thread based and Semaphore based Bulkheads (but soon learned otherwise) so perhaps there is something there - I will now go and look in detail at the PR but wanted to respond now initially.

McFoggy:
The problem here is that we introduce the couple `@Asynchronous + @Bulkhead` instead of having just `@Bulkhead`. Both are not equivalent as one is queued, non blocking wiling to throw an Exception while the other is different. At least I understood it like that.

I agree with @johnament proposal in #180 (having a SPI to generate threads) that should solve this.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/BulkheadRapidRetry55MethodSynchBean.java
Emily-Jiang: The class name was wrong! The test itself is to tests async aspect. The correct fix is to rename the class rather than repurpose the test. Otherwise, there is no async test left.

Emily-Jiang:
@Bulkhead + @Asynchronous = threading flavour of bulkhead
@Bulkhead = semaphore flavour of bulkhead

You can directly manage the thread pool in the app server yourselves. It is implementation detail.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/BulkheadRapidRetry55MethodSynchBean.java
Pandrex247: ? Yes there is: https://github.com/eclipse/microprofile-fault-tolerance/blob/master/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/BulkheadRapidRetry55MethodAsynchBean.java

It's almost identical except the linked async bean isn't application scoped

Emily-Jiang:
Just clarify: the test method names should be fixed to reflect the tests instead of changing the tests.

hutchig:
I think removing @Asynchronous from Bulkhead55ClassSynchronousRetryBean is a good change and is there in error. Both the tests that use this class (in BulkheadSynchRetryTest) make use of the 'threads' method to loop round and call the ExecutorService to create parallelism (I did this in the earlier tests I wrote) so the Asynch is not in Bulkhead55ClassSynchronousRetryBean to create parallelism but due to a copy and paste error most likely. As the bean class name, the test class name and the tests all have "Sychronous" in the title the sensible direction is to make the bean Synchronous and adjust for the lack of queueing.

Pandrex247:
@Emily-Jiang but the tests are wrong!
These tests are meant to be testing **synchronous** behaviour, any test annotated with `@Asynchronous` is by definition not synchronous! They follow different semantics as you yourself put:

> @bulkhead + @asynchronous = threading flavour of bulkhead
@bulkhead = semaphore flavour of bulkhead

There are already tests for asynchronous bulkheads (`BulkheadAsynchTest` and `BulkheadAsynchRetryTest`), many of them extremely similar to what these tests are already doing. If we change the method names to match what the tests are doing, rather than changing the tests as I've done, what we actually have is no synchronous bulkhead tests, which is a gigantic hole in the TCK.

hutchig:
The same is true for BulkheadRapidRetry550MethodSynchBean - it is used in a test for the Synchronous Bulkhead in BulkheadSynchRetryTest.java that uses the threads() method and thus executor and was intended to test Synch behaviours I think. I did most of the beans' header comments in bulk at the same time and did this by looking at the annotations rather than  jumping to how they were used at all - hence the mistake in the Asynch annotation being there being reflected in the comment. I think the Asynchronous annotation should be removed and any difference in the test due to no queue at the Bulkhead accounted for.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/BulkheadRapidRetry55MethodSynchBean.java
Emily-Jiang: thanks @Pandrex247 !

Emily-Jiang:
@Pandrex247 thanks for the further explanation. Since @hutchig stated it is an test error, we have tests to cover async, it is good to fix the tests. Thanks again @Pandrex247 for your patience!

Emily-Jiang:
We will run against Open Liberty to make sure. @johnament can you assess whether these tests working on hammock, as I know you are implementing FT as well?

hutchig:
BulkheadRapidRetry55ClassSynchBean is from a later set of tests.

That test does:
    * Check we do not loose anything from the queue due to exceptions covered
      by Retry at the class level. The Checker backends will throw an exception
      the first time their 'perform' method is called but will run OK when
      retried.
So this does certainly need a queue to do this test so that means the bean
has to be Asynch.  So this is not really a test that lives in BulkheadSynRetryTest
I think the function is covered in BulkheadAsynchRetryTest so I think the test
testBulkheadQueReplacesDueToClassRetryFailures should probably be deleted.
That is the only test that uses rrClassBean and that is the only bean that uses
 BulkheadRapidRetry55ClassSynchBean. So I think the test(testBulkheadQueReplacesDueToClassRetryFailures), the bean variable(rrClassBean) and the class (BulkheadRapidRetry55ClassSynchBean) should be deleted.

hutchig:
The same reasoning applies to BulkheadRapidRetry55MethodSynchBean.java
this is only used in the variable rrMethodBean which is only used in testBulkheadQueReplacesDueToMethodRetryFailures.

This test is similar to testBulkheadMethodAsynchronousRetry55Trip.
So I think BulkheadRapidRetry55MethodSynchBean.java, rrMethodBean
and testBulkheadQueReplacesDueToMethodRetryFailures should all be deleted. 

I think this whole problem stemmed from me originally thinking that both
Sync and Asynch Bulkheads had queues - then changing a set of invalid
tests to get them to work rather than stepping back and removing them.

Emily-Jiang:
@hutchig since @McFoggy is trying to re-purpose the tests to cover sync, I think it is better than deleting them and left a hole there.

Emily-Jiang:
fyi @McFoggy. I had a chat with @hutchig. Gordon will do a PR on the top of this PR to remove a couple of obsolete tests and adopt a couple of tests from your PR.

hutchig:
I will do a PR (which is queing behind the task I am doing now) which takes on board the points
above, includes the mentioned function, but which will retains the organisation of the test client names/contents etc. I can't do a test for the Sync queue - as it does not exist (my mistake :-) )

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadSynchRetryTest.java
hutchig: I think the changes above are good.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadSynchRetryTest.java
```diff
@@ -237,11 +241,26 @@ public void testBulkheadRetryClassSynchronous55() {
...
+     */
+    @Test()
+    public void testIgnoreWaitingTaskQueueBulkhead() {
+        int threads = 30;
+        int maxSimultaneousWorkers = 5;
+        int expectedTasks = 5;
+        TestData td = new TestData(new CountDownLatch(expectedTasks));
+        threads(threads, zeroRetryWaitingQueueBean, maxSimultaneousWorkers, expectedTasks, td);
+        td.check();
+    }
```
hutchig: I think the test above is fine.

PR #195, opened 2017-11-14
#183 - clarify the config on annotation

Emily-Jiang:

### DIFF
--- a/spec/src/main/asciidoc/configuration.asciidoc
+++ b/spec/src/main/asciidoc/configuration.asciidoc
@@ -44,6 +44,8 @@ The value of each parameter can be overridden individually or globally.
 * Override individual parameters
 The annotation parameters can be overwritten via config properties in the naming convention of `<classname>/<methodname>/<annotation>/<parameter>`.
 
+The `<classname>` and `<methodname>` must be the class name and method name where the annotation is declared upon. 
+
 In the following code snippet, in order to override the `maxRetries` for serviceB invocation to 100, set the config property `com.acme.test.MyClient/serviceB/Retry/maxRetries=100`
 Similarly to override the `maxDuration` for ServiceA, set the config property
 


McFoggy:
@johnament if we go this way then we could even think about allowing to configure that at any "level" (in terms of inheritance/override/visibility) it is relevant to look at _(and for which clarifications are needed see #192)_.
For example in a class hierarchy of `A#m() > B > C`, if an FT annotation is defined only on A class, then a configuration provided for A, A#m, B or C could change the behavior of a `C#m()` call.
This behavior would match your proposal I think.

On the other hand we can also have a simpler first version saying that only the explicitly defined places can be overridden. In above example, we could only define a configuration for the annotation at A class level. This is perhaps the initially proposed Emily proposal.

Can't we adopt a step wise approach by defining the easiest/simplest one first (Emily's one), then enhance it for a next version (your proposal with respect of inheritance/override discover rules)?

Emily-Jiang:
+1 @McFoggy ! This limitation can be further improved in the next release once the implementors are catching up. @johnament you can go ahead to provide a bit more in your implementation. I don't see there is conflict.

McFoggy:
FYI there is also a paragraph at the bottom of the file [spec file]( http://microprofile.io/project/eclipse/microprofile-fault-tolerance/spec/src/main/asciidoc/configuration.asciidoc) 

> If an annotation is not present, the configured properties are ignored. ...

that expect an annotation to be present in order to modify attributes.

johnament:
Err no there is a big difference.  Consider the following class hierarchy:

```java
public abstract class AbstractTolerantBean {
   protected abstract POJO getPojo();

   @Timeout
   @Retry
   public POJO execute() {
       return getPojo();
   }

}

@ApplicationScoped
public class A extends AbstractTolerantBean { ... }

@ApplicationScoped
public class B extends AbstractTolerantBean { ... }
```

As a user, I now need to know that the configuration for `A` and `B` is actually based on the base class, but if I introduce a new fault tolerant method in `B` then that is configured under `B`'s namespace.  This is very difficult to grasp.  In addition, most CDI lifecycle methods need not fire for `AbstractTolerantBean` so there is no real connotation here.

I also believe we need to have a TCK test to go with this.

McFoggy:
@johnament the way the configuration should work should be independent from the way the configuration is reported to the end user and thus it can be left to the implementations to clarify that.

On your example I can imagine an implementation exposing something like (in pseudo model it could be JSON served via REST, a GUI to administrate, ...):

```
+ AbstractTolerantBean#execute
    - Timeout
        ° value=1000
        ° unit=MILLIS
    - Retry
        ° maxRetries=3
        ° delay=0
        ° delayUnit=MILLIS
        ° ...
+ A#execute
    - Timeout
        ° value=1000 defined by AbstractTolerantBean#execute.Timeout.value
        ° unit=MILLIS defined by AbstractTolerantBean#execute.Timeout.unit
    - Retry
        ° maxRetries=3 defined by AbstractTolerantBean#execute.Retry.maxRetries
        ° delay=0 defined by AbstractTolerantBean#execute.Retry.maxRetries
        ° delayUnit=MILLIS defined by AbstractTolerantBean#execute.Retry.maxRetries
        ° ...
+ B#execute
    - Timeout
        ° value=1000 defined by AbstractTolerantBean#execute.Timeout.value
        ° unit=MILLIS defined by AbstractTolerantBean#execute.Timeout.unit
    - Retry
        ° maxRetries=3 defined by AbstractTolerantBean#execute.Retry.maxRetries
        ° delay=0 defined by AbstractTolerantBean#execute.Retry.maxRetries
        ° delayUnit=MILLIS defined by AbstractTolerantBean#execute.Retry.maxRetries
        ° ...
```
Having such a reporting, knowing where/how to defined a configuration becomes trivial _(in respect of the specification possibilities, see below)_.

Then if we use the step wise approach for the spec:

- in step 1 (current), the spec will expect an implementation to accept a configuration only where the annotation is present. Thus the spec would not expect your implementation to accept `B/execute/Timeout/value=2000`. **But** on the other hand it should not prevent you from accepting it.

- in step 2, the spec will specify that your implementation must accept `B/execute/Timeout/value=2000`

Thus after specifying an explicit `Timeout` for B using `B/execute/Timeout/value=2000`, then B configuration could be reported as 

```
+ B#execute
    - Timeout
        ° value=2000
        ° unit=MILLIS defined by AbstractTolerantBean#execute.Timeout.unit
```

Emily-Jiang:
yes. The spec can be limited in the 1st release (the spec is correct in theory) and then widen the usage in the next release. We have one question:
Do we want to improve the tck and release a 1.0.1 release or 
continue doing a 1.1.0 release to improve the spec and tck?

I am ok with either approach. What do you think as you two are implementing the current spec? Do you @McFoggy @johnament  want to implement the 1.1.0 directly and then release it?

Emily-Jiang:
closes #183

Issue #197, opened 2017-11-21
Circuit Breaker Rolling failure window behaviour: follow up to issue #188


neilgsyoung:
We need to clarify the behaviour of Circuit Breakers as discussed in issue #188. I think that the original  [CircuitBreaker#testCircuitDefaultSuccessThreshold](https://github.com/eclipse/microprofile-fault-tolerance/blob/1.0/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/CircuitBreakerTest.java) test was working as designed but that we need some clearer words in the Spec to explain the behaviour.

The test illustrates the transition of the Circuit between closed, open and half-open states. In the test, when the circuit state is closed, a rolling failure window is in play, configured through the requestVolumeThreshold and failureRatio parameters. The parameters define a rolling failure window of 4 executions with a failureRatio of 0.75. Three of the four executions fail, so that the circuit transitions to open state. The test then delays sufficiently for the circuit to transition to half-open state. In half-open state a "trial success" window comes into play. The successThreshold parameter allows the configuration of the number of trial calls that must  succeed before the circuit can be closed but in the test this parameter is allowed to default to a single execution. After a successful trial execution the circuit is closed. The transition to closed state means that the trial success window is closed and a new empty rolling failure window is created. Another four executions are then required before the Circuit is opened once more.

I'd like to suggest some revised words in the Spec, together with some more explanation around the CircuitBreaker#testCircuitDefaultSuccessThreshold test and a couple of new Circuit Breaker tests.

Emily-Jiang:
close this issue as the PR #201 was merged.

PR #201, opened 2017-11-21
circuit breaker rolling failure window spec and test (issue 197)

neilgsyoung:


McFoggy:
The `reset of records` looks a bit as an implementation detail of an existing implementation but at least it provides backward compatibility with 1.0 tck tests (which was broken by #191).

Emily-Jiang:
close #197

Issue #185, opened 2017-10-30
Clarify usage of Timeout with Asynchronous and cover it by tests


OndrejM:
The specification mentions that `@Timeout` can be used with `@Asynchronous`, but doesn't explain how it should behave and it's also not covered by tests.

I propose that if both are used and timeout happens, the specification should mandate that the future is canceled at the time the execution takes longer than timeout, not after execution is finished. It would also mean that probably a background thread would need to monitor all async timeout calls and check if they take longer than timeout but that's implementation detail. If `future.get` is called, then it should throw `TimeoutException` at the time when a timeout occurs, not when the execution is finished.

This is to notify the caller of a timeout immediately and prevent delays in case the asynchronous task is waiting on a blocking call.

This should also be covered by tests. Currently, tests only cover using `@Retry` without any other annotations.

Emily-Jiang:
Ondrej,

The spec has the following `[claim ](http://microprofile.io/project/eclipse/microprofile-fault-tolerance/spec/src/main/asciidoc/timeout.asciidoc)`


When @Timeout is used without @Asynchronous, the current thread will be interrupted with a call to Thread.interrupt() on reaching the specified timeout duration. The interruption will only work in certain scenarios. The interruption will not work for the following situations:

    The thread is blocked on blocking I/O (database, file read/write), an exception is thrown only in case of waiting for a NIO channel

    The thread isn’t waiting (CPU intensive task) and isn’t checking for being interrupted

    The thread will catch the interrupted exception (with a general catch block) and will just continue processing, ignoring the interrupt

In the above situations, it is impossible to suspend the execution. The execution thread will finish its process. If the execution takes longer than the specified timeout, the TimeoutException will be thrown and the execution result will be discarded.
Is it still unclear? I agree a separate test should be added to test this.

Emily-Jiang:
close this one as tck and spec are updated.

PR #202, opened 2017-11-22
@Asynchronous and @Timeout spec and test: issue #185

neilgsyoung:

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsyncTimeoutTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsyncTimeoutTest.java
@@ -1,0 +1,231 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2016-2017 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsyncTimeoutTest.java
```diff
@@ -0,0 +1,231 @@
...
+        }
+        catch (InterruptedException e) {
+            throw new AssertionError("testAsyncTimeout: unexpected InterruptedException on future.get()");
+        }
+        catch (TimeoutException e) {
+            throw new AssertionError("testAsyncTimeout: unexpected TimeoutException on future.get()");
+        }
+        end = System.currentTimeMillis();
+
+        duration = end - start;
```
Emily-Jiang: do you want to assert the duration to be less than the timeout value?

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynctimeout/clientserver/AsyncTimeoutClient.java
```diff
@@ -0,0 +1,78 @@
...
+    
+    /**
+     * serviceA is a slow running service that will take 5 seconds in normal operation. Here it is
+     * configured to time out after 2 seconds.
+     */
+    @Timeout(2000)
+    @Asynchronous
+    public Future<Connection> serviceA() throws InterruptedException {
+
+        Thread.sleep(5000);
```
Emily-Jiang: should move the Thread.sleep down to the make connection to represent the service invocation

PR #208, opened 2017-12-07
Gordon pr1930 tweek

hutchig:
The PR is built on top of pull 1930 and tries to take the very valid points from that forwards.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/Bulkhead5RapidRetry12MethodSynchBean.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/Bulkhead5RapidRetry12MethodSynchBean.java
@@ -19,24 +19,30 @@
  *******************************************************************************/
 package org.eclipse.microprofile.fault.tolerance.tck.bulkhead.clientserver;
 
+import java.time.temporal.ChronoUnit;
 import java.util.concurrent.Future;
-
+import javax.enterprise.context.ApplicationScoped;
 import org.eclipse.microprofile.fault.tolerance.tck.bulkhead.Utils;
 import org.eclipse.microprofile.faulttolerance.Bulkhead;
+import org.eclipse.microprofile.faulttolerance.Retry;
+import org.eclipse.microprofile.faulttolerance.exceptions.BulkheadException;
 
 /**
- * A simple method level Semaphore @Bulkhead(10)
+ * A simple method level synchronous @Bulkhead bean that has a retry option, 
+ * with a waitingTaskQueue value that should be ignored.
  * 
- * @author Gordon Hutchison
+ * @author Andrew Pielage
  */
-
-public class BulkheadMethodSemaphore10Bean implements BulkheadTestBackend {
-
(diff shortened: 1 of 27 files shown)


tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/Bulkhead5RapidRetry12MethodSynchBean.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * A simple method level synchronous @Bulkhead bean that has a retry option, 
+ * with a waitingTaskQueue value that should be ignored.
+ * 
+ * @author Andrew Pielage
+ */
+public class Bulkhead5RapidRetry12MethodSynchBean implements BulkheadTestBackend {
+    
+    @Override
+    @ApplicationScoped
```
Emily-Jiang: This is not a producer. Why put ApplicationScoped annotation on?

hutchig: This is not 'new' code. I just renamed it to a more consistent name. The author tag says Andrew Pielage.  If the original ApplicationScvoped came from me IMHO I think it is a good thing..to have some variety and more would be better. We don't have great coverage of scopes (standard or custom) and too many of the test beans have the same scope. For example - do we test that the bulkhead 'in' count is per target bean instance correctly scoped and not per bean class or call site...no, not yet.

Emily-Jiang: It is wrong to have that annotation. I

hutchig: Why not? Can't one have fault tolerance to a AppScoped singleton? Please explain? Sure the test does not exploit it but it is a valid thing to have and thus test. Please explain why not.

Emily-Jiang:
@ApplicationScoped is only used for controlling bean's lifecycle. In this case, the annotation is directly specified on a non-producer method, which does not produce a bean. It is wrong to be there.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/Bulkhead5RapidRetry12MethodSynchBean.java
hutchig: Agh - just spotted it was on the method and not the class and hence a scoping  - apologies. 
 I can only assume that this is a copy and paste error either by Andrew or me. I will remove it with an additional commit tomorrow first thing.

hutchig:
Commit 912cb20 added to the PR removes the method level @ApplicationScoped annotations.

Emily-Jiang:
Thanks @hutchig @Pandrex247 !

Issue #182, opened 2017-10-28
Add more configuration based TCK tests


johnament:
The following behaviors are not tested:

- Having a property set for both the method and the class
- Having a property set in a global fashion that overrides the other class/method annotations
- Property configuration for global
- Bulkhead configuration
- CircuitBreaker configuration
- Timeout configuration
- Exception class conversion (should this be moved to the config spec?)
- TimeUnit conversion (should this be moved to the config spec?)

Emily-Jiang:
Thanks for reporting this @johnament ! Feel free to submit PR if you get to it first.

Emily-Jiang:
@antoinesd can you help with this?

antoinesd:
yes @Emily-Jiang

antoinesd:
Added PR. Test related to converters should be added to config IMO

PR #209, opened 2017-12-15
#microprofile-fault-tolerance-182 Add more configuration based TCK tests

antoinesd:
Solve #182


neilgsyoung:
These are good tests. When I ran them against OpenLiberty I found one test issue in ConfigPropertyGlobalVsClassTest. I think the global config property override should mean that we see 7 retries (8 executions). I'll merge this PR and fix the test under issue #215

PR #237, opened 2018-03-16
issue-109

### LINKED ISSUE #109: Disable individual Fault Tolerance annotation using external config
Based on the current state of Fault Tolerance configuration, It is not possible to disable specific annotation individually or globally. 

Possible solution could be :

1. Introduce an enabled element with default value true in annotation to disable it using external config source.
e.g : `com.acme.test.MyClient/serviceA/CircuitBreaker/enabled=false`
2. The annotation can be disabled via system properties in the naming convention of `<classname>/<methodname>/<annotation>=false`
e.g : `com.acme.test.MyClient/serviceA/CircuitBreaker=false`

antoinesd:
Fix #109 by Adding spec and test

### DIFF
--- a/spec/src/main/asciidoc/configuration.asciidoc
+++ b/spec/src/main/asciidoc/configuration.asciidoc
@@ -152,3 +152,5 @@ For instance the following config will disable bulkhead policy globally:
 `Bulkhead/enabled=false`
 
 Policy will be disabled everywhere ignoring existing policy annotations on methods and classes.
+
+If the above configurations patterns are used with another value than `false` (i.e. `<classname>/<methodname>/<annotation>/enabled=whatever`) non-portable behaviour results.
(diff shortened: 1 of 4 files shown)


spec/src/main/asciidoc/configuration.asciidoc
```diff
@@ -107,6 +107,48 @@ public class MyClient{
...
+
+Policy will be disabled even if the policy is also defined at class level
+
+* Disabling a policy at class level
+
+A policy can be disabled at class level with the following config property and value:
+
+`<classname>/<annotation>/enabled=false`
+
+For instance the following config will disable fallback policy on `com.acme.test.MyClient` class:
```
Emily-Jiang: Fallback is only a method level annotation. You might choose Retry for this explanation.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/disableEnv/DisableAnnotationGloballyTest.java
```diff
@@ -0,0 +1,141 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2016-2017 Contributors to the Eclipse Foundation
```
Emily-Jiang: The copyright year needs to be 2016-2018

antoinesd:
Added mention of config pattern without `enabled=false` being non-portable

antoinesd:
@Emily-Jiang corrected property names and added test for asynchronous policy deactivation

PR #238, opened 2018-03-19
Fix BulkheadAsynchRetryTest#testBulkheadMethodAsynchronousRetry55Trip()

### LINKED ISSUE #232: BulkheadAsynchRetryTest - TestData.check() can be  invoked before async wrappers are executed
In the `BulkheadAsynchRetryTest.testBulkheadMethodAsynchronousRetry55Trip()` the `TestData.check()` can be  invoked before async `BulkheadRapidRetry55MethodAsynchBean.test()` wrappers are executed.

If only one wrapper is executed the test fails with:
> java.lang.AssertionError:  Workers are not in parrallel.  expected [false] but found [true]
	at org.eclipse.microprofile.fault.tolerance.tck.bulkhead.BulkheadAsynchRetryTest.testBulkheadMethodAsynchronousRetry55Trip(BulkheadAsynchRetryTest.java:154)

A simple fix might be to initialize the `TestData` with `CountDownLatch(10)`.

mkouba:
- TestData.check() must not be invoked before async wrappers are executed
- resolves #232

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadAsynchRetryTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadAsynchRetryTest.java
@@ -86,7 +86,7 @@ public class BulkheadAsynchRetryTest extends Arquillian {
     /**
      * This is the Arquillian deploy method that controls the contents of the
      * war that contains all the tests.
-     * 
+     *
      * @return the test war "ftBulkheadAsynchTest.war"
      */
     @Deployment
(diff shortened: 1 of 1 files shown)


Emily-Jiang:
@mkouba did you sign off the commit? Your PR failed ip validation check.

hutchig:
There can be another problem with this testcase in a highly stressed test system...
Although the Retry annotation has a very very short period of time (1microsecond) and
will retry only 10 times, if the system takes in the order of 1 second to do these 10 retries
it can give enough time to allow one or more of the initial workers to 'complete' and get
out of the Bulkhead - thus letting 'in' the 11th task - which will cause the test to fail.
I will raise a seperate issue if none already covers it.
(See https://github.com/OpenLiberty/open-liberty/issues/2994 for further data).

mkouba:
@hutchig thanks! Speaking of "1 microsecond delay" - note that a sleep interval of value less than 15 ms is questionable since some operating systems do not provide such precision.

hutchig:
Yes you are absolutely right @mkouba it is not a realistic number - this was more a case of the smallest valid number/unit rather than a sensible one.  For the purposes of the test I wanted the 10 retries to occur 'as fast as possible' and the fastest thing to do (with the ability to justify) was to pick the edge case of a smaller than matters number - it does not make real-world sense - you are right :-).

hutchig:
See also #253

PR #241, opened 2018-04-05
#microprofile-fault-tolerance-220 Add support for exponential backoff retry

antoinesd:

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java
+++ b/api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017 Contributors to the Eclipse Foundation
+ * Copyright (c) 2017-2018 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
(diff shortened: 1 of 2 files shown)


antoinesd:
DO NOT MERGE. TCK should be amended when syntax is validated

api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java
```diff
@@ -109,4 +110,27 @@
...
+     *
+     * When this value is greater than 1, backoff retry is activated.
+     * {@link #delay()} between retries will be increased exponentially with this value until
+     * the delay reaches {@link #backoffMaxDelay()}
+     * Value must be equals greater than 1
+     *
+     * @return the backoff factor
+     */
+    @Nonbinding
+    double backoffDelayFactor() default 1;
```
Emily-Jiang: Can we make the default to 2 to match failsafe default?

api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java
```diff
@@ -109,4 +110,27 @@
...
+
+
+    /**
+     *
+     * @return the maximum delay value. No limit if it's set to 0. Value is ignored if {@link #backoffDelayFactor()} is 1
+     */
+    @Nonbinding
+    long backoffMaxDelay() default 0;
```
Emily-Jiang: maxDelay must be great than delay. Otherwise, IllegalArgumentException should be thrown.

antoinesd: how would you you deactivate backoff retry then?

Emily-Jiang: As discussed on the previous call, this method is not needed. We have delay specified already. @antoinesd will you be able to update this? If not, I can submit a PR.

Issue #234, opened 2018-03-09
Add metrics status automatically for FT


Emily-Jiang:
Based on the feedback we got from the community, we need to display the metrics for FT actions.

Look at each annotation in turn:
1. Retry - it might be useful to display the time of retries. Maybe @Counter can be automatically added.

2. Timed - ideally it is good to display how long each method takes to run. @Timer can be used.
3. CircuitBreaker - good to list the failure rate, circuit open time, request ranges etc. Histogram can be used?

4. Bulkhead - nice to list the waiting requests, how long it waited etc, @Gauge can be used?

Azquelt:
I've tried to follow prometheus best practices, particularly on [instrumentation][1] and [naming of metrics][2].

To make sense of the metrics I've suggested, it should be noted that prometheus will collect a snapshot of all of the metrics on a regular basis (default every 15 seconds) and provides lots of facilities for comparing and processing collected metric data.

For example, while a counter of the number of calls made to a method isn't useful on its own, it can be used in the following ways:
 - it can be fed into the `rate()` function to produce a graph of the rolling average rate of calls per second over the last minute
 - if you also have a count of the calls which failed, you could also divide the failures per second by the total calls per second to see what proportion of calls to the method are failing over time

# Proposed metrics

For `@Retry`:
 - Counter of calls (`${name}_calls_total`)
 - Counter of retries (`${name}_retries_total`)
 - Counter of calls which failed despite retrying (`${name}_failures_total`)

Rationale: 
 - User knows that they're calling an unreliable API so it may be worthwhile to monitor how unreliable it is so that they can see when it's approaching the point where it may impact the system.

For `@Timeout`:
 - Histogram of execution times (`${name}_duration_seconds`)
 - Counter of calls (`${name}_calls_total`)
 - Counter of timeouts (`${name}_timeouts_total`)

Rationale:
 - User is concerned about how long the method takes to return, so we should monitor this

For `@CircuitBreaker`:
 - Gauge of breaker state (`${name}_breaker_open`)
   - should we include a % time open?
 - counter of calls (`${name}_calls_total`)
 - counter of failures (`${name}_failures_total`)
 - counter of failures caused by the breaker being open (`${name}_breaker_open_failures_total`)

Rationale:
 - User is concerned about API breaking to the point that they'd just rather not use it until it's working again. We should monitor how close we're getting to that threshold and how often we're crossing it.

For `@Bulkhead`:
[...]

PR #257, opened 2018-04-24
Integration with Microprofile Metrics

Azquelt:
For  #234

### DIFF
--- a/spec/src/main/asciidoc/configuration.asciidoc
+++ b/spec/src/main/asciidoc/configuration.asciidoc
@@ -154,3 +154,12 @@ For instance the following config will disable bulkhead policy globally:
 Policy will be disabled everywhere ignoring existing policy annotations on methods and classes.
 
 If the above configurations patterns are used with another value than `false` (i.e. `<classname>/<methodname>/<annotation>/enabled=whatever`) non-portable behaviour results.
+
+=== Configuring Metrics Integration
+
+The integration with MicroProfile Metrics can be disabled by setting a config property named `MP_Fault_Tolerance_Metrics_Enabled` to the value `false`.
+If this property is absent or set to `true` then the integration with MicroProfile Metrics will be enabled and the metrics listed earlier in this specification
+will be added automatically for every method annotated with a `@Retry`, `@Timeout`, `@CircuitBreaker` or `@Bulkhead` annotation.
+
+In order to prevent from any unexpected behaviours, the property `MP_Fault_Tolerance_Metrics_Enabled` will only be read when the application starts.
+Any dynamic changes afterwards will be ignored until the application is restarted.
--- a/spec/src/main/asciidoc/metrics.asciidoc
+++ b/spec/src/main/asciidoc/metrics.asciidoc
@@ -1,0 +1,237 @@
+//
+// Copyright (c) 2018 Contributors to the Eclipse Foundation
+//
+// See the NOTICE file(s) distributed with this work for additional
+// information regarding copyright ownership.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// You may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// Contributors:
+// Andrew Rouse
+
+== Integration with Microprofile Metrics
+
+When Microprofile Fault Tolerance and Microprofile Metrics are used together, metrics are automatically added for each of
+the methods annotated with a `@Retry`, `@Timeout`, `@CircuitBreaker` or `@Bulkhead` annotation.
+
+=== Names
+
+The automatically added metrics follow a consistent pattern which includes the fully qualified name of the annotated method.
+In the tables below, the placeholder `<name>` should be replaced by the fully qualified method name.
+
+=== Metrics added for `@Retry`, `@Timeout`, `@CircuitBreaker` and `@Bulkhead`
+
+Implementations must ensure that if any of these annotations are present on a method, then the following metrics are added only once for that method.
+
+[cols="8,3,4,9"]
+|===
+| Name | Type | Unit | Description
+
+|`ft.<name>.invocations.total`
+| Counter | None
+| The number of times the method was called
+
+|`ft.<name>.invocations.failed.total`
+| Counter | None
+| The number of times the method was called and, after all Fault Tolerance actions had been processed, threw a `Throwable`
+|===
+
+=== Metrics added for `@Retry`
+
+[cols="8,3,4,9"]
+|===
+| Name | Type | Unit | Description
+
+|`ft.<name>.retry.callsSucceededNotRetried.total`
+| Counter | None
+| The number of times the method was called and succeeded without retrying
+
+|`ft.<name>.retry.callsSucceededRetried.total`
+| Counter | None
+| The number of times the method was called and succeeded after retrying at least once
+
+|`ft.<name>.retry.callsFailed.total`
+| Counter | None
+| The number of times the method was called and ultimately failed after retrying
+
+|`ft.<name>.retry.retries.total`
+| Counter | None
+| The total number of times the method was retried
+|===
+
+Note that the sum of `ft.<name>.retry.callsSucceededNotRetried.total`, `ft.<name>.retry.callsSucceededRetried.total` and `ft.<name>.retry.callsFailed.total` will give the total number of calls for which the Retry logic was run.
+
+=== Metrics added for `@Timeout`
+
+[cols="8,3,4,9"]
+|===
+| Name | Type | Unit | Description
+
+|`ft.<name>.timeout.executionDuration`
+| Histogram | Nanoseconds
+| Histogram of execution times for the method
+
+|`ft.<name>.timeout.callsTimedOut.total`
+| Counter | None
+| The number of times the method timed out
+
+|`ft.<name>.timeout.callsNotTimedOut.total`
+| Counter | None
+| The number of times the method completed without timing out
+|===
+
+Note that the sum of `ft.<name>.timeout.callsTimedOut.total` and `ft.<name>.timeout.callsNotTimedOut.total` will give the total number of calls for which the Timeout logic was run. This may be larger than the total number of invocations of the method it's also annotated with `@Retry` because the Timeout logic is applied to each Retry attempt, not to the whole method invocation time.
+
+=== Metrics added for `@CircuitBreaker`
+
+[cols="8,3,4,9"]
+|===
+| Name | Type | Unit | Description
+
+|`ft.<name>.circuitbreaker.callsSucceeded.total`
+| Counter | None
+| Number of calls allowed to run by the circuit breaker that returned successfully
+
+|`ft.<name>.circuitbreaker.callsFailed.total`
+| Counter | None
+| Number of calls allowed to run by the circuit breaker that then failed
+
+|`ft.<name>.circuitbreaker.callsPrevented.total`
+| Counter | None
+| Number of calls prevented from running by an open circuit breaker
+
+|`ft.<name>.circuitbreaker.open.total`
+| Gauge<Long> | Nanoseconds
+| Amount of time the circuit breaker has spent in open state
+
+|`ft.<name>.circuitbreaker.halfOpen.total`
+| Gauge<Long> | Nanoseconds
+| Amount of time the circuit breaker has spent in half-open state
+
+|`ft.<name>.circuitbreaker.closed.total`
+| Gauge<Long> | Nanoseconds
+| Amount of time the circuit breaker has spent in closed state
+
+|`ft.<name>.circuitbreaker.opened.total`
+| Counter | None
+| Number of times the circuit breaker has moved from closed state to open state
+|===
+
+Note that the sum of `ft.<name>.circuitbreaker.callsSucceeded.total`, `ft.<name>.circuitbreaker.callsFailed.total` and `ft.<name>.circuitbreaker.callsPrevented.total` will give the total number of calls for which the Bulkhead logic was run.
+
+Similarly, the sum of `ft.<name>.circuitbreaker.open.total`, `ft.<name>.circuitbreaker.halfOpen.total` and `ft.<name>.circuitbreaker.closed.total` will give the total time that the circuit breaker has been active for.
+
+
+=== Metrics added for `@Bulkhead`
+
+[cols="8,3,4,9"]
+|===
+| Name | Type | Unit | Description
+
+|`ft.<name>.bulkhead.concurrentExecutions`
+| Gauge<Long> | None
+| Number of currently running executions
+
+|`ft.<name>.bulkhead.callsAccepted.total`
+| Counter | None
+| Number of calls accepted by the bulkhead
+
+|`ft.<name>.bulkhead.callsRejected.total`
+| Counter | None
+| Number of calls rejected by the bulkhead
+
+|`ft.<name>.bulkhead.executionDuration`
+| Histogram | Nanoseconds
+| Histogram of method execution times. This does not include any time spent waiting in the bulkhead queue.
+
+|`ft.<name>.bulkhead.waitingQueue.population`*
+| Gauge<Long> | None
+| Number of executions currently waiting in the queue
+
+|`ft.<name>.bulkhead.waiting.duration`*
+| Histogram | Nanoseconds
+| Histogram of the time executions spend waiting in the queue
+|===
+
+*Only added if the method is also annotated with `@Asynchronous`
+
+== Notes
+
+Metrics added by this specification will appear as application metrics for the application which uses the Fault Tolerance annotations.
+
+Future versions of this specification may change the definitions of the metrics which are added to take advantage of
+enhancements in the MicroProfile Metrics specification.
+
+If more than one annotation is applied to a method, the metrics associated with each annotation will be added for that method.
+
+All of the counters are monotonic and count the number of events which occurred since the application started.
+It is expected that these counters will be sampled regularly by monitoring software which is then able to compute deltas
+or moving averages from the gathered samples.
+
+=== Annotation Example
+
+[source, java]
+----
+package com.exmaple;
+
+@Timeout(1000)
+public class MyClass {
+
+    @Retry
+    public void doWork() {
+        // work
+    }
+
+}
+----
+
+This class would result in the following metrics being added.
+
+* `ft.com.example.MyClass.doWork.invocations.total`
+* `ft.com.example.MyClass.doWork.invocations.failed`
+* `ft.com.example.MyClass.doWork.retry.callsSucceededNotRetried.total`
+* `ft.com.example.MyClass.doWork.retry.callsSucceededRetried.total`
+* `ft.com.example.MyClass.doWork.retry.callsFailed.total`
+* `ft.com.example.MyClass.doWork.retry.retries.total`
+* `ft.com.example.MyClass.doWork.timeout.executionDuration`
+* `ft.com.example.MyClass.doWork.timeout.callsTimedOut.total`
+* `ft.com.example.MyClass.doWork.timeout.callsNotTimedOut.total`
+
+Now imagine the `doWork()` method is called and the invocation goes like this:
+* On the first attempt, the invocation takes more than 1000ms and times out
+* On the second attempt, something goes wrong and the method throws an `IOException`
+* On the third attempt, the method returns successfully and the result of this attempt is returned to the user
+
+After this sequence, the value of these metrics would be as follows:
+
+`ft.com.example.MyClass.doWork.invocations.total = 1` +
+The method has been called once.
+
+`ft.com.example.MyClass.doWork.invocations.failed = 0` +
+No exceptions were propagated back to the caller.
+
+`ft.com.example.MyClass.doWork.retry.callsSucceededNotRetried.total = 0` +
+`ft.com.example.MyClass.doWork.retry.callsSucceededRetried.total = 1` +
+`ft.com.example.MyClass.doWork.retry.callsFailed.total = 0` +
+Only one call was made, and it succeeded after some retries.
+
+`ft.com.example.MyClass.doWork.retry.retries.total = 2` +
+Two retries were made during the invocation.
+
+`ft.com.example.MyClass.doWork.timeout.executionDuration` +
+The `Histogram` will have been updated with the length of time taken for each attempt. It will show a count of `3` and will have calculated averages and percentiles from the execution times.
+
+`ft.com.example.MyClass.doWork.timeout.callsTimedOut.total = 1` +
+One of the attempts timed out.
+
+`ft.com.example.MyClass.doWork.timeout.callsNotTimedOut.total = 2` +
+Two of the attempts did not time out.
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/BulkheadMetricBean.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/BulkheadMetricBean.java
@@ -1,0 +1,119 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+package org.eclipse.microprofile.fault.tolerance.tck.metrics;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import javax.enterprise.context.ApplicationScoped;
+
+import org.eclipse.microprofile.faulttolerance.Asynchronous;
+import org.eclipse.microprofile.faulttolerance.Bulkhead;
+
+@ApplicationScoped
+public class BulkheadMetricBean {
+    
+    // The Atomicness is not important, this is just an integer holder
+    private AtomicInteger executionCount = new AtomicInteger(0);
+    
+    private static final long WAIT_TIMEOUT = 3L * 1000 * 1_000_000; 
+    
+    /**
+     * Wait for {@code future} to be completed
+     */
+    @Bulkhead(2)
+    public void waitFor(Future<?> future) {
+        doWaitFor(future);
+    }
+
+    /**
+     * Separate waitFor method for testing execution time histogram
+     */
+    @Bulkhead(2)
+    public void waitForHistogram(Future<?> future) {
+        doWaitFor(future);
+    }
+
+    /**
+     * WaitFor method for testing async calls
+     */
+    @Asynchronous
+    @Bulkhead(value = 2, waitingTaskQueue = 2)
+    public Future<Void> waitForAsync(Future<?> future) {
+        doWaitFor(future);
+        return CompletableFuture.completedFuture(null);
+    }
+    
+    private void doWaitFor(Future<?> future) {
+        try {
+            executionStarted();
+            future.get();
+        }
+        catch (InterruptedException e) {
+            throw new RuntimeException("Test was interrupted", e);
+        }
+        catch (ExecutionException e) {
+            throw new RuntimeException("Passed Future threw an exception", e);
+        }
+        finally {
+            executionEnded();
+        }
+    }
+    
+    /**
+     * Wait for the given number of method executions to be running in this bean
+     * <p>
+     * This method will wait three seconds before returning an exception
+     */
+    public void waitForRunningExecutions(int executions) {
+        synchronized (executionCount) {
+            long startTime = System.nanoTime();
+            try {
+                while (executionCount.get() != executions && (System.nanoTime() - startTime) < WAIT_TIMEOUT) {
+                    executionCount.wait(500);
+                }
+            }
+            catch (InterruptedException e) {
+                // Stop waiting
+            }
+            
+            if (executionCount.get() != executions) {
+                // Timed out waiting
+                throw new RuntimeException("Timed out waiting for executions to start, expected " + executions + " but there were " + executionCount);
+            }
+        }
+    }
+    
+    private void executionStarted() {
+        synchronized (executionCount) {
+            executionCount.incrementAndGet();
+            executionCount.notifyAll();
+        }
+    }
+    
+    private void executionEnded() {
+        synchronized (executionCount) {
+            executionCount.decrementAndGet();
+            executionCount.notifyAll();
+        }
+    }
+}
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/BulkheadMetricTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/BulkheadMetricTest.java
@@ -1,0 +1,230 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+package org.eclipse.microprofile.fault.tolerance.tck.metrics;
+
+import static org.eclipse.microprofile.fault.tolerance.tck.metrics.util.Exceptions.expectBulkheadException;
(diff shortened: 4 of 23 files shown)


Azquelt:
Rebased, conflict resolved, squashed and signoffs added.

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// Contributors:
+// Andrew Rouse
+
+== Integration with Microprofile Metrics
+
+When Microprofile Fault Tolerance and Microprofile Metrics are used together, metrics are automatically added for each of
+the methods annotated with a Fault Tolerance annotation.
```
Emily-Jiang: except Asynchronous annotation

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/CircuitBreakerMetricTest.java
```diff
@@ -0,0 +1,104 @@
...
+        assertThat(m.getCircuitBreakerOpenedDelta(), is(1L));
+
+        // Second failure, causes circuit to open
+        expectTestException(() -> cbBean.doWork(Result.FAIL));
+        
+        assertThat(m.getCircuitBreakerState(), is(2));
+        assertThat(m.getCircuitBreakerCallsSuccessDelta(), is(0L));
+        assertThat(m.getCircuitBreakerCallsFailureDelta(), is(2L));
+        assertThat(m.getCircuitBreakerCallsCircuitOpenDelta(), is(0L));
+        assertThat(m.getCircuitBreakerOpenedDelta(), is(1L));
```
Emily-Jiang: I think the 3rd request will trigger the circuit to open. The circuit will only be operated on a request though.

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+
+When Microprofile Fault Tolerance and Microprofile Metrics are used together, metrics are automatically added for each of
+the methods annotated with a `@Retry`, `@Timeout`, `@CircuitBreaker` or `@Bulkhead` annotation.
+
+=== Names
+
+The automatically added metrics follow a consistent pattern which includes the fully qualified name of the annotated method.
+In the tables below, the placeholder `<name>` should be replaced by the fully qualified method name with all `.` characters replaced with `_`.
+
+=== Metrics added for `@Retry`
```
OndrejM: I think that also a metric for average number of retries per call would be useful to monitor stability of the call.

OndrejM: With name `ft_<name>_retry_retries_average`

OndrejM: Or the number of retries per call could be a histogram, with min, max and mean number of retries per call.

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+
+[cols="8,3,4,9"]
+|===
+| Name | Type | Unit | Description
+
+|`ft_<name>_circuitbreaker_state`
+| Gauge<Integer> | None
+| Current state of the circuit breaker. `0` = Closed, `1` = Half-open, `2` = Open.
+
+|`ft_<name>_circuitbreaker_calls_success_total`
```
OndrejM: During hangout, we agreed to rename to `ft_<name>_circuitbreaker_calls_succeeded_total`

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+| Counter | None
+| Number of calls allowed to run by the circuit breaker that returned successfully
+
+|`ft_<name>_circuitbreaker_calls_failure_total`
```
OndrejM: During hangout, we agreed to rename to `ft_<name>_circuitbreaker_calls_failed_total`

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+| Counter | None
+| Number of calls allowed to run by the circuit breaker that then failed
+
+|`ft_<name>_circuitbreaker_calls_circuit_open_total`
```
OndrejM: During hangout, we agreed to rename to `ft_<name>_circuitbreaker_calls_prevented_total`

spec/src/main/asciidoc/metrics.asciidoc
OndrejM: It should be documented that total of succeeded + failed + prevented = sum of all calls in total

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+
+[cols="8,3,4,9"]
+|===
+| Name | Type | Unit | Description
+
+|`ft_<name>_timeout_duration`
+| Histogram | Nanoseconds
+| Histogram of execution times for the method
+
+|`ft_<name>_timeout_calls_timeout_true_total`
```
OndrejM: During hangout, we agreed to rename to `ft_<name>_timeout_calls_timed_out_total`

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+| Counter | None
+| The number of times the method timed out
+
+|`ft_<name>_timeout_calls_timeout_false_total`
```
OndrejM: During hangout, we agreed to rename to `ft_<name>_timeout_calls_not_timed_out_total`

spec/src/main/asciidoc/metrics.asciidoc
OndrejM: We should document that `timeout_calls_timed_out` + `timeout_calls_not_timed_out` = total sum of calls

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+| Number of times the circuit breaker has moved from closed state to open state
+|===
+
+=== Metrics added for `@Bulkhead`
+
+[cols="8,3,4,9"]
+|===
+| Name | Type | Unit | Description
+
+|`ft_<name>_bulkhead_concurrent_executions`
```
OndrejM: Maybe rename to `ft_<name>_bulkhead_concurrent_executions_current` - suffix with `current` to indicate that it's an actual state and not a counter or total.

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+
+[cols="8,3,4,9"]
+|===
+| Name | Type | Unit | Description
+
+|`ft_<name>_retry_calls_success_immediate_total`
+| Counter | None
+| The number of times the method was called and succeeded without retrying
+
+|`ft_<name>_retry_calls_success_retry_total`
```
OndrejM: To be in line with what was suggested for circuitbreaker, this should be renamed to `ft_<name>_retry_calls_succeeded_retried_total`

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+| Counter | None
+| The number of times the method was called and succeeded after retrying at least once
+
+|`ft_<name>_retry_calls_failure_total`
```
OndrejM: To be in line with what was suggested for circuitbreaker, this should be renamed to `ft_<name>_retry_calls_failed_total`

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+The automatically added metrics follow a consistent pattern which includes the fully qualified name of the annotated method.
+In the tables below, the placeholder `<name>` should be replaced by the fully qualified method name with all `.` characters replaced with `_`.
+
+=== Metrics added for `@Retry`
+
+[cols="8,3,4,9"]
+|===
+| Name | Type | Unit | Description
+
+|`ft_<name>_retry_calls_success_immediate_total`
```
OndrejM: To be in line with what was suggested for circuitbreaker, this should be renamed to `ft_<name>_retry_calls_succeeded_immediately_total`. 

But I suggest rather the following: `ft_<name>_retry_calls_succeeded_not_retried_total`

spec/src/main/asciidoc/metrics.asciidoc
OndrejM: Should document that `calls failed` + `calls succeded immediately` + `calls succeeded with retry` = total calls

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+// Andrew Rouse
+
+== Integration with Microprofile Metrics
+
+When Microprofile Fault Tolerance and Microprofile Metrics are used together, metrics are automatically added for each of
+the methods annotated with a `@Retry`, `@Timeout`, `@CircuitBreaker` or `@Bulkhead` annotation.
+
+=== Names
+
+The automatically added metrics follow a consistent pattern which includes the fully qualified name of the annotated method.
```
OndrejM: A question - if multiple apps are deployed and some of them use the same class with the same FT annotations, would that mean there's only one combined metric for both annotated methods in both apps?

AFAIK, FT Metrics 1.0 doesn't allow exposing the same metric multiple times for each application. It's possible to add tags to a metric but that wouldn't help if 2 metrics would have the same name.

There's an issue open to add support for metrics coming from different applications: https://github.com/eclipse/microprofile-metrics/issues/65. 

We should at least specify that FT metrics should be exposed in the same way as application-defined metrics. Later, when the Metrics spec comes to a solution, FT metrics from different apps would be handled automatically.

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+    public void doWork() {
+        // work
+    }
+
+}
+----
+
+This class would result in the following metrics being added.
+
+* `ft_com_example_MyClass_doWork_retry_calls_success_immediate_total`
```
OndrejM: Fix this example after metrics are renamed according to the comments.

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,177 @@
...
+
+|`ft_<name>_circuitbreaker_calls_failed_total`
+| Counter | None
+| Number of calls allowed to run by the circuit breaker that then failed
+
+|`ft_<name>_circuitbreaker_calls_prevented_total`
+| Counter | None
+| Number of calls prevented from running by an open circuit breaker
+
+|`ft_<name>_circuitbreaker_accumulated_open`
```
OndrejM: I suggest to rename to ft_<name>_circuitbreaker_open_time_total so that "total" is at the end and "open" is closer to the beginning. Also, I think that "time total" is more descriptive than "accumulated"

Apply same for half open and closed

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,177 @@
...
+| The number of times the method completed without timing out
+|===
+
+=== Metrics added for `@CircuitBreaker`
+
+[cols="8,3,4,9"]
+|===
+| Name | Type | Unit | Description
+
+|`ft_<name>_circuitbreaker_calls_succeeded_total`
```
OndrejM: It should be documented that total of succeeded + failed + prevented = sum of all calls in total

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,177 @@
...
+
+[cols="8,3,4,9"]
+|===
+| Name | Type | Unit | Description
+
+|`ft_<name>_timeout_execution_duration`
+| Histogram | Nanoseconds
+| Histogram of execution times for the method
+
+|`ft_<name>_timeout_calls_timed_out_total`
```
OndrejM: We should document that timeout_calls_timed_out + timeout_calls_not_timed_out = total sum of calls

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,177 @@
...
+The automatically added metrics follow a consistent pattern which includes the fully qualified name of the annotated method.
+In the tables below, the placeholder `<name>` should be replaced by the fully qualified method name with all `.` characters replaced with `_`.
+
+=== Metrics added for `@Retry`
+
+[cols="8,3,4,9"]
+|===
+| Name | Type | Unit | Description
+
+|`ft_<name>_retry_calls_success_no_retry_total`
```
OndrejM: To be in line with what was suggested for circuitbreaker, this should be renamed to `ft_<name>_retry_calls_succeeded_not_retried_total`

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,177 @@
...
+| Counter | None
+| The number of times the method was called and succeeded without retrying
+
+|`ft_<name>_retry_calls_success_retry_total`
```
OndrejM: To be in line with what was suggested for circuitbreaker, this should be renamed to `ft_<name>_retry_calls_succeeded_retried_total`

spec/src/main/asciidoc/configuration.asciidoc
```diff
@@ -154,3 +154,9 @@ For instance the following config will disable bulkhead policy globally:
 Policy will be disabled everywhere ignoring existing policy annotations on methods and classes.
 
 If the above configurations patterns are used with another value than `false` (i.e. `<classname>/<methodname>/<annotation>/enabled=whatever`) non-portable behaviour results.
+
+=== Configuring Metrics Integration
+
+The integration with MicroProfile Metrics can be disabled by setting a config property named `MP_Fault_Tolerance_Metrics_Enabled` to the value `false`.
```
OndrejM: Since MP configuration values can be changed during runtime it should be specified what happens if the value changes. I guess any change during runtime wouldn't have any effect but it's worth to clarify this.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/BulkheadMetricBean.java
```diff
@@ -0,0 +1,64 @@
...
+import javax.enterprise.context.ApplicationScoped;
+
+import org.eclipse.microprofile.faulttolerance.Asynchronous;
+import org.eclipse.microprofile.faulttolerance.Bulkhead;
+
+@ApplicationScoped
+public class BulkheadMetricBean {
+
+    @Bulkhead(2)
+    public void waitFor(long milliseconds) {
```
OndrejM: In tests, it's better to control the length of execution. Unnecessary sleeps make the tests unreliable and long to run. So it's better to pass a future instead of timeout to have complete control over the timing, such as:

```
public void waitFor(Future condition) {
    condition.get();
}
```

and call it with:

```
CompletableFuture cf = new CompletableFuture();
waitFor(cf);
// do something that should happen meanwhile
cf.complete();
```

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/BulkheadMetricBean.java
```diff
@@ -0,0 +1,64 @@
...
+    public void waitForHistogram(long milliseconds) {
+        doWaitFor(milliseconds);
+    }
+
+    /**
+     * WaitFor method for testing async calls
+     */
+    @Asynchronous
+    @Bulkhead(value = 2, waitingTaskQueue = 2)
+    public Future<Void> waitForAsync(long milliseconds) {
```
OndrejM: The same as above, better to use future instead of timeout.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/BulkheadMetricTest.java
```diff
@@ -0,0 +1,158 @@
...
+    public void bulkheadMetricTest() throws InterruptedException, ExecutionException {
+        MetricGetter m = new MetricGetter(BulkheadMetricBean.class, "waitFor");
+        m.baselineCounters();
+        
+        Future<?> f1 = async.run(() -> bulkheadBean.waitFor(3000));
+        Future<?> f2 = async.run(() -> bulkheadBean.waitFor(3000));
+        
+        assertThat("concurrent executions", m.getBulkheadConcurrentExecutions().get(), is(2L));
+        
+        f1.get();
```
OndrejM: If waitFor is rewritten using a completable future, it should complete the future passed to waitFor earlier

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/CircuitBreakerMetricTest.java
```diff
@@ -0,0 +1,99 @@
...
+        return war;
+    }
+    
+    @Inject private CircuitBreakerMetricBean cbBean;
+    
+    @BeforeTest
+    public void closeTheCircuit() throws Exception {
+        // Assume the circuit is open
+        // Wait for it to half close, then close it with passing work
+        Thread.sleep(1500);
```
OndrejM: Could be improved to sleep only if the circuit is really open. The state can be retrieved from `MetricGetter`.

spec/src/main/asciidoc/metrics.asciidoc
Azquelt: You should be able to calculate the average retries per call from the other metrics since you can compute the total number of calls and we also expose the total number of retries.

Prometheus makes it fairly easy to do this sort of calculation and graph the results over time.

spec/src/main/asciidoc/metrics.asciidoc
Azquelt: Yes, good point.

spec/src/main/asciidoc/metrics.asciidoc
Azquelt: Yes, this is a good point. I've already added a note to say that the metric names may change in response to enhancements and changes in the metrics spec.

It would also be good to call out that they'll appear in the same way as application-defined metrics.

As well as the problem of metrics from two applications clashing, we've also got a similar problem if a class has two methods with the same name and different arguments. That's on my list to bring up at tomorrow's meeting.

spec/src/main/asciidoc/metrics.asciidoc
Azquelt: This is not strictly true.

If a method is annotated with `@Retry` and `@Timeout`, we may end up attempting to run the method multiple times. Each of those attempts can either time out or not and that should be reflected in the metrics.

spec/src/main/asciidoc/configuration.asciidoc
Azquelt: Good point, I will add this.

spec/src/main/asciidoc/metrics.asciidoc
Azquelt: I agree the metric names should be consistent and I'm fine with changing them all to the past tense, though I'll raise it at the meeting tomorrow before I go through and change them all.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/BulkheadMetricBean.java
Azquelt: Ah, that's a much better way of controlling the execution order. I'll update the test to use this technique.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/CircuitBreakerMetricTest.java
Azquelt: That would have been true, although we just removed that metric in the meeting this afternoon because it would be very easy to create a misleading graph unless it was sampled very rapidly.

We could possibly check the accumulated time metrics instead, as long as we were careful not to hang the test if those metrics were broken in the implementation under test.

Azquelt:
Thanks for your comments @OndrejM, they were really helpful, I will work on the changes tomorrow.

spec/src/main/asciidoc/metrics.asciidoc
Emily-Jiang: `ft_<name>_circuitbreaker_time_open` prometheus output will append `_seconds`

spec/src/main/asciidoc/metrics.asciidoc
Emily-Jiang: Decided not to do this.

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,177 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// Contributors:
+// Andrew Rouse
+
+== Integration with Microprofile Metrics
```
OndrejM: As we clarified on the hangout that all Counters are monotonic and we're going to remove `_total` suffix we need to state this also in the spec. The Metrics counters can be monotonic but can also be decremented, see [Counted](https://github.com/eclipse/microprofile-metrics/blob/master/api/src/main/java/org/eclipse/microprofile/metrics/annotation/Counted.java#L105).

I suggest to replace "Counter" in every metric type with "Monotonic Counter"

OndrejM:
This issue suggests that adding `_total` suffix isn't part of MP Metrics: https://github.com/eclipse/microprofile-metrics/issues/137

There's another issue to add it, but it's unresolved: https://github.com/eclipse/microprofile-metrics/issues/172

OndrejM:
So I would leave the `_total` suffix in all counter metrics, because MP metrics doesn't add it for prometheus exporter. As Prometheus convention is to end counters with `_total` it's better to have it there.

spec/src/main/asciidoc/metrics.asciidoc
Azquelt: I've added this in the `Notes` section at the bottom of the file.

Azquelt:
I've pushed the latest changes to the spec but haven't updated the TCK yet.

Things I know are outstanding:
* Should we be using dots to separate name components rather than underscores?
* Should all metrics be registered at application startup?
* How should we handle name clashes? (either between classes in different applications, or between methods on the same class)

OndrejM:
I suggest merging the PR as is now and continue the discussion in the issue #234 followed by another PR.

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -88,15 +107,15 @@ In the tables below, the placeholder `<name>` should be replaced by the fully qu
 | Counter | None
 | Number of calls prevented from running by an open circuit breaker
 
-|`ft_<name>_circuitbreaker_accumulated_open`
+|`ft_<name>_circuitbreaker_time_open`
```
OndrejM: Since we're going to leave the `_total` suffix for counters as Metrics doesn't add it, this should also have the `_total` suffix: `ft_<name>_circuitbreaker_time_open_total`

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -88,15 +107,15 @@ In the tables below, the placeholder `<name>` should be replaced by the fully qu
...
 | Gauge<Long> | Nanoseconds
 | Amount of time the circuit breaker has spent in open state
 
-|`ft_<name>_circuitbreaker_accumulated_half_open`
+|`ft_<name>_circuitbreaker_time_half_open`
```
OndrejM: Add `_total` suffix

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -88,15 +107,15 @@ In the tables below, the placeholder `<name>` should be replaced by the fully qu
...
 | Gauge<Long> | Nanoseconds
 | Amount of time the circuit breaker has spent in half-open state
 
-|`ft_<name>_circuitbreaker_accumulated_closed`
+|`ft_<name>_circuitbreaker_time_closed`
```
OndrejM: Add `_total` suffix

Azquelt:
@OndrejM I've just pushed a couple of changes which were requested and updated the names of metrics in the TCK to match the spec. I'm happy for this to be merged and to continue further discussion in #234.

Emily-Jiang:
@Azquelt 

- Should we be using dots to separate name components rather than underscores?
use dot.
- Should all metrics be registered at application startup?
yes
- How should we handle name clashes? (either between classes in different applications, or between methods on the same class)
we should document this limitation and when tag is introduced, these limitation will be removed.

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,235 @@
...
+
+When Microprofile Fault Tolerance and Microprofile Metrics are used together, metrics are automatically added for each of
+the methods annotated with a `@Retry`, `@Timeout`, `@CircuitBreaker` or `@Bulkhead` annotation.
+
+=== Names
+
+The automatically added metrics follow a consistent pattern which includes the fully qualified name of the annotated method.
+In the tables below, the placeholder `<name>` should be replaced by the fully qualified method name with all `.` characters replaced with `_`.
+
+=== Metrics added for `@Retry`, `@Timeout`, `@CircuitBreaker` and `@Bulkhead`
```
donbourne: You will run into naming collisions if you try to reuse the same metric names for different FT annotations.  For example, if `@Retry` and `@Timeout` both declare a metric called `ft_<name>_invocations_total` they will either fail at metric registration time (if reusable is set to false) or double count invocations (if reusable is set to true).

Azquelt: Understood. The intent here was to say that if any of these annotations are present on the method, then the following metrics are added. If two or more annotations are present, then the metrics should still only be added once.

I'll update the wording to clarify this.

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,235 @@
...
+
+== Integration with Microprofile Metrics
+
+When Microprofile Fault Tolerance and Microprofile Metrics are used together, metrics are automatically added for each of
+the methods annotated with a `@Retry`, `@Timeout`, `@CircuitBreaker` or `@Bulkhead` annotation.
+
+=== Names
+
+The automatically added metrics follow a consistent pattern which includes the fully qualified name of the annotated method.
+In the tables below, the placeholder `<name>` should be replaced by the fully qualified method name with all `.` characters replaced with `_`.
```
donbourne: In the base metrics we use `.`s in the metric names.  For prometheus format we automatically convert the `.` characters to `_`, but in json format we leave the `.` characters as is.  mpMetrics is intended to be used by more than just prometheus.  The dot notation is preferred for Graphite.  dropwizard metrics natively uses dots between parts of a name.  So, the net of that is we may want to specify here what the metric names should look like in the code (and JSON) as well as for Prometheus.

spec/src/main/asciidoc/metrics.asciidoc
donbourne: mpMetrics doesn't currently add `_total` to the end of counters -- I'd suggest we indicate that counter names should end in `_total`.  If mpMetrics adds `_total` to the end of counter names in prometheus format it should avoid doing so to metric names that already end in _total.

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,235 @@
...
+
+=== Metrics added for `@Timeout`
+
+[cols="8,3,4,9"]
+|===
+| Name | Type | Unit | Description
+
+|`ft_<name>_timeout_execution_duration`
+| Histogram | Nanoseconds
+| Histogram of execution times for the method
```
donbourne: note that, in the prometheus format, when a metric is in nanoseconds it will automatically be converted to seconds.

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,235 @@
...
+| Counter | None
+| Number of calls accepted by the bulkhead
+
+|`ft_<name>_bulkhead_calls_rejected_total`
+| Counter | None
+| Number of calls rejected by the bulkhead
+
+|`ft_<name>_bulkhead_execution_duration`
+| Histogram | Nanoseconds
+| Histogram of executions times
```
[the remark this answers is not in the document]
donbourne: If you have an `@Timeout` and and `@Bulkhead` on the same method, you will end up with two histograms with basically the same contents.  Histograms are expensive (they store the most recent 1000 samples to do computations), so it might be worth having one `ft_<name>_execution_duration` that is used for any FT annotation that requires execution timing.

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,235 @@
...
+Two retries were made during the invocation.
+
+`ft_com_example_MyClass_doWork_timeout_execution_duration` +
+The `Histogram` will have been updated with the length of time taken for each attempt. It will show a count of `3` and will have calculated averages and percentiles from the execution times.
+
+`ft_com_example_MyClass_doWork_timeout_calls_timed_out_total = 1` +
+One of the attempts timed out.
+
+`ft_com_example_MyClass_doWork_timeout_calls_not_timed_out_total = 2` +
+Two of the attempts did not time out.
```
donbourne: I really like this example.  It helps tie everything together.

spec/src/main/asciidoc/configuration.asciidoc
Azquelt: Done

spec/src/main/asciidoc/metrics.asciidoc
Azquelt: There's already a note at the bottom that all the counters are monotonic. As we're now keeping the `total` suffix, I think this is sufficient.

I'm reluctant to put "Monotonic Counter" in the type column of the table as `Counter` is the interface in the Metrics API which must be used.

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,235 @@
...
+
+[cols="8,3,4,9"]
+|===
+| Name | Type | Unit | Description
+
+|`ft_<name>_invocations_total`
+| Counter | None
+| The number of times the method was called
+
+|`ft_<name>_invocations_failed`
```
Azquelt: Should be `_invocations_failed_total` as it's a monotonic counter metric

OndrejM:
All looks good, I§m OK to merge. @pilhuhn, can you review and approve if OK?

Azquelt:
Just pushed one last change where I'd used `org.junit.Assert` instead of `org.hamcrest.MatcherAssert`

Azquelt:
Emily asked me to do one more push with the TCKs for `.invocation.total` and `.invocation.failed.total` metrics.

Unsquashed version is on the branch linked above.

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,237 @@
...
+| Amount of time the circuit breaker has spent in closed state
+
+|`ft.<name>.circuitbreaker.opened.total`
+| Counter | None
+| Number of times the circuit breaker has moved from closed state to open state
+|===
+
+Note that the sum of `ft.<name>.circuitbreaker.callsSucceeded.total`, `ft.<name>.circuitbreaker.callsFailed.total` and `ft.<name>.circuitbreaker.callsPrevented.total` will give the total number of calls for which the Bulkhead logic was run.
+
+Similarly, the sum of `ft.<name>.circuitbreaker.timeOpen.total`, `ft.<name>.circuitbreaker.timeHalfOpen.total` and `ft.<name>.circuitbreaker.timeClosed.total` will give the total time that the circuit breaker has been active for.
```
Emily-Jiang: I would say `ft.<name>.circuitbreaker.open.total` and `ft.<name>.circuitbreaker.halfOpen.total`

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,237 @@
...
+
+|`ft.<name>.bulkhead.callsRejected.total`
+| Counter | None
+| Number of calls rejected by the bulkhead
+
+|`ft.<name>.bulkhead.executionDuration`
+| Histogram | Nanoseconds
+| Histogram of method execution times. This does not include any time spent waiting in the bulkhead queue.
+
+|`ft.<name>.bulkhead.queuePopulation`*
```
Emily-Jiang: `ft.<name>.bulkhead.queuePopulation` => `ft.<name>.bulkhead.waitingQueuePopulation`

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -0,0 +1,237 @@
...
+| Gauge<Long> | None
+| Number of executions currently waiting in the queue
+
+|`ft.<name>.bulkhead.waitDuration`*
```
Emily-Jiang: `ft.<name>.bulkhead.waitDuration` => `ft.<name>.bulkhead.atWaitingQueueDuration`

spec/src/main/asciidoc/metrics.asciidoc
OndrejM: maybe to separate waitingQueue and population: `ft.<name>.bulkhead.waitingQueue.population`

spec/src/main/asciidoc/metrics.asciidoc
OndrejM: I suggest rather `ft.<name>.bulkhead.waitingQueue.duration` - it has the same prefix as the histogram `ft.<name>.bulkhead.waitingQueue.population` and would appear next to each other

Azquelt: Done (did not change the `executionDuration` metrics.

spec/src/main/asciidoc/metrics.asciidoc
Azquelt: Done

spec/src/main/asciidoc/metrics.asciidoc
Azquelt: Done

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -153,11 +153,11 @@ Similarly, the sum of `ft.<name>.circuitbreaker.timeOpen.total`, `ft.<name>.circ
 | Histogram | Nanoseconds
 | Histogram of method execution times. This does not include any time spent waiting in the bulkhead queue.
 
-|`ft.<name>.bulkhead.queuePopulation`*
+|`ft.<name>.bulkhead.waitingQueue.population`*
 | Gauge<Long> | None
 | Number of executions currently waiting in the queue
 
-|`ft.<name>.bulkhead.waitDuration`*
+|`ft.<name>.bulkhead.waitingQueue.duration`*
```
Emily-Jiang: `ft.<name>.bulkhead.waitingQueue.duration` => `ft.<name>.bulkhead.waiting.duration` as it describes the execution and is not measuring the waiting queue

Azquelt: Done.

Azquelt:
Squashed and ready to merge.

Issue #251, opened 2018-04-17
Use the latest release process to create FT 1.1


Emily-Jiang:
This repo needs to use the latest release process, same as Config repo.

PR #258, opened 2018-04-27
#251 update release scripts

Emily-Jiang:


pom.xml
```diff
@@ -287,21 +280,32 @@
             <build>
                 <plugins>
                     <plugin>
-                        <groupId>org.sonatype.plugins</groupId>
```
OndrejM: We don't need this plugin anymore? It's still in place for REST client: https://github.com/eclipse/microprofile-rest-client/blob/master/pom.xml#L297

OndrejM: The properties `autorelease` and `keepStagingReposOnFailure` are left defined but unused if we remove this plugin definition.

OndrejM: The sonatype plugin speeds up the process of uploading artifacts to sanatype. Using the default maven dpeloy plugin is very slow.

Emily-Jiang: good point. Let me put it back in.

OndrejM:
Looks good now, good luck with running the Jenkins job ;-)

PR #264, opened 2018-05-02
263 Add Connection to disableEnv test jars

benjamin-confino:
The connection class is used by AsyncClient in the tests in the disableEnv package. Without it the test fails with an injection error.


Emily-Jiang:
@benjamin-confino you need to sign off with -s on your commit.

benjamin-confino:
I repushed a signed commit

Emily-Jiang:
fix the tests for issue #109

Issue #247, opened 2018-04-09
Policies use subclasses of java.lang.Throwable to specify failures


mkouba:
However, according to the interceptors spec (Business Method Interceptors) around-invoke methods have the following signature:
> Object <METHOD>(InvocationContext) throws Exception

As a result, you cannot e.g. specify `org.eclipse.microprofile.faulttolerance.Retry.retryOn()` to be a custom `Throwable` (other than `java.lang.Error`and `java.lang.Exception`). A CDI container usually wraps such `Throwable` in a runtime exception.

The spec should probably specify that only exceptions and error are supported.

Emily-Jiang:
yes, we can clarify this in the spec.

PR #265, opened 2018-05-03
#247 clarify javadoc on Retry

Emily-Jiang:

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java
+++ b/api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java
@@ -97,14 +97,16 @@ public @interface Retry {
 
     /**
      *
-     * @return Specify the failure to retry on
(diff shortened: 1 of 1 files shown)


Emily-Jiang:
@mkouba please review

api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java
```diff
@@ -97,14 +97,16 @@
 
     /**
      *
-     * @return Specify the failure to retry on
+     * @return Specify the failure to retry on. Only a type of {@link java.long.Error} or {@link java.lang.Exception} can be specified.
```
mkouba: I think it should be something like: Only a type assignable from `java.long.Error` or `java.lang.Exception` can be used.

Emily-Jiang:
@mkouba please recheck to see you are happy with the spec wording changes?

Issue #110, opened 2017-08-23
CompletionStage return type support in Asynchronous


jGauravGupta:
As mention in Asynchronous doc, The method annotated with Asynchronous must return a Future.
We should also consider to support `java.util.concurrent.CompletionStage`.

New Java EE 8 api like `javax.enterprise.event.Event.fireAsync(event)` in CDI 2.0 and JAX-RS 2.1 supports the `CompletionStage` return type.

Emily-Jiang:
I had a further thought on this. Since the current javadoc says it must return a Future. CompleteableFuture is a Future and it has a CompletionStage. It should be fine. what do you think @jGauravGupta ?

OndrejM:
I commented on [#35](https://github.com/eclipse/microprofile-fault-tolerance/issues/35#issuecomment-326897126) before I found this issue - I agree with @jGauravGupta, Async. should support `CompletionStage`. Supporting only `Future` feels very awkward and it's also blocking, there's no way around it.

We should aim for supporting `CompletionStage` in FT 1.1, and also define that the default executor for `...Async()` methods should have the correct security context or naming context associated with it, similar to the thread that handles the async processing.

jGauravGupta:
It's ok @Emily-Jiang , we can target for `CompletionStage` interface support in the next release FT 1.1 .

Emily-Jiang:
agreed @jGauravGupta @OndrejM ! We can get this done in FT 1.1.

OndrejM:
Isn't this done with #267? Can we close this?

Emily-Jiang:
strangely. A link should not trigger this issue to be reopened. I'll close this.

Issue #245, opened 2018-04-09
Asynchronous execution - clarify the purpose of the returned `Future` object


mkouba:
The spec does not specify whether the returned `Future` object is only used as a way to pass the result value or whether we should always delegate to the underlying `Future`, e.g. what happens if an asynchronous method returns a `Future` object and the client invokes `Future#isDone()`? Should it return true once the `@Asynchronous` method completes? Or should it return true when the returned `Future` is done?

Emily-Jiang:
It is the latter. It should it return true when the returned Future is done?

mkouba:
Also what happens if `@Asynchronous` is combined with `@Retry`?

OndrejM:
As I suggested in #250, `@Asynchronous` should have the highest priority and so `@Retry` should happen in a separate thread. When `@Retry` gives an exception, the method should return a future with exception instead of throwing an exception. We need to update the spec once we have an agreement.

Emily-Jiang:
revisited this on today's hangout...
should it return true when the returned Future is done? 
I think it should return true when the returned Future is done.

Emily-Jiang:
@mkouba can you join the call on 12th September to discuss all asynchronous-related issues?

mkouba:
Probably not but @antoinesd could probably stand in for me?

antoinesd:
Yes @mkouba, I will

Azquelt:
As part of #110 to add support for `CompletionState` being returned from `@Asynchronous` methods, the Javadoc for `@Asynchronous` was mostly re-written and is more precise about the behaviour of the returned object while the method is running, when it has thrown an exception and when it has returned a value.

Could you check whether it's clear and precise enough to answer your questions?

PR #267, opened 2018-05-04
Issue #110 - Asynchronous with CompletionStage return type

OndrejM:
Also clarifies behavior when an exception is thrown from an asynch. method and how the value returned from proxy and returned from asynch. invocation relate to each other.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/faulttolerance/Asynchronous.java
+++ b/api/src/main/java/org/eclipse/microprofile/faulttolerance/Asynchronous.java
@@ -24,21 +24,27 @@ import java.lang.annotation.Inherited;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
+import java.util.concurrent.CompletableFuture;
 
 import javax.interceptor.InterceptorBinding;
 
 /**
  * Wrap the execution and invoke it asynchronously.
- * Any methods marked with this annotation must return {@link java.util.concurrent.Future}. 
+ * Any methods marked with this annotation must return one of:
+ * <ul>
+ *   <li>{@link java.util.concurrent.Future}</li>
+ *   <li>{@link java.util.concurrent.CompletionStage}</li>
+ * </ul>
  * Otherwise, {@link org.eclipse.microprofile.faulttolerance.exceptions.FaultToleranceDefinitionException} occurs.
  * Example usage:
- * 
+ *
  * <pre>
  * <code>@Asynchronous</code>
  * public Future&lt;String&gt; getString() {
  *  return CompletableFuture.completedFuture("hello");
  * }
  * </pre>
+ *
  * @author <a href="mailto:emijiang@uk.ibm.com">Emily Jiang</a>
  */
 @Documented
@@ -47,4 +53,15 @@ import javax.interceptor.InterceptorBinding;
 @InterceptorBinding
 @Inherited
 public @interface Asynchronous {
+
+    public static final class CompletedFuture {
+        
+        private CompletedFuture() {} // this is a util class only for static methods
+
+        public static <U> CompletableFuture<U> exceptionally(Throwable ex) {
+            CompletableFuture<U> future = new CompletableFuture<>();
+            future.completeExceptionally(ex);
+            return future;
+        }
+    }
 }
--- a/api/src/main/java/org/eclipse/microprofile/faulttolerance/Asynchronous.java
+++ b/api/src/main/java/org/eclipse/microprofile/faulttolerance/Asynchronous.java
@@ -39,8 +39,9 @@ import javax.interceptor.InterceptorBinding;
  * 
  * <p>
  * When a method marked with this annotation is executed, the call returns immediately while the method
- * is executed in another thread. Returned Future or CompletionStage is not completed until the method 
- * running asynchronously returns or throws an exception.
+ * is executed in another thread. The returned Future or CompletionStage is not completed until the result 
+ * of the method running asynchronously is completed (either normally or exceptionally) or the method 
+ * throws an exception.
  * </p>
  * 
  * <p>After a method marked with this annotation returns normally, the Future or CompletionStage returned
@@ -98,14 +99,4 @@ import javax.interceptor.InterceptorBinding;
 @Inherited
 public @interface Asynchronous {
 
-    public static final class CompletedFuture {
-        
-        private CompletedFuture() {} // this is a util class only for static methods
-
-        public static <U> CompletableFuture<U> exceptionally(Throwable ex) {
-            CompletableFuture<U> future = new CompletableFuture<>();
-            future.completeExceptionally(ex);
-            return future;
-        }
-    }
 }
--- a/api/src/main/java/org/eclipse/microprofile/faulttolerance/Asynchronous.java
+++ b/api/src/main/java/org/eclipse/microprofile/faulttolerance/Asynchronous.java
@@ -36,15 +36,59 @@ import javax.interceptor.InterceptorBinding;
  *   <li>{@link java.util.concurrent.CompletionStage}</li>
  * </ul>
  * Otherwise, {@link org.eclipse.microprofile.faulttolerance.exceptions.FaultToleranceDefinitionException} occurs.
+ * 
+ * <p>
+ * When a method marked with this annotation is executed, the call returns immediately while the method
+ * is executed in another thread. Returned Future or CompletionStage is not completed until the method 
+ * running asynchronously returns or throws an exception.
+ * </p>
+ * 
+ * <p>After a method marked with this annotation returns normally, the Future or CompletionStage returned
+ * in the calling thread delegate all method calls to the object returned by the method. 
+ * CompletionStage returned in the calling thread is completed when the stage returned by the method is completed
+ * and with the same completion value or exception.</p>
+ * 
+ * <p>If a method marked with this annotation throws an exception when invoked, the exception isn't thrown 
+ * in the callers thread therefore it's not possible to catch it with a {@code try..catch} block. 
+ * The exception is propagated asynchronously:
+ * </p>
+ * <ul>
+ *   <li>If the method declares {@link java.util.concurrent.Future} as the return type, 
+ * calling {@link java.util.concurrent.Future#get()} will throw an 
+ * {@link java.util.concurrent.ExecutionException} wrapping the original exception</li>
+ *   <li>If the method declares {@link java.util.concurrent.CompletionStage} as the return type, 
+ * it is completed exceptionally with the exception.</li>
+ * </ul>
+ * It's recommended that methods throw only runtime exceptions to avoid unnecessary {@code try..catch} blocks.
+ * 
+ * <p>If a class is annotated with this annotation, all class methods that could be marked with this 
+ * annotation are treated as if they were marked with this annotation.
+ * </p>
+ * 
+ * <p>
  * Example usage:
+ * </p>
  *
  * <pre>
- * <code>@Asynchronous</code>
- * public Future&lt;String&gt; getString() {
+ * <code>@Asynchronous
+ * public CompletionStage&lt;String&gt; getString() {
  *  return CompletableFuture.completedFuture("hello");
  * }
+ * </code>
  * </pre>
  *
+ * <p>
+ * Example call with exception handling:
+ * </p>
+ *
+ * <pre><code>
+ * CompletionStage stage = getString().exceptionally(e -&gt; {
+ *     handleException(e); 
+ *     return null;
+ * });
+ * </code>
+ * </pre>
+ * 
  * @author <a href="mailto:emijiang@uk.ibm.com">Emily Jiang</a>
  */
 @Documented
--- a/api/src/main/java/org/eclipse/microprofile/faulttolerance/Asynchronous.java
+++ b/api/src/main/java/org/eclipse/microprofile/faulttolerance/Asynchronous.java
@@ -24,7 +24,6 @@ import java.lang.annotation.Inherited;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
-import java.util.concurrent.CompletableFuture;
 
 import javax.interceptor.InterceptorBinding;
 
--- a/spec/src/main/asciidoc/asynchronous.asciidoc
+++ b/spec/src/main/asciidoc/asynchronous.asciidoc
@@ -17,6 +17,7 @@
 // limitations under the License.
 // Contributors:
 // Emily Jiang
+// Ondro Mihalyi
 
 [[asynchronous]]
 == Asynchronous
@@ -28,12 +29,12 @@ This thread should have the correct security context or naming context associate
 === Asynchronous Usage
 
 A method or a class can be annotated with `@Asynchronous`, which means the method or the methods under the class will be invoked by a separate thread.
-The method annotated with `@Asynchronous` must return a `Future`, otherwise, `FaultToleranceDefinitionException` occurs.
+The method annotated with `@Asynchronous` must return a `Future` or a `CompletionStage` from the `java.util.concurrent` package. Otherwise, `FaultToleranceDefinitionException` occurs.
 
 [source, java]
 ----
 @Asynchronous
-public Future<Connection> serviceA() {
+public CompletionStage<Connection> serviceA() {
    Connection conn = null;
    counterForInvokingServiceA++;
    conn = connectionService();
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousCSTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousCSTest.java
@@ -29,6 +29,7 @@ import javax.inject.Inject;
 
 import org.eclipse.microprofile.fault.tolerance.tck.asynchronous.AsyncClassLevelClient;
 import org.eclipse.microprofile.fault.tolerance.tck.asynchronous.AsyncClient;
+import org.eclipse.microprofile.fault.tolerance.tck.asynchronous.CompletableFutureHelper;
 import org.eclipse.microprofile.fault.tolerance.tck.util.Connection;
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.testng.Arquillian;
@@ -75,15 +76,14 @@ public class AsynchronousCSTest extends Arquillian {
     public void testAsyncIsNotFinished() {
         CompletableFuture<Void> waitingFuture = newWaitingFuture();
         
-        CompletionStage<Connection> resultFuture = null;
-        resultFuture = client.serviceCS(waitingFuture);
+        CompletionStage<Connection> resultFuture = client.serviceCS(waitingFuture);
 
-        Assert.assertFalse(resultFuture.toCompletableFuture().isDone());
+        Assert.assertFalse(isCompleted(resultFuture));
 
         complete(waitingFuture);
 
         try {
-            resultFuture.toCompletableFuture().join();
+            waitUntilCompleted(resultFuture);
         }
         catch (CompletionException e) {
             handleCompletionException(e);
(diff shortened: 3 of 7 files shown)


OndrejM:
Solves #110 and I believe that also #245

OndrejM:
@jGauravGupta, @mkouba  can you please review if this solves #110 (@jGauravGupta) and #245 (@mkouba) ?

api/src/main/java/org/eclipse/microprofile/faulttolerance/Asynchronous.java
```diff
@@ -47,4 +97,15 @@
 @InterceptorBinding
 @Inherited
 public @interface Asynchronous {
+
+    public static final class CompletedFuture {
```
mkouba: What is this for? A protected method would usually return completed Future/CompletionStage, i.e. `CompletableFuture.completedFuture("hello")` from your example and throw an exception if something's wrong.

api/src/main/java/org/eclipse/microprofile/faulttolerance/Asynchronous.java
```diff
@@ -24,21 +24,71 @@
...
  * 
+ * <p>
+ * When a method marked with this annotation is executed, the call returns immediately while the method
+ * is executed in another thread. Returned Future or CompletionStage is not completed until the method 
+ * running asynchronously returns or throws an exception.
+ * </p>
+ * 
+ * <p>After a method marked with this annotation returns normally, the Future or CompletionStage returned
+ * in the calling thread delegate all method calls to the object returned by the method. 
+ * CompletionStage returned in the calling thread is completed when the stage returned by the method is completed
```
mkouba: I think this contradicts the sentence in the previous paragraph: `CompletionStage is not completed until the method running asynchronously returns`.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousCSTest.java
```diff
@@ -0,0 +1,314 @@
...
+     * done if called right after the operation
+     */
+    @Test
+    public void testAsyncIsNotFinished() {
+        CompletableFuture<Void> waitingFuture = newWaitingFuture();
+        
+        CompletionStage<Connection> resultFuture = null;
+        resultFuture = client.serviceCS(waitingFuture);
+
+        Assert.assertFalse(resultFuture.toCompletableFuture().isDone());
```
mkouba: Note that `CompletionStage` impls need not support `toCompletableFuture()`...

api/src/main/java/org/eclipse/microprofile/faulttolerance/Asynchronous.java
OndrejM: Valid point, I'll change the previous paragraph to something like "The returned Future or CompletionStage is not completed until the result of the method running asynchronously is completed (either normally or exceptionally) or the method throws an exception."

api/src/main/java/org/eclipse/microprofile/faulttolerance/Asynchronous.java
OndrejM: There's no way to create a CF completed exceptionally in one line of code, as in Java 8 there's nothing like `CompletableFuture.completedFutureExceptionally(new Exception())`. Although it should be possible to just throw an exception, I wanted to add this convenience method to allow returning a CF completed exceptionally as an alternative.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousCSTest.java
OndrejM: You're right, I didn't realize that. Do you have any idea how to simple find out if a CompletionStage is completed? I could create another CompletableFuture which would be completed in CompletionStage's final callback but maybe you know of a better solution?

Example of the solution I have in mind:

```
CompletableFuture resultFutureStatus = new CompletableFuture();
resultFuture.whenComplete( (v, e) -> {
 if (e != null) {
  resultFutureStatus.completeExceptionally(e);
 } else {
  resultFutureStatus.complete(null);
 }
}

Assert.assertFalse(resultFutureStatus.isDone());
```

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousCSTest.java
```diff
@@ -0,0 +1,314 @@
...
+        catch (CompletionException e) {
+            handleCompletionException(e);
+        }
+    }
+
+    /**
+     * Test that the callbacks added to the initial stage are executed
+     * after the stage returned by the asynchronous method call is completed.
+     * 
+     * The callbacks added inside method invokation must be called first and
```
michalszynkiewicz: typo - invokation -> invocation

api/src/main/java/org/eclipse/microprofile/faulttolerance/Asynchronous.java
mkouba: Yes, I know. But I'm not sure it's worth the effort (I think it would rarely be used) and also whether `Asynchronous` is the best place to declare such a convenient library class/method.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousCSTest.java
mkouba: Well, I think you could make use of some synchronization aid such as `java.util.concurrent.CountDownLatch`, e.g. something like:
````java
AtomicBoolean completedNormally = new AtomicBoolean(false);
CountDownLatch latch = new CountDownLatch(1);
foo.someAsyncMethod().whenComplete((v, e) -> {
  latch.countDown();
  if(e == null) {
    completedNormally.set(true);
  }
latch.await(500, TimeUnit.MILLISECONDS);
Assert.assertTrue(completedNormally.get());
````
And the `AtomicBoolean` is unnecessary if you only want to test whether a stage is complete or not.

File: api/src/main/java/org/eclipse/microprofile/faulttolerance/Asynchronous.java (hunk 1 above)
Comment (OndrejM) [reply api/src/main/java/org/eclipse/microprofile/faulttolerance/Asynchronous.java]: OK, you're probably right that it won't be used very often. And Java 9 provides the functionality with `CF.failedFuture` so I'll remove it.

File: tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousCSTest.java (hunk 3 above)
Comment (OndrejM) [reply tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousCSTest.java]: I solved this using a CompletableFuture instead of CuntDownLatch and AtomicBoolean, I think it's simpler. Please review if it's correct.

File: api/src/main/java/org/eclipse/microprofile/faulttolerance/Asynchronous.java [hunk 5]
```diff
@@ -24,21 +24,72 @@
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
+import java.util.concurrent.CompletableFuture;
```
mkouba: Unused import...

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/CompletableFutureHelper.java
```diff
@@ -0,0 +1,46 @@
+package org.eclipse.microprofile.fault.tolerance.tck.asynchronous;
```
mkouba: Missing license...

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/CompletableFutureHelper.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /**
+     * Returns a future that is completed when the stage is completed and has the same value or exception as the completed stage. It's supposed to be equivalent to calling {@link CompletionStage#toCompletableFuture()} but works with any CompletionStage and doesn't throw {@link java.lang.UnsupportedOperationException}.
+     * @param <U> The type of the future result
+     * @param stage Stage to convert to a future
+     * @return Future converted from stage
+     */
+    public static <U> CompletableFuture<U> toCompletableFuture(CompletionStage<U> stage) {
+        CompletableFuture future = new CompletableFuture();
```
mkouba: Why the raw type?

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousCSTest.java
mkouba: Looks good but I did not have enough time to go through all the tests and details...

spec/src/main/asciidoc/asynchronous.asciidoc
```diff
@@ -28,12 +29,12 @@ This thread should have the correct security context or naming context associate
 === Asynchronous Usage
 
 A method or a class can be annotated with `@Asynchronous`, which means the method or the methods under the class will be invoked by a separate thread.
-The method annotated with `@Asynchronous` must return a `Future`, otherwise, `FaultToleranceDefinitionException` occurs.
+The method annotated with `@Asynchronous` must return a `Future` or a `CompletionStage` from the `java.util.concurrent` package. Otherwise, `FaultToleranceDefinitionException` occurs.
```
jGauravGupta: @Asynchronous should also support subclass of `Future` or `CompletionStage` interface.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/CompletableFutureHelper.java
OndrejM: A mistake, I'll fix it. thanks @mkouba

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/CompletableFutureHelper.java
OndrejM: License added

mkouba:
@OndrejM I think the broken TCK test is a problem. How would you implement this clarification and still pass the TCK? So if we merge this PR we would have to fix the test or add it to some official exclude list.

@Emily-Jiang Ondrej is right that this PR only contains clarifications and new stuff, i.e. it _does not break the API_. On the other hand, it may break existing apps relying on impl-specific behavior. From my point of view, having this part of the spec unspecified is more like a bug because it prevents from writing portable apps. What are the rules for backward compatibility and breaking changes in MP anyway?

OndrejM:
@mkouba, @Emily-Jiang for the tests, I would like to fix them together with this PR. It doesn't make sense to me that asynchronous methods would throw exceptions - waiting for exceptions with try..catch requires a blocking call. There's very little value in blocking execution of asynchronous methods. Asynchronous methods should always provide the result or exception via the returned future to avoid blocking calls.

I didn't find any explanation of how exceptions should be handled for asynchronous calls. This is covered only in the TCK and therefore I believe no user can rely on the contract, which is undocumented. 

The specification v1.0 is inconsistent. When `@Timeout` is used with `@Asynchronous`, the returned future should When @Timeout is used with @Asynchronous, then a separate thread will be spawned to perform the `TimeoutException` is thrown from `Future.get` and not from the asynchrnous method. On the other hand, when `@Asynchronous` is used with `@Bulkhead`, the spec says that "BulkheadException will be thrown", without specifying any details. Only the TCK shows that the exception is thrown from the asynch. method and not from the returned `Future`, which is IMHO wrong and should be fixed to behave as explained for `@Timeout`.

The spec doesn't say anything about exception handling when `@Retry` and `@CircuitBreaker` is used with `@Asynchronous`. It only says that they can be combined with it.

So the only thing that the proposed update breaks is the exception handling of asynch. bulk-headed calls and some tests that catch exceptions for asynch. calls. All of this is in my opinion wrong from the start and shouldn't have ended up in the FT 1.0 spec at all.

OndrejM:
@mkouba @Emily-Jiang if the backwards compatibility for asynchronous bulkheads is really the only issue (as this is I think the only thing that would be broken with this update) then we can leave the special behavior for bulkheads and specify that the method would throw an exception just for bulkheads. It makes sense that bulkheads have higher priority than asynchronous execution and a blocking call in case of bulkheads shouldn't really be blocking for a long time in case of a bulkhead exception.

mkouba:
> It makes sense that bulkheads have higher priority than asynchronous execution...

I'm not so sure. Bulkhead is per spec used to _"limit the number of the concurrent calls to an instance"_. And in case of  `@Asynchronous` an instance is in fact invoked in a worker/async thread. I think that for users it would be easier to understand something like _"whenever you add `@Asynchronous` check `ExecutionException`"_.

Also this might complicate things if you combine `@Bulkhead` with `@Retry` and `@Asynchronous`. A user probably expects that retry policy is applied if `BulkheadException` is thrown. In other words, implementors would have to apply retry on the call site and also in the worker thread. Which means that (as Ondrej pointed out) the execution may block on the call site.

> However, some tests (such as AsyncTimeoutTest.testAsyncTimeout) should be updated to not catch any other exception from asynchronous calls as this is not mandated by the FT 1.0 spec and catching exceptions in these tests is simply wrong.

+1

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousCSTest.java
```diff
@@ -0,0 +1,327 @@
...
+        future.complete(null);
+    }
+
+    private void handleCompletionException(CompletionException e) throws AssertionError {
+        throw new AssertionError("testAsync: unexpected Exception calling service: "
+                + e.getCause().getMessage());
+    }
+    
+    private static Connection waitUntilCompleted(CompletionStage<Connection> resultFuture) {
+        return CompletableFutureHelper.toCompletableFuture(resultFuture).join();
```
Azquelt: `toCompletableFuture` is an optional method. Do we need to specify that it must be implemented if we're using it in the TCK?

Alternatively we could pass a method to `.handle()` which would record when it completes and whether it's exceptional or cancelled.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousCSTest.java
```diff
@@ -0,0 +1,327 @@
...
+     * if called after waiting enough time to end the operation
+     */
+    @Test
+    public void testAsyncIsFinished() {
+        CompletableFuture<Void> waitingFuture = newWaitingFuture();
+        
+        CompletionStage<Connection> resultFuture = client.serviceCS(waitingFuture);
+        complete(waitingFuture);
+
+        Assert.assertTrue(isCompleted(resultFuture));
```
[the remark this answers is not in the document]
Azquelt: I think that would work. Might also make sense to change the method name as well so it's obvious that it's more complex than `future.isDone()`.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousCSTest.java
```diff
@@ -0,0 +1,327 @@
...
+    /**
+     * A helper method to complete a waiting future with a more readable syntax
+     */
+    private void complete(CompletableFuture<?> future) {
+        future.complete(null);
+    }
+
+    private void handleCompletionException(CompletionException e) throws AssertionError {
+        throw new AssertionError("testAsync: unexpected Exception calling service: "
+                + e.getCause().getMessage());
```
Azquelt: If you're going to catch the exception and throw an `AssertionError`, can you include the completion exception as the cause so that the test results include the full stack trace?

Alternatively, just let the `CompletionException` be thrown with no special handling.

api/src/main/java/org/eclipse/microprofile/faulttolerance/Asynchronous.java
OndrejM: Done

spec/src/main/asciidoc/asynchronous.asciidoc
OndrejM: Won't do

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/CompletableFutureHelper.java
OndrejM: Done

Emily-Jiang:
@OndrejM do you have time to fix the test? Do you want @Azquelt to help out the tests as I think we need to make some progress on this

api/src/main/java/org/eclipse/microprofile/faulttolerance/Asynchronous.java
```diff
@@ -29,16 +29,66 @@
...
+ * <ul>
+ *   <li>If the method declares {@link java.util.concurrent.Future} as the return type, 
+ * calling {@link java.util.concurrent.Future#get()} will throw an 
+ * {@link java.util.concurrent.ExecutionException} wrapping the original exception</li>
+ *   <li>If the method declares {@link java.util.concurrent.CompletionStage} as the return type, 
+ * it is completed exceptionally with the exception.</li>
+ * </ul>
+ * It's recommended that methods throw only runtime exceptions to avoid unnecessary {@code try..catch} blocks.
+ * 
+ * <p>If a class is annotated with this annotation, all class methods that could be marked with this
```
Azquelt: I'm not sure this is clear enough. Reading this paragraph, I couldn't work out whether it would mean this class is a valid configuration or not:
```
@Asynchronous
public class MyClass {
   public String myMethodA() {}
   public Future<?> myMethodB() {}
}
```

I think this class should cause a definition exception to be thrown because `myMethodA` is not valid.

Can we just say:
> If a class is annotated with this annotation, all class methods are treated as if they were marked with this annotation.

If we really wanted to specify that methods that don't return `Future` or `CompletionStage` should be ignored, we could say
> If a class is annotated with this annotation, all class methods which return `Future` or `CompletionStage` are treated as if they were marked with this annotation.

I also couldn't see any TCKs that check validation of `@Asynchronous` methods or classes, so we should add some.

OndrejM: I certainly had the second interpretation in mind, but now I can't see a usecase for mixing sync and async method where it would make sense to use `@Asynchronous` on the class level. 

Therefore I agree with:

> If a class is annotated with this annotation, all class methods are treated as if they were marked with this annotation.

And I would add:

> If one of the methods doesn't return Future/CompletionStage, {@link org.eclipse.microprofile.faulttolerance.exceptions.FaultToleranceDefinitionException} occurs (at deploy time if the bean is discovered during deployment). 


I would also add some tests to cover this.

If in a rare case most of methods are async and only a handful of methods are sync, `@Asynchronous` can still be added to all async methods. I think it's very rare and it's not worth to simplify at the expense of complexity.

api/src/main/java/org/eclipse/microprofile/faulttolerance/Asynchronous.java
```diff
@@ -29,16 +29,66 @@
 
 /**
  * Wrap the execution and invoke it asynchronously.
- * Any methods marked with this annotation must return {@link java.util.concurrent.Future}. 
+ * Any methods marked with this annotation must return one of:
+ * <ul>
+ *   <li>{@link java.util.concurrent.Future}</li>
+ *   <li>{@link java.util.concurrent.CompletionStage}</li>
+ * </ul>
  * Otherwise, {@link org.eclipse.microprofile.faulttolerance.exceptions.FaultToleranceDefinitionException} occurs.
```
OndrejM: This exception should be thrown at deploy time if the class is discovered during deployment. We need to update the wording and also cover it by a test if not already.

Emily-Jiang:
As discussed in our hangout last week and this week, I am going to merge in now as I have done the 1.1.2 release.

PR #268, opened 2018-05-08
Metrics add fallback counter

Azquelt:
Add a new counter named `ft.<name>.fallback.calls.total` for how many times a fallback method or handler is called.

Currently, there's no way to find this out. Also, with the way `ft.<name>.invocations.failed.total` is specified, that counter will only increment for a method annotated with `@Fallback` if the fallback method also throws an exception.

This metric gives you a better idea of how often your main logic fails, as well as allowing you to see if your fallback itself is failing.

For #234

### DIFF
--- a/spec/src/main/asciidoc/metrics.asciidoc
+++ b/spec/src/main/asciidoc/metrics.asciidoc
@@ -21,14 +21,14 @@
 == Integration with Microprofile Metrics
 
 When Microprofile Fault Tolerance and Microprofile Metrics are used together, metrics are automatically added for each of
-the methods annotated with a `@Retry`, `@Timeout`, `@CircuitBreaker` or `@Bulkhead` annotation.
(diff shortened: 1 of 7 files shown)


tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/AllMetricsBean.java
```diff
@@ -38,9 +39,13 @@
     @Bulkhead(3)
     @Timeout(value = 1000, unit = ChronoUnit.MILLIS)
     @CircuitBreaker(failureRatio = 1.0, requestVolumeThreshold = 20)
+    @Fallback(fallbackMethod = "doFallback")
```
OndrejM: The TCK should also cover the case when a fallback handler is used: `@Fallback(MyFallbackHandler.class)`

Azquelt: Done. I've added as a second test in `FallbackMetricTest`.

Issue #266, opened 2018-05-03
Allow explicit trues in Disabled Fault Tolerance policies with configuration 


benjamin-confino:
In configuration.asciidoc under Disabled Fault Tolerance policies with configuration there is the following statement:

`If the above configurations patterns are used with another value than `false` (i.e. `<classname>/<methodname>/<annotation>/enabled=whatever`) non-portable behaviour results`.

I believe that the spec should cover the value `true` as well as `false` and state that should the value true explicitly set then it will override a value of false on a higher scope. This will allow you do something like disable the Retry annotation globally but override it for a single method while you debug that method.

Emily-Jiang:
We should improve this. @antoinesd what do you think?

PR #270, opened 2018-05-08
#266 improve the spec to take in explicit trues

Emily-Jiang:

### DIFF
--- a/spec/src/main/asciidoc/configuration.asciidoc
+++ b/spec/src/main/asciidoc/configuration.asciidoc
@@ -113,7 +113,12 @@ Similarly, the property `com.acme.ClassA/Retry/maxRetries` will be ignored if `@
 
 === Disabled Fault Tolerance policies with configuration
 
-Fault Tolerance policies can be disabled with configuration at method level, class level or globally for all deployment.
+Fault Tolerance policies can be disabled with configuration at method level, class level or globally for all deployment. If multiple configurations are specified, method-level configuration overrides class-level configuration, which then overrides global configuration. e.g.
+* `com.acme.test.MyClient/methodA/CircuitBreaker/enabled=true`
+* `com.acme.test.MyClient/CircuitBreaker/enabled=false`
+* `CircuitBreaker/enabled=false`
+For the above scenario, all occurrence of `CircuitBreaker` for the application is disabled except the method `methodA` on the class `com.acme.test.MyClient`.
+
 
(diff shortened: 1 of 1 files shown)


Emily-Jiang:
just spec change..... tck changes will come soon.

spec/src/main/asciidoc/configuration.asciidoc
```diff
@@ -113,7 +113,12 @@ Similarly, the property `com.acme.ClassA/Retry/maxRetries` will be ignored if `@
 
 === Disabled Fault Tolerance policies with configuration
 
-Fault Tolerance policies can be disabled with configuration at method level, class level or globally for all deployment.
+Fault Tolerance policies can be disabled with configuration at method level, class level or globally for all deployment. If multiple configurations are specified, method-level configuration overrides class-level configuration, which then overrides global configuration. e.g.
+* `com.acme.test.MyClient/methodA/CircuitBreaker/enabled=true`
+* `com.acme.test.MyClient/CircuitBreaker/enabled=false`
```
Azquelt: This middle line doesn't make a lot of sense. It's valid but has no effect because CircuitBreaker is disabled globally.

I think the example would be clearer if this line was removed.

OndrejM: Alternatively, change the example to be global=false (as is), class-level=true, method-level=false and the explanation should state that all circuit breakers in the app are disabled, only those in the class except the method methodA are enabled.

spec/src/main/asciidoc/configuration.asciidoc
```diff
@@ -153,4 +158,4 @@ For instance the following config will disable bulkhead policy globally:
 
 Policy will be disabled everywhere ignoring existing policy annotations on methods and classes.
 
-If the above configurations patterns are used with another value than `false` (i.e. `<classname>/<methodname>/<annotation>/enabled=whatever`) non-portable behaviour results.
+If the above configurations patterns are used with another value than `true` or `false` (i.e. `<classname>/<methodname>/<annotation>/enabled=whatever`) non-portable behaviour results.
```
OndrejM: Grammar: "other values than" instead of "another value than"

Azquelt:
@OndrejM I've addressed your spec feedback and created TCKs, but I've had to create a new PR because I don't have permission to push directly to branches in this repository.

Would you review my changes in #275? Thanks.

PR #272, opened 2018-05-09
State what happens when metric names clash

Azquelt:
I'm not sure how specific we should be here.

We could go all the way and explain how each metric type should be combined - in most cases this is a simple sum. We could then add tests to ensure that metrics are combined correctly in all cases.

From an implementation point of view, this isn't too arduous, though some additional work would be required to combine gauges.

However, I'm not sure how valuable combining metrics is for the user. Essentially we're just trying to do something sensible if this situation happens so for the moment, I've just added a simple test which ensures that the implementation doesn't throw an exception and increments the basic invocation metric.

### DIFF
--- a/spec/src/main/asciidoc/metrics.asciidoc
+++ b/spec/src/main/asciidoc/metrics.asciidoc
@@ -28,6 +28,9 @@ the methods annotated with a `@Retry`, `@Timeout`, `@CircuitBreaker`, `@Bulkhead
 The automatically added metrics follow a consistent pattern which includes the fully qualified name of the annotated method.
 In the tables below, the placeholder `<name>` should be replaced by the fully qualified method name.
 
+If two methods have the same fully qualified name then the metrics for those methods will be combined. The result of this combination
+is non-portable and may vary between implementations. For portable behavior, monitored methods in the same class should have unique names.
+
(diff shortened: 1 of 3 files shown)


spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -28,6 +28,8 @@ the methods annotated with a `@Retry`, `@Timeout`, `@CircuitBreaker` or `@Bulkhe
 The automatically added metrics follow a consistent pattern which includes the fully qualified name of the annotated method.
 In the tables below, the placeholder `<name>` should be replaced by the fully qualified method name.
 
+If two methods have the same fully qualified name then the metrics for those methods will be combined.
```
OndrejM: I would at least explain that the end result is not portable and may vary across implementations. For portable behavior, monitored methods in the same class should have unique names.

Azquelt: Makes sense, I will add that.

Azquelt: Done.

Azquelt:
I have made the requested changes and rebased on master to resolve a conflict.

OndrejM:
I'm merging this as it's not anything huge,  more or less a clarification

PR #275, opened 2018-05-11
Allow annotations to be explicitly enabled via config

Azquelt:
Allow annotations to be explicitly enabled via config as well as being disabled.

This allows a Fault Tolerance annotation to be disabled globally, but then enabled for a specific class or method.

Add TCKs to test this behaviour.

This is to address #266 

This includes and completes the work that @Emily-Jiang started under #270

### DIFF
--- a/spec/src/main/asciidoc/configuration.asciidoc
+++ b/spec/src/main/asciidoc/configuration.asciidoc
@@ -114,10 +114,10 @@ Similarly, the property `com.acme.ClassA/Retry/maxRetries` will be ignored if `@
 === Disabled Fault Tolerance policies with configuration
 
 Fault Tolerance policies can be disabled with configuration at method level, class level or globally for all deployment. If multiple configurations are specified, method-level configuration overrides class-level configuration, which then overrides global configuration. e.g.
-* `com.acme.test.MyClient/methodA/CircuitBreaker/enabled=true`
-* `com.acme.test.MyClient/CircuitBreaker/enabled=false`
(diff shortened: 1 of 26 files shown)


Azquelt:
The last commit is unrelated but is needed to allow the project to build.

Azquelt:
This ended up being a fairly large change. There are several different ways that config parameters can override each other and testing them across all the annotations makes for a fairly large test matrix.

Given that we want to run very similar tests in lots of different configurations, I've written a new "client" bean with behaviour designed to make it very easy to test whether or not the annotations are enabled.

This allows the actual test methods to be a few lines in most cases. (The exception is the Bulkhead tests because bulkheads need several executions running in parallel and this takes a few more lines to set up.)

The methods on the new client bean throw `TestException` as where the old client bean threw a `RuntimeException`. This was because I found that the old circuitBreaker enablement tests were assuming that a caught `RuntimeException` was thrown from the test bean, whereas it could actually be a `CircuitBreakerOpenException` as that's a subclass of `RuntimeException`. Using a separate exception class to act as a dummy application exception avoids the potential for this sort of mistake.

I've added bulkhead enablement tests which were previously missing.

Note: I'm slightly unsure about the bulkhead tests. I'm not sure it's guaranteed that the ApplicationScoped context is active on the new thread, but I went with it since that's the approach we already use in `BulkheadSynchTest`.

I've refactored the way the config is declared to try to minimise copy paste errors since the meat of these tests is setting up different configurations.

PR #279, opened 2018-05-16
278- new tests to ensure enabling annotations overrides MP_Fault_Tolerance…

benjamin-confino:
…_NonFallback_Enabled


hutchig:
Ben is about to add another commit to this so best not to merge right now.

PR #320, opened 2018-09-21
TCK: Tests in org.eclipse.microprofile.fault.tolerance.tck.metrics do not add beans.xml to deployments

### LINKED ISSUE #318: TCK: Tests in org.eclipse.microprofile.fault.tolerance.tck.metrics do not add beans.xml to deployments
For example https://github.com/eclipse/microprofile-fault-tolerance/blob/master/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/BulkheadMetricTest.java#L49-L56

For compatibility reasons at least an empty `beans.xml` should be added.

antoinesd:
Fixes #318

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/AllMetricsTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/AllMetricsTest.java
@@ -1,5 +1,4 @@
 /*
- *******************************************************************************
  * Copyright (c) 2018 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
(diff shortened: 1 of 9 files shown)


tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/AllMetricsTest.java
```diff
@@ -45,7 +45,8 @@ public static WebArchive deploy() {
         WebArchive war = ShrinkWrap.create(WebArchive.class, "ftMetricAll.war")
                 .addClasses(AllMetricsBean.class)
                 .addPackage(Packages.UTILS)
-                .addPackage(Packages.METRIC_UTILS);
+                .addPackage(Packages.METRIC_UTILS)
+                .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml");
```
mkouba: Note that in a war, the `beans.xml` file must be named `WEB-INF/beans.xml` or `WEB-INF/classes/META-INF/beans.xml`. I'm not sure where does shrinkwrap add the resource when using `WebArchive#addAsManifestResource`...

PR #321, opened 2018-09-27
Fix javadoc warnings and build issue with long lines

brunobat:
Some housekeeping 
We were missing a few javadoc and the compiler was complaining.
In a recent PR, a class had lines greater that 150 chars and the building was failing because of that.


genie-microprofile:

Issue #312, opened 2018-09-13
Retry: What happens when delay is less than jitter?


Azquelt:
The `@Retry` javadoc says that a random value between `jitter` and `-jitter` is added to `delay` to find out how much delay should be applied to each retry attempt.

What happens when the value of `delay` is less than the value of `jitter`?

This question is particularly pertinent because the default configuration has `delay=0ms, jitter=200ms`. Obviously we can't delay for `-200ms`, but what is the expected behaviour?

Azquelt:
Personally, I think that if `delay=0` then there should be no delay. I can see the following options:

* Make it invalid for jitter to be less than delay. Change the default jitter to 0.
* Specify that jitter is set to the same value as delay if jitter is greater than delay.

Emily-Jiang:
@antoinesd and Bruno to check the current behavior on Thorntail and Safeguard and rediscuss

Emily-Jiang:
@brunobat to contribute tcks.

Azquelt:
I think this has been cleared up, thanks :)

PR #325, opened 2018-10-10
Issue #312 - Retry: What happens when delay is less than jitter? 

brunobat:
Test to make sure the minimum effective delay is 0, even when jitter > delay.
Tested this in Safegard... It implements Jitter after all. These were the effective delays, in ms:
[1, 0, 0, 1, 0, 0, 28, 279, 0, 86, 19, 0, 370, 360, 142, 325, 1, 0, 200, 350, 1, 369, 0, 1, 0, 0, 0, 0, 43, 0, 0, 0, 51, 230, 37, 0, 1, 87, 0, 149, 70]

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java
+++ b/api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java
@@ -81,6 +81,8 @@ public @interface Retry {
     /**
      * Set the jitter to randomly vary retry delays for. The value must be greater than or equals to 0. 
      * 0 means not set.
+     * The effective delay will be [delay - jitter, delay + jitter] and allways greater than or equal to 0.
+     * Negative effective delays will be 0.
      * @return the jitter that randomly vary retry delays by. e.g. a jitter of 200 milliseconds
      * will randomly add between -200 and 200 milliseconds to each retry delay.
      */
--- a/api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java
+++ b/api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java
@@ -83,7 +83,7 @@ public @interface Retry {
      * Set the jitter to randomly vary retry delays for. The value must be greater than or equals to 0.
      * 0 means not set.
      * </p>
-     * The effective delay will be [delay - jitter, delay + jitter] and allways greater than or equal to 0.
+     * The effective delay will be [delay - jitter, delay + jitter] and always greater than or equal to 0.
      * Negative effective delays will be 0.
      *
      * @return the jitter that randomly vary retry delays by. e.g. a jitter of 200 milliseconds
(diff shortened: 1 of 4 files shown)


genie-microprofile:

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryTest.java
```diff
@@ -126,6 +128,37 @@ public void testRetryWithDelay() {
...
+        }
+        catch (RuntimeException ex) {
+            // Expected
+        }
+
+        final int retryCountForConnectionService = retryClientWithNoDelayAndJiter.getRetryCountForConnectionService();
+        final int fastDelays = retryClientWithNoDelayAndJiter.fastDelays();
+
+        // 60% to discard false positives
+        Assert.assertTrue(fastDelays < (retryCountForConnectionService * 0.6),
```
Azquelt: I think there's a fairly high chance that this test will fail due to random chance, assuming that the jitter is evenly distributed between -400ms and 400ms and then clamped to be above zero (maybe about 20% of the time, depending on how many retry attempts are needed)

Azquelt:
Could we also clarify what the expected behaviour is? I note that it wasn't recorded in the issue.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryTest.java
Azquelt: On the other hand, if you're expecting the jitter to be evenly distributed between 0 and 400ms, then you're probably safe.

brunobat: Yes, there some risk here. My Idea is that jitter is evenly distributed between -400ms and 400ms but all values <0 will be 0. That leads to ~50% of the values being between  0 and 400ms. The 60% means that, 40% of the values will be between 5ms and 400ms... This can be fine tuned.

brunobat:
I will add enhancements to the documentation.

Azquelt:
I'm not sure that this behaviour is the most intuitive or helpful, but I am satisfied that this test ensures we get some jitter when `delay = 0` and that the javadoc changes do clarify what's expected.

api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java
```diff
@@ -81,6 +81,8 @@
     /**
      * Set the jitter to randomly vary retry delays for. The value must be greater than or equals to 0. 
      * 0 means not set.
+     * The effective delay will be [delay - jitter, delay + jitter] and allways greater than or equal to 0.
```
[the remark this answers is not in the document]
brunobat: Thanks Andrew. I just fixed it.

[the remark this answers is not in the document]
Emily-Jiang: misspelled `always`

PR #331, opened 2018-10-22
Make failRetryOnceThenFallback match its fallback method

Azquelt:
A previous change updated the return type of DisableAnnotationClient but
did not change the return type of its fallback method. This is an
invalid configuration and causes a FaultToleranceDefinitionException at
startup, causing tests to fail when they shouldn't.


genie-microprofile:

Issue #313, opened 2018-09-19
Specify interactions with AroundInvoke interceptors


Azquelt:
How does Fault Tolerance interact with `@AroundInvoke` interceptors?

* Does fault tolerance happen before other interceptors?
should it be lib_after(3000, open liberty) or lib_before (1000, thorntail), safeguard (400)?

We agree on to set the initial value of platform_after + 10 (4010)
Transactional - 200

@Retry
@MyLog(3)
@MyPrint(4)
public void myInvok() {

...
}
1. log->print->FT->method
most desired sequence.
2. FT->log->print->method

3. log->FT->print->method

* after?
* Is it itself in an interceptor with a specific priority?
  * Is the priority configurable?
* How do we expect other interceptors to be called when used with
  * `@Retry` (interceptors later in the chain get called more than once?)
  * `@Asynchronous`
    * Interceptors later in the chain get called on another thread?
    * Interceptors later in the chain see the result from the method?
    * Interceptors earlier in the chain see the result placeholder returned by fault tolerance?
    * Interceptors earlier in the chain may be called and return before the invocation of the method actually starts?
  * `@CircuitBreaker` (interceptors later in the chain may not get called at all?)
  * `@Timeout` (interceptors later in the chain are included in the the timeout timer, any results they return may get discarded in favor of a TimeoutException?)

Implementation note: If Fault Tolerance is implemented as an interceptor, then the InvocationContext needs to tolerate an interceptor that
* Doesn't call `proceed()` at all
* Calls `proceed()` more than once
* Calls `proceed()` from a different thread after the interceptor method has returned

Emily-Jiang:
mp.fault.tolerance.interceptor.priority=Priority.PLATFORM_AFTER+10 use this property to configure the priority if necessary

Emily-Jiang:
How do we expect other interceptors to be called when used with

    @Retry (see above)
    @Asynchronous
        Interceptors later in the chain get called on another thread? yes
        Interceptors later in the chain see the result from the method? yes
        Interceptors earlier in the chain see the result placeholder returned by fault tolerance? yes
        Interceptors earlier in the chain may be called and return before the invocation of the method actually starts? yes
    @CircuitBreaker (interceptors later in the chain may not get called at all?) yes
    @Timeout (interceptors later in the chain are included in the the timeout timer, any results they return may get discarded in favor of a TimeoutException?) yes

Emily-Jiang:
[...]

PR #332, opened 2018-10-23
Test to check the CDI and FT interceptors ordering

carlosdlr:


genie-microprofile:

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/ejb/EjbComponent.java
```diff
@@ -0,0 +1,51 @@
...
+
+import javax.ejb.Stateless;
+import javax.inject.Inject;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * An EJB component to show the ordering between EJB, CDI
+ * with FT annotations
+ */
+@Stateless
```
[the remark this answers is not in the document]
carlosdlr: Hello @Emily-Jiang The idea of this test is to validate the behavior of the different interceptors EJB and CDI interacting with MicroProfile interceptors to see if the order configured for MicroProfile can be affected by those interceptors.

Let me know if this test scope is ok or just validate against CDI.

Thanks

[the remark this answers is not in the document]
carlosdlr: @Emily-Jiang ok got it, I will modify  this test to allow it to be included in the TCK

Thank you

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/interceptor/LateFtInterceptor.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * An interceptor that is called after the FT interceptor
+ * in the chain and count the invocation.
+ * @author carlosdlr
+ */
+@Interceptor
+@InterceptLate
+@Priority(Interceptor.Priority.LIBRARY_AFTER)
```
Azquelt: In #313 we decided that the FT interceptor would have priority `PLATFORM_AFTER+10`. The LateFtInterceptor needs to have a priority larger than this for the test to work.

`LIBRARY_AFTER = 3000`, `PLATFORM_AFTER = 4000`

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/interceptor/FaultToleranceInterceptorTest.java
```diff
@@ -0,0 +1,93 @@
...
+            .addClasses(InterceptorComponent.class, EarlyFtInterceptor.class, LateFtInterceptor.class, CounterFactory.class)
+             .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml")
+            .as(JavaArchive.class);
+
+        return ShrinkWrap.create(WebArchive.class, "interceptorFtCdi.war")
+            .addAsLibrary(testJar);
+    }
+
+    @Test
+    public void testEjbAsync() throws InterruptedException, ExecutionException {
```
Azquelt: This test was originally written just to test that async works at all with EJBs. It doesn't actually use interceptors at all.

We need to take it out, or change it so that it's testing async with interceptors (e.g. testing that the interceptors fire in the correct order and on the correct threads)

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/interceptor/FaultToleranceInterceptorTest.java
```diff
@@ -0,0 +1,93 @@
...
+        Future<String> result = testInterceptor.asyncGetString();
+        assertEquals(result.get(), "OK");
+    }
+
+    @Test
+    public void testEjbRetryInterceptors() {
```
Azquelt: Remove "Ejb" from the name of this test.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/interceptor/LateFtInterceptor.java
carlosdlr: @Azquelt But this FT ordering is already implemented? because if you executed this test, for example, in liberty container the asserts are working good. Can you help me to clarify this?
Thanks

Emily-Jiang: @carlosdlr We have to update the initial interceptor order when we implement FT 2.0. The tests you are adding is for FT 2.0.

Azquelt: I originally wrote the test against the OpenLiberty 1.1 implementation, which set the priority as `LIBRARY_BEFORE`. So the interceptor priorities were correct for OpenLiberty's FT 1.1 support, but, as Emily said, they're not correct for what we specified for 2.0.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/interceptor/FaultToleranceInterceptorTest.java
carlosdlr: this was fixed

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/interceptor/FaultToleranceInterceptorTest.java
carlosdlr: this was fixed and the test for the async annotation was added.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/interceptor/LateFtInterceptor.java
carlosdlr: This was fixed changing the later interceptor priority to PLATFORM_AFTER+100 after the FT priority which is  PLATFORM_AFTER+10

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/interceptor/EarlyFtInterceptor.java
```diff
@@ -0,0 +1,74 @@
...
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * An interceptor that is called before the FT interceptor
+ * in the chain and count the invocation.
+ * @author carlosdlr
+ */
+@Interceptor
+@InterceptEarly
+@Priority(Interceptor.Priority.LIBRARY_BEFORE - 1) // Before FT Interceptor
```
Emily-Jiang: This should be PLATFORM_AFTER+30

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/interceptor/EarlyFtInterceptor.java
```diff
@@ -0,0 +1,74 @@
...
+public class EarlyFtInterceptor {
+
+    @Inject
+    @CounterId("EarlyFtInterceptor")
+    private AtomicInteger counter;
+
+    @Inject
+    @OrderId("EarlyOrderFtInterceptor")
+    private Queue<String> order;
```
Emily-Jiang: Where is this order used?

Azquelt:
Still not sure about these changes, the Queue is still being updated at injection time, not execution time and I'm not sure that the order of creation of the interceptors is defined.

Suggestions:
 * Make the queue producer method `@ApplicationScoped` instead of dependent - we've only got one queue, so there's no need for it to be dependant.
 * Remove the `@CounterId` annotation from the queue producer method and from everywhere the queue is injected
 * Inside the `@AroundInvoke` method of each interceptor, call `queue.add("<name of interceptor>");`
 * Inside the bean, call `queue.add("<name of bean>");`
 * In the test class, add a new method annotated with `@BeforeMethod` and in that method call `queue.clear()`
 * In the test method, assert that the contents of the queue is what we expect (EarlyInterceptor, LateInterceptor, BeanMethod)

To be honest, the way you're heading towards with using a queue to record the order in which interceptors are called is probably better and simpler than the test I originally wrote which used the counters. We could probably rewrite the retry test as well to use the queue instead of the counters.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/interceptor/FaultToleranceInterceptorTest.java
```diff
@@ -84,27 +67,34 @@ public static WebArchive deploy() {
...
-        assertEquals(methodCounter.get(), 6, "methodCounter"); // Method called six times (1 + 5 retries)
-        assertEquals(earlyInterceptorCounter.get(), 2, "earlyInterceptorCounter");
-        assertEquals(lateInterceptorCounter.get(), 7, "lateInterceptorCounter");
+        String [] expectedOrder = {"EarlyOrderFtInterceptor","LateOrderFtInterceptor","serviceRetryA",
+            "LateOrderFtInterceptor","serviceRetryA"}; //executes 1 more time the later interceptor and the bean method
+        // annotated with maxRetries = 1
+        assertEquals(orderFactory.getOrderQueue().toArray(), expectedOrder);
+    }
+
+    @AfterTest
```
Azquelt: I think you want `@AfterMethod` here rather than `@AfterTest`

A "test" in TestNG is a larger grouping which can include lots of test classes (I find this super-confusing because I'm used to thinking about each method as a "test")

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/interceptor/OrderQueueProducer.java
```diff
@@ -0,0 +1,46 @@
...
+public class OrderQueueProducer {
+
+    private final Queue<String> orderQueue = new ConcurrentLinkedQueue<>();
+
+    /**
+     * producer method that will be called by other beans when is needed
+     * @return order queue
+     */
+    @Produces
+    public synchronized Queue<String> getOrderQueue() {
```
Azquelt: We don't actually use this method as a producer - tests are calling this method directly - so you can remove the `@Produces` annotation and update the javadoc on the method and class.

Alternatively, you could use it as a producer and `@Inject Queue<String> orderQueue` everywhere you're currently doing `@Inject OrderQueueProducer orderFactory`.

I also don't think that this method needs to be `synchronized`.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/interceptor/FaultToleranceInterceptorTest.java
```diff
@@ -84,27 +67,34 @@ public static WebArchive deploy() {
...
         Future<String> result = testInterceptor.asyncGetString();
         assertEquals(result.get(), "OK");
-        orderKeeper.poll();
         String [] expectedOrder = {"EarlyOrderFtInterceptor","LateOrderFtInterceptor","asyncGetString"};
-        assertEquals(orderKeeper.toArray(), expectedOrder);
+        assertEquals(orderFactory.getOrderQueue().toArray(), expectedOrder);
     }
 
-    @Test
+    @Test(priority = 2)
```
Azquelt: Is the `priority` required? I can't see why it would be needed.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/interceptor/FaultToleranceInterceptorTest.java
```diff
@@ -0,0 +1,100 @@
...
+        return ShrinkWrap.create(WebArchive.class, "interceptorFtCdi.war")
+            .addAsLibrary(testJar);
+    }
+
+    /**
+     * This test validates the interceptors execution order after call a method
+     * annotated with Asynchronous FT annotation, using a queue type FIFO (first-in-first-out).
+     * The head of the queue is that element that has been on the queue the longest time.
+     * In this case is validating that the early interceptor is executed at first.
+     * @throws InterruptedException
```
brunobat: We will get compiler warnings if the description is missing on the Throws. Only these 2 are missing.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/interceptor/FaultToleranceInterceptorTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+    @Test
+    public void testRetryInterceptors() {
+        try {
+            testInterceptor.serviceRetryA();
+            fail("Exception not thrown");
+        }
+        catch (Exception e) {
+
+        } // Expected
```
brunobat: Please put the //expected to avoid the empty try/catch empty block.

carlosdlr: done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/interceptor/FaultToleranceInterceptorTest.java
carlosdlr: done

Issue #322, opened 2018-10-03
Does @Timeout clear the Thread interrupted flag when the method returns


Azquelt:
Then `@Timeout` is used, the thread is interrupted with `Thread.interrupt()`. Ideally the method checks this flag, or is running some waiting method which will throw an `InterruptedException`.

However, it's possible for the method to ignore this, which would result in the interrupted flag still being set when the method eventually returns.

Should we specify that fault tolerance should clear the flag when the method returns if a timeout has occurred?

Emily-Jiang:
The consensus on today's call is: we should clarify in the spec and add the tck tests.

Azquelt:
#333 adds this test for synchronous execution. It's not possible to test for asynchronous execution because you can't get hold of the work thread after fault tolerance has finished with it. I'll make a new PR to also update the spec.

PR #314, opened 2018-09-19
Specify interactions between annotations more completely

Azquelt:
Adds detail about how fault tolerance annotations interact and adds some
detail to @Asynchronous which was previously only present in the
Javadoc.

Addresses #291

PR #333, opened 2018-10-24
TCKs for interactions between Fault Tolerance annotations

Azquelt:
These are the TCKs to match the spec changes in #314 

There are quite a lot of tests here:
* Test `@Timeout` with methods which ignore thread interruption
  * Check that when using `@Asynchronous`, results are returned or retries are started as soon as the timeout fires, rather than waiting for the method body to return
  * Check that although results are returned when the timeout fires, the method continues to count as running by the bulkhead until it actually returns
* Add tests for `@Bulkhead` and `@Retry`
  * Check that each retry _attempt_ checks that the bulkhead is not full before running and releases its place when it ends
  * Check that when using `@Asynchronous`, retry attempts join the back of the bulkhead queue
* Add tests for `@CircuitBreaker` and `@Bulkhead`
  * Check that the circuit breaker is checked before attempting to enter the bulkhead
  * Check that a `BulkheadException` is counted as a circuit breaker failure, as long as `failOn` includes `BulkheadException` (or a superclass)
* Extend existing tests for `CircuitBreaker` and `Retry`
  * Check that `CircuitBreakerOpenException`s are retried, resulting in a successful execution if the circuit closes while retry attempts are being made
  * Check that `Retry` only retries `CircuitBreakerOpenException` if `retryOn` includes `CircuitBreakerOpenException` and `abortOn` does not.
* Update existing `@Bulkhead` tests to conform with the new spec
  * Particularly around how exceptions are thrown when using `@Asynchronous`
  * This involved some refactoring to ensure we check the result of every method call because exceptions are now thrown from the returned `Future` or `CompletionStage`
    * This in turn exposed some tests which weren't working correctly and required fixing
* Add tests that `Retry` only retries methods which fail due to timeouts if it would retry a `TimeoutException` (based on `retryOn` and `abortOn`)
* Add tests that `Retry` only retries methods which fail due to a full bulkhead if it would retry a `BulkheadException` (based on `retryOn` and `abortOn`)
* Add tests that `CircuitBreaker` only considers a method that times out to be a failure if it considers a `TimeoutException` a failure (based on `failOn`)

In addition, I also made the following changes which were not directly related to the issue:
* Check that the calling thread is not interrupted when the method returns (this one is for #322 and fitted nicely with the other non-interruptable method timeout tests I was adding)
* Add an explicit test that a full bulkhead throws a `BulkheadException` (this may have been caught somewhere else, but I couldn't see a test that filled the bulkhead and actually checked for the bulkhead exception)
* Some refactoring to allow reuse of existing utility classes
* Improving the error reporting from some tests I had trouble debugging
* Fixing two tests by making the deployment non-testable for tests which were testing invalid applications (I think I messed this up when I added the tests originally, they don't pass without this change)

Final note: The first commit (08d9de9) is from #331, I'll remove it when that PR gets merged.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/CircuitBreakerBulkheadTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/CircuitBreakerBulkheadTest.java
@@ -1,0 +1,256 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+package org.eclipse.microprofile.fault.tolerance.tck;
+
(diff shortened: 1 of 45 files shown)


genie-microprofile:

Emily-Jiang:
At this week's Andrew ran through the tests. The tests look ok.

Issue #310, opened 2018-09-12
Can Fallback be used without Retry or CircuitBreaker?


Azquelt:
Can you use `@Fallback`, either on its own or in conjunction with `@Bulkhead`, `@Asynchronous` or `@Timeout`

Currently the spec initially says that `@Fallback` can be used once `@Retry` or `@CircuitBreaker` has failed enough. It then later says
> The fallback should be triggered when an exception occurs. For instance, `BulkheadException`, `CircuitBreakerOpenException`, `TimeoutException` should trigger the fallback.

In the TCK tests, `@Fallback` is always used with `@Retry`, except in `FallbackMetricBean` which is only trying to tests metrics.

Can we either specify that `@Fallback` can be used both alone and with any combination of other FT annotations, or define what should happen if it's used without `@Retry` or `@CircuitBreaker`?

Either way, we should add some TCKs.

kenfinnigan:
I've used `@Fallback` on its own, so that should definitely be made clear that that is permissable

Emily-Jiang:
we should add spec clarification and tck

Azquelt:
@brunobat is happy to work on this, is that ok with you @Emily-Jiang ?

Azquelt:
Looks like this has been completed, thanks @brunobat

PR #339, opened 2018-11-07
#310 - Add TCK test to validate that @Fallback can be used alone

brunobat:
Apart from the TCK tests I've added a clarification to the documentation.

### DIFF
--- a/spec/src/main/asciidoc/fallback.asciidoc
+++ b/spec/src/main/asciidoc/fallback.asciidoc
@@ -21,7 +21,9 @@
 
 [[fallback]]
 == Fallback
-Fallbacks are invoked once a Retry or CircuitBreaker has failed enough times.
+Fallback provides an graceful alternative execution in case of failure.
+
+Fallbacks are invoked when used alone, once a Retry or CircuitBreaker has failed enough times.
 
 For a Retry, Fallback is handled any time the Retry would exceed its maximum number of attempts.
 
--- a/spec/src/main/asciidoc/fallback.asciidoc
+++ b/spec/src/main/asciidoc/fallback.asciidoc
@@ -21,9 +21,9 @@
 
 [[fallback]]
 == Fallback
-Fallback provides an graceful alternative execution in case of failure.
+Fallback method is invoked when the method with @Fallback annotation completed exceptionally.
 
-Fallbacks are invoked when used alone, once a Retry or CircuitBreaker has failed enough times.
+Fallback annotation can be used on its own or together with other Fault Tolerance annotations.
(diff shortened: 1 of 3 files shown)


genie-microprofile:

spec/src/main/asciidoc/fallback.asciidoc
```diff
@@ -21,7 +21,9 @@
 
 [[fallback]]
 == Fallback
-Fallbacks are invoked once a Retry or CircuitBreaker has failed enough times.
+Fallback provides an graceful alternative execution in case of failure.
+
+Fallbacks are invoked when used alone, once a Retry or CircuitBreaker has failed enough times.
```
Emily-Jiang: This line should be reworded
Fallback method is invoked when the method with `@Fallback` annotation completed exceptionally. `Fallback` annotation can be used on its own or together with other Fault Tolerance annotations.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/FallbackTest.java
```diff
@@ -253,4 +256,38 @@ public void testFallbackMethodWithArgsSuccess() {
...
         
     }
+
+    /**
+     * Test that a Fallback service can work alone, without other annotations.<br>
+     * <p>
+     * ServiceA fallback is driven by a class implementing FallbackHandler. ServiceA is only called once.
+     */
+    @Test
+    public void testStandaloneClassLevelFallback() {
```
Emily-Jiang: The annotation `Fallback` is method level annotation only as it is less useful to be on class level.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/fallback/clientserver/FallbackOnlyClient.java
```diff
@@ -0,0 +1,64 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2017 Contributors to the Eclipse Foundation
```
Emily-Jiang: should be 2018

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/FallbackTest.java
Emily-Jiang: ah. You meant `Fallback` handler. please change the method name to be testStandaloneHandlerFallback

carlosdlr:
looks good @brunobat

PR #349, opened 2018-11-26
Spec that interrupted status must be cleared after timeout

### LINKED ISSUE #322: Does @Timeout clear the Thread interrupted flag when the method returns

Azquelt:
This avoids the possibility that a later call to `Thread.interrupted()`
returns `true` due to a timeout in an earlier method call.

Fixes #322


Azquelt:
TCKs were already added under #333.

Relevant test is `TimeoutUninterruptableTest.testTimeout()`

I could only work out how to test this for synchronous execution as we can't get the thread back when doing asynchronous execution.

PR #350, opened 2018-11-26
Clarify circuitbreaker metrics which report time spent in each state

### LINKED ISSUE #335: Clarify Gauge metrics added for @CircuitBreaker
These are `ft.<name>.circuitbreaker.open.total`, `ft.<name>.circuitbreaker.halfOpen.total` and `ft.<name>.circuitbreaker.closed.total`.

The spec states that _"the sum will give the total time that the circuit breaker has been active for"_. I'm not quite sure what _"active"_ means in this sentence. The circuit breakers are shared objects (i.e. their state can be modified from multiple threads) and they're "active" since they're created/initialized.

Let's assume that we have bean `Foo`:
```java
@Dependent
class Foo {
  @CircuitBreaker
  void ping() {
  }
}
```

T1: `Foo#ping()` -> CB is created and is CLOSED
T2: `ft.Foo.ping.circuitbreaker.closed.total` = T2 - T1; `ft.Foo.ping.circuitbreaker.halfOpen.total` = 0, `ft.Foo.ping.circuitbreaker.open.total` = 0, 

Are these metrics values correct?

Azquelt:
Fixes #335


genie-microprofile:

PR #351, opened 2018-11-26
Add test class to RetryTest deployment

Azquelt:
One of the test classes was missing which caused the deployed application to fail to start.


genie-microprofile:

PR #353, opened 2018-11-26
Shorten wait time on BulkheadMetricTest

### LINKED ISSUE #337: TCK - BulkheadMetricTest#bulkheadMetricHistogramTest() sleeps for 3 seconds?
See also https://github.com/eclipse/microprofile-fault-tolerance/blob/1.1.2/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/BulkheadMetricTest.java#L162

I don't think that any tck test should sleep for 3 seconds. And if it's really needed, it should be well described why it is needed and the value should be probably a configurable property so that it would be possible to scale the value depending on the machine performance.

Azquelt:
Waiting for 3s was longer than necessary

Fixes #337 


genie-microprofile:

PR #356, opened 2018-11-27
Specify and test the locations where fallbackMethods may be declared

### LINKED ISSUE #311: Can a fallbackMethod be non-public?
Currently the fallbackMethod example in the spec shows a private fallback method, however there are no TCK tests which have non-public fallback methods.

### LINKED ISSUE #328: Does a fallbackMethod need to be on the same class?
The spec currently says:
> 7.1.2. Specify the fallbackMethod
> This is used if the fallbackMethod is on the same class as the method to call fall back.

To me, this sounds like the fallbackMethod must be on the same class as the method which references it.

If this is the case, can we add some negative tests where we reference a fallbackMethod on a superclass and check that this returns a definition exception?

If this is not the case, can we answer the following:
* can the fallbackMethod be on a superclass?
  * can the fallbackMethod be a private method on a superclass?
  * can the fallbackMethod be a default method on an interface?
* can the fallbackMethod be on a subclass?

Azquelt:
Specify and test that fallback methods may be declared on the same class, a superclass or an implemented interface of the class which declares the method annotated with `@Fallback`.

Includes spec changes and new TCKs.

Fixes #328 
Fixes #311


genie-microprofile:

Azquelt:
Realised I missed out requirements on accessibility of fallback methods from the spec. I've pushed another commit to add that.

Azquelt:
Yes to adding _Test_ to all the test classes

Methods on interfaces are always public in Java 8 and can only be either public or private in Java 9+. For Java 8 we could test:
* Class method lists an interface method with a default implementation as a fallbackMethod (though we do already test that referencing an interface method implemented by a subclass works)
* Interface method with a default implementation is annotated with `@Fallback` and references another interface method as the fallbackMethod. I assume that annotating an interface method with an interceptor binding is allowed?

For Java9 we could additionally test:
* Interface method with a default implementation is annotated with `@Fallback` and references a private interface method as the fallbackMethod.
* Negative test that a Class method cannot reference a private method on an implemented interface.

I don't know if we have any way of writing tests that are only valid on certain Java levels.

I couldn't work out a way that it would be valid to have a fallbackMethod on an anonymous inner class as non-static inner classes cannot be beans.

Adding a test for an abstract method with generics should be fine.

brunobat:
About the default methods... 
I think we should stick with the defaults for the Java version of the current Microprofile version, which I think it's Java 8. 

About the anonymous inner class...
Let me think about it.

spec/src/main/asciidoc/fallback.asciidoc
```diff
@@ -83,8 +83,14 @@ This is used if the `fallbackMethod` is on the same class as the method to call
...
-The return type of `fallbackForServiceB` must be `String` and the argument list for `fallbackForServiceB` must be the same as `ServiceB`.
-Otherwise, the `FaultToleranceDefinitionException` exception will be thrown.
+
+When `fallbackMethod` is used a `FaultToleranceDefinitionException` will be thrown if any of the following constraints are not met:
+
+* The named fallback method must be on the same class, a superclass or an implemented interface of the class which declares the annotated method
+* The named fallback method must have the same parameter types as the annotated method (after resolving any type variables)
+* The named fallback method must have the same return type as the annotated method (after resolving any type variables)
+* If the named fallback method is not declared by the same class as the annotated method, it must not be `private`
+* If the named fallback method is not in the same package as the annotated method, it must be either `protected` or `public`
```
Emily-Jiang: see above. Can be deleted with the above suggestion.

Azquelt:
Having trouble getting the test cases using interface default methods to work. It seems like Weld is inconsistent about whether annotations on default interface methods are exposed to interfaces or not. I've raised [WELD-2551](https://issues.jboss.org/browse/WELD-2551).

spec/src/main/asciidoc/fallback.asciidoc
Azquelt: Done

Issue #323, opened 2018-10-09
TCK CircuitBreakerTest - test methods reusing CircuitBreakerClientDefaultSuccessThreshold#serviceA()


mkouba:
Specifically `testCircuitInitialSuccessDefaultSuccessThreshold()`, `testCircuitLateSuccessDefaultSuccessThreshold()` and `testCircuitDefaultSuccessThreshold()`. 

This is not correct as the tests may affect each other. E.g. `testCircuitDefaultSuccessThreshold()` leaves the CB OPEN... 

Note that there is no portable way to reset the CB state for a particular bean/method. 

The bean could be reused but every test should be performed in a different deployment so that it is certain that the CB is in its inital state.

Azquelt:
@carlosdlr will update these tests

Emily-Jiang:
thanks @carlosdlr !

PR #363, opened 2018-12-04
TCK CircuitBreakerTest - test methods reusing CircuitBreakerClientDefaultSuccessThreshold

carlosdlr:
TCK CircuitBreakerTest CircuitBreakerClientDefaultSuccessThreshold Refactor

https://github.com/eclipse/microprofile-fault-tolerance/issues/323

The tests testCircuitInitialSuccessDefaultSuccessThreshold and
testCircuitLateSuccessDefaultSuccessThreshold were moved to an independent
test to avoid dependency between tests that uses the same bean in this
case CircuitBreakerClientDefaultSuccessThreshold, that will can generate
possible failures when the circuit breaker leaves open.
To ensure that each of these tests uses a fresh circuit breaker was created
an independent deployment for this purpose.

Signed-off-by: carlos de la rosa <kusanagi12002@gmail.com>


genie-microprofile:

Issue #246, opened 2018-04-09
Asynchronous execution with retry policy - clarify if a Future that completes exceptionally triggers a retry


mkouba:
This issue is related to #245. See also the following example. Note that the `Future` might be actually obtained from another part of the app.

````java
@Asynchronous
@Retry(maxRetries = 2)
public Future<String> retry() throws IOException {
   CompletableFuture<String> future = new CompletableFuture<>();
   future.completeExceptionally(new IOException("Simulated error"));
   return future;
}
````

Emily-Jiang:
It should trigger the retry if a Future that completes exceptionally. We should add tests for this.

mkouba:
If so, the spec should clarify that the `Future` returned is an internal construct which delegates to a `Future` returned from the method.

OndrejM:
We should also mandate that if FT annotations are applied on an `@Asynchronous` method, they finish the future exceptionally instead of throwing an exception from the method. As discussed in #250

ggam:
Any chance to get this for the next version? Can I help here?

Emily-Jiang:
Let's try to get this done for FT 2.0.

Emily-Jiang:
@brunobat will take a look to contribute some tcks.

Azquelt:
We discussed this a little on the hangout this week and I don't think this is possible for methods which return a `Future`.

There's no way to trigger something to run when a `Future` completes, you can only probe it to see if it's done yet. If we wanted to apply `Retry` around the completion of a Future, we'd either have to dedicate a thread to waiting for the Future to finish, have a background task which periodically polls Futures to see if they're done yet or only actually do the Retry when someone calls `get()`. None of those options seem like a good idea.

For a `CompletionStage` it's possible because we can add an action to be run after the `CompletionStage` completes.

OndrejM:
Hi, I think that we can easily support Future with the same semantics as for synchronous executions after `Future.get()` is called. I explain it https://github.com/eclipse/microprofile-fault-tolerance/issues/362#issuecomment-448502738.

I wasn't on the hangout on 28 Nov so I may be missing some arguments. But the argument in the meeting minutes isn't really valid: "Future that ends in an exception cannot be retried because the get() can only be called once". This is simply because a failed call can be executed within the Future, within a single `get()` call. However, as I explained in my comment above, this may not be worth to specify and implement and a deprecating `Future` is also a good option.

Emily-Jiang:
[...]

PR #371, opened 2018-12-12
#246 - TCK for Asynchronous execution with retry policy - part1

brunobat:


brunobat:
Add tests in cases were we return CompletionStage in absence of the `@Asynchronous` annotation in the method.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryConditionTest.java
```diff
@@ -215,4 +227,144 @@ public void testClassLevelRetryWithAbortOnTrue() {
...
+    /**
+     * Persistent Error condition inside a CompletableFuture.
+     * Will not retry because method is not marked with @Asynchronous
+     * ServiceB will always complete exceptionally with IOException.
+     */
+    @Test
+    public void testNoAsynWilNotRetryExceptionally() {
+        CompletableFuture<String> future = new CompletableFuture<>();
+
+        asyncRetryClient.serviceBFailExceptionally(future);
```
Azquelt: What's the reason for passing in `future` here, rather than creating it inside `serviceBFailExceptionally`?

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryConditionTest.java
```diff
@@ -215,4 +227,144 @@ public void testClassLevelRetryWithAbortOnTrue() {
...
+
+        assertCompleteExceptionally(future, IOException.class, "Simulated error");
```
Azquelt: Why does this assert on the value of `future` which was supplied as an argument, rather than the value returned from `serviceBFailExceptionally`?

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryConditionTest.java
```diff
@@ -215,4 +227,144 @@ public void testClassLevelRetryWithAbortOnTrue() {
...
+    /**
+     * Temporary error. Will retry 3 times, the first 2 will fail.
+     * ServiceC uses {@link org.eclipse.microprofile.faulttolerance.Asynchronous}.
+     */
+    @Test
+    public void testRetrySuccess() {
+        final CompletionStage<String> future = asyncRetryClient.serviceC();
+
+        assertCompleteOk(future, "Success");
+        assertEquals(2, asyncRetryClient.getCountInvocationsServC());
```
Azquelt: Looking at `serviceC` I think we expect 3 invocations here: 2 failures and 1 success

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryConditionTest.java
```diff
@@ -215,4 +227,144 @@ public void testClassLevelRetryWithAbortOnTrue() {
...
+    /**
+     * Temporary error. Will retry 3 times, the first 2 will fail deep in a CompletableFuture chained execution.
+     * ServiceD uses {@link org.eclipse.microprofile.faulttolerance.Asynchronous} and chains 2 CompletableFutures.
+     */
+    @Test
+    public void testRetryChainSuccess() {
+        final CompletionStage<String> future = asyncRetryClient.serviceD();
+
+        assertCompleteOk(future, "Success");
+        assertEquals(2, asyncRetryClient.getCountInvocationsServD());
```
Azquelt: I think we expect 3 invocations here too (2 failing, 1 succeeding)

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryConditionTest.java
```diff
@@ -215,4 +227,144 @@ public void testClassLevelRetryWithAbortOnTrue() {
...
+
+    /**
+     * Temporary error. Will retry 3 times, the first 2 will fail in a CompletableFuture parallel execution.
+     * ServiceG uses {@link org.eclipse.microprofile.faulttolerance.Asynchronous} and 2 CompletableFutures.
+     */
+    @Test
+    public void testRetryParallelSuccess() {
+        CompletableFuture<String> future = new CompletableFuture<>();
+
+        asyncRetryClient.serviceG();
```
Azquelt: Should this be `serviceF`?

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryConditionTest.java
```diff
@@ -215,4 +227,144 @@ public void testClassLevelRetryWithAbortOnTrue() {
...
+
+        assertCompleteOk(future, "Success then Success");
+        assertEquals(2, asyncRetryClient.getCountInvocationsServG());
```
Azquelt: also `ServF` here

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryConditionTest.java
```diff
@@ -46,7 +57,8 @@
     private @Inject RetryClientAbortOn clientForAbortOn;
     private @Inject RetryClassLevelClientRetryOn clientForClassLevelRetryOn;
     private @Inject RetryClassLevelClientAbortOn clientForClassLevelAbortOn;
-    
+    private @Inject AsyncRetryClient asyncRetryClient;
```
Azquelt: Add this class to the application in the `deploy()` method.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryConditionTest.java
```diff
@@ -215,4 +227,144 @@ public void testClassLevelRetryWithAbortOnTrue() {
...
+
+        assertCompleteOk(future, "Success then Success");
+        assertEquals(2, asyncRetryClient.getCountInvocationsServG());
+    }
+
+    private void assertCompleteExceptionally(final CompletionStage<String> future,
+                                             final Class<? extends Throwable> exceptionClass,
+                                             final String exceptionMessage) {
+        try {
+            future.toCompletableFuture().get(200, TimeUnit.MILLISECONDS);
```
Azquelt: `toCompletableFuture` is an optional method, we probably shouldn't rely on it.

There's already a utility in `CompletableFutureHelper.toCompletableFuture()` which creates a new completable future from a `CompletionStage` without using this method. (Remember to also add the class to the list in `deploy()`)

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asyncretry/clientserver/AsyncRetryClient.java
```diff
@@ -0,0 +1,242 @@
...
+
+import static java.util.Objects.nonNull;
+
+/**
+ * A client to demonstrate the combination of the @Retry and @Asynchronous annotations.
+ *
+ * @author <a href="mailto:bbaptista@tomitribe.com">Bruno Baptista</a>
+ */
+@RequestScoped
+public class AsyncRetryClient {
```
Azquelt: The javadoc for every method in this class says `Service A` rather than the actual method name. ;)

Might be better to just take the name out of the javadoc.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryConditionTest.java
```diff
@@ -215,4 +227,144 @@ public void testClassLevelRetryWithAbortOnTrue() {
...
+        }
+        catch (ExecutionException ee) {
+            Assert.assertTrue(exceptionClass.isInstance(ee.getCause()), "Cause of ExecutionException was " + ee.getCause());
+            assertEquals(exceptionMessage, ee.getCause().getMessage());
+        }
+    }
+
+    private void assertCompleteOk(final CompletionStage<String> future, final String expectedMessage) {
+        try {
+            assertEquals(expectedMessage, future.toCompletableFuture().get(200, TimeUnit.MILLISECONDS));
```
Azquelt: This wait is too short because `Retry` has a default jitter of 200ms.

Could we increase the timeout to a couple of seconds? (a working implementation should never hit this timeout anyway, so I think a slightly longer wait is justified)

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryConditionTest.java
```diff
@@ -215,4 +227,144 @@ public void testClassLevelRetryWithAbortOnTrue() {
...
+    /**
+     * Persistent Error condition. Will retry 2 times and still throw exception.
+     * ServiceA uses {@link org.eclipse.microprofile.faulttolerance.Asynchronous} and will always return IOException.
+     */
+    @Test
+    public void testAsyncRetryExceptionally() {
+        final CompletionStage<String> future = asyncRetryClient.serviceA();
+
+        assertCompleteExceptionally(future, IOException.class, "Simulated error");
+        assertEquals(2, asyncRetryClient.getCountInvocationsServA());
```
Azquelt: 2 retries -> 3 invocations

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryConditionTest.java
```diff
@@ -215,4 +227,144 @@ public void testClassLevelRetryWithAbortOnTrue() {
...
+     */
+    @Test
+    public void testNoAsynWilNotRetryExceptionally() {
+        CompletableFuture<String> future = new CompletableFuture<>();
+
+        asyncRetryClient.serviceBFailExceptionally(future);
+
+        assertCompleteExceptionally(future, IOException.class, "Simulated error");
+        // no retries
+        assertEquals("No retries are expected",0, asyncRetryClient.getCountInvocationsServBFailExceptionally());
```
Azquelt: No retries, but we still expect one invocation.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryConditionTest.java
```diff
@@ -215,4 +227,144 @@ public void testClassLevelRetryWithAbortOnTrue() {
...
+
+        try {
+            asyncRetryClient.serviceBFailException(future);
+            fail("Was expecting an exception");
+        }
+        catch (RuntimeException e) {
+            assertEquals("Simulated error", e.getMessage());
+        }
+        // 2 retries
+        assertEquals(2, asyncRetryClient.getCountInvocationsServBFailException());
```
Azquelt: 2 retries -> 3 invocations

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryConditionTest.java
brunobat: It's just an additional scenario where an uncompleted future is passes as a parameter. Probably it doesn't add much.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryConditionTest.java
brunobat: In this case it has the same outcome because the object reference is the same, but I agree the the returned value should be tested. Thanks.

brunobat: A few other test have a problem related to this. Issuing a fix

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asyncretry/clientserver/AsyncRetryClient.java
```diff
@@ -190,6 +191,26 @@
...
+     *
+     * @return a {@link CompletionStage}
+     */
+    @Asynchronous
+    @Retry(maxRetries = 2)
+    public CompletionStage<String> serviceH() {
+        countInvocationsServH++;
+        // fails twice
+        if (countInvocationsServH < 2) {
+            throw new RuntimeException("Simulated error");
```
Azquelt: I think this only fails once

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryConditionTest.java
```diff
@@ -281,7 +281,7 @@ public void testNoAsynRetryOnMethodException() {
     }
 
     /**
-     * Temporary error. Will retry 3 times, the first 2 will fail.
+     * Temporary error. Will retry 2 times, the first 2 calls fail will fail.
```
Azquelt: End of this line doesn't make sense.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryConditionTest.java
```diff
@@ -293,7 +293,7 @@ public void testRetrySuccess() {
     }
 
     /**
-     * Temporary error. Will retry 3 times, the first 2 will fail deep in a CompletableFuture chained execution.
+     * Temporary error. Will retry 2 times, the first 2 calls fail fail deep in a CompletableFuture chained execution.
```
Azquelt: "fail fail"

PR #374, opened 2018-12-17
373 - Change system time to use nanoseconds

benjamin-confino:
As per the java documentation nanoTime is the appropriate method for checking the duration of a running	thread. Using nanoTime will also protect the tests from any changes to the system clock while the tck is running.

Signed-off-by: Benjamin Confino <benjamic@uk.ibm.com>


genie-microprofile:

PR #378, opened 2018-12-20
Test CDI and FT interceptors ordering changing FT interceptor priority

carlosdlr:
FT interceptor priority issue 313

https://github.com/eclipse/microprofile-fault-tolerance/issues/313

New tests added to validate the CDI interceptors against FT interceptor
when the FT interceptor priority is changed and the CDI interceptors
are injected to the context using xml or annotation base configuration.

Signed-off-by: carlos de la rosa <kusanagi12002@gmail.com>

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/interceptor/ftPriorityChange/FaultToleranceInterceptorPriorityChangeAnnotationConfTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/interceptor/ftPriorityChange/FaultToleranceInterceptorPriorityChangeAnnotationConfTest.java
@@ -19,9 +19,6 @@
  *******************************************************************************/
 package org.eclipse.microprofile.fault.tolerance.tck.interceptor.ftPriorityChange;
 
-import org.eclipse.microprofile.fault.tolerance.tck.interceptor.EarlyFtInterceptor;
-import org.eclipse.microprofile.fault.tolerance.tck.interceptor.InterceptorComponent;
-import org.eclipse.microprofile.fault.tolerance.tck.interceptor.LateFtInterceptor;
 import org.eclipse.microprofile.fault.tolerance.tck.interceptor.OrderQueueProducer;
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.testng.Arquillian;
@@ -57,8 +54,8 @@ public class FaultToleranceInterceptorPriorityChangeAnnotationConfTest extends A
         JavaArchive testJar = ShrinkWrap
             .create(JavaArchive.class, "interceptorChangeFtAnnotation.jar")
             .addClasses(InterceptorComponent.class, EarlyFtInterceptor.class, LateFtInterceptor.class, OrderQueueProducer.class)
-             .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml")
-            .addAsManifestResource(new StringAsset("mp.fault.tolerance.interceptor.priority=2100"), "microprofile-config.properties")
+            .addAsManifestResource(new StringAsset("mp.fault.tolerance.interceptor.priority=3700"), "microprofile-config.properties")
+            .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml")
             .as(JavaArchive.class);
 
         return ShrinkWrap.create(WebArchive.class, "interceptorChangeFtAnnotation.war")
@@ -77,8 +74,8 @@ public class FaultToleranceInterceptorPriorityChangeAnnotationConfTest extends A
     public void testAsync() throws InterruptedException, ExecutionException {
         Future<String> result = testInterceptor.asyncGetString();
         assertEquals(result.get(), "OK");
-        String [] expectedOrder = {"EarlyOrderFtInterceptor","LateOrderFtInterceptor","asyncGetString"};
-        /*After changing the FT interceptor priority the order continues working in the same way using annotated interceptors injection*/
+        String [] expectedOrder = {"asyncGetString","LateOrderFtInterceptor","EarlyOrderFtInterceptor"};
+        /*After changing the FT interceptor priority the first interceptor to be called must be FT interceptor*/
         assertEquals(orderFactory.getOrderQueue().toArray(), expectedOrder);
     }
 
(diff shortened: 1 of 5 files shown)


genie-microprofile:

Azquelt:
If I'm reading this right, in both cases (beans.xml and Priority annotation) you're changing the priority of the Fault Tolerance interceptor(s) in a way where you expect the same behaviour as when you didn't change the priority of it?

Could you change `FaultToleranceInterceptorPriorityChangeAnnotationConfTest` so that you set the FT interceptor priority in such a way that it changes the expected behaviour? (e.g. setting it to a lower value than the `EarlyFtInterceptor` or to a higher value than the `LateFtInterceptor`)

The other test is fine since when the interceptors are defined in the `beans.xml`, the priority of the FT interceptor shouldn't make any difference.

Azquelt:
@genie-microprofile please test this

carlosdlr:
Hello @Azquelt yes you are right I will make

Thanks

Emily-Jiang:
@carlosdlr can you please make the change as indicated so that we can merge this PR? thanks

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/interceptor/ftPriorityChange/FaultToleranceInterceptorPriorityChangeAnnotationConfTest.java
```diff
@@ -0,0 +1,104 @@
...
+        try {
+            testInterceptor.serviceRetryA();
+            fail("Exception not thrown");
+        }
+        catch (Exception e) {
+            assertEquals(e.getMessage().trim(), "retryGetString failed");
+        }
+
+        String [] expectedOrder = {"serviceRetryA","EarlyOrderFtInterceptor","LateOrderFtInterceptor",
+            "serviceRetryA","EarlyOrderFtInterceptor","LateOrderFtInterceptor"};
```
Azquelt: Unless I missed something, I was expecting this to be `{"EarlyOrderFtInterceptor","LateOrderFtInterceptor", "serviceRetryA", "EarlyOrderFtInterceptor","LateOrderFtInterceptor", "serviceRetryA"};

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/interceptor/ftPriorityChange/FaultToleranceInterceptorPriorityChangeAnnotationConfTest.java
```diff
@@ -0,0 +1,104 @@
...
+     * The head of the queue is that element that has been on the queue the longest time.
+     * In this case is validating that the early interceptor is executed at first.
+     * @throws InterruptedException if the current thread was interrupted while waiting
+     * @throws ExecutionException if the computation threw an exception
+     */
+    @Test
+    public void testAsync() throws InterruptedException, ExecutionException {
+        Future<String> result = testInterceptor.asyncGetString();
+        assertEquals(result.get(), "OK");
+        String [] expectedOrder = {"asyncGetString","EarlyOrderFtInterceptor","LateOrderFtInterceptor"};
```
Azquelt: Similarly, should this not be `{"EarlyOrderFtInterceptor","LateOrderFtInterceptor", "asyncGetString"}` ?

carlosdlr: Hello @Azquelt,

This is not the case because the idea of this the is to validate that if the FT priority is changed to be called at first must work.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/interceptor/ftPriorityChange/FaultToleranceInterceptorPriorityChangeAnnotationConfTest.java
carlosdlr: Hello @Azquelt,

This is not the case because the idea of this the is to validate that if the FT priority is changed to be called at first must work.

Azquelt: Right, but even if the FT interceptor is called first, the method body is still called at the end.

The difference I expect to with the FT interceptor being first is that the interceptors are called on each retry, rather than only once at the start.

carlosdlr: Ok so for no make sense this test because as a developer if i change the FT priority i expect some change in the order, i think is better wait after the implementation to add this test. What do you think?

Azquelt: It looks like we have the following:
```
EarlyOrderFtInterceptor - priority 3800
LateOrderFtInterceptor - priority 3900
FT interceptor base - default priority 4010, but configured here to be 3700
```

If the FT interceptor is left at the **default priority**, I would expect the order to be "Early, Late, FT".

Therefore, in the order queue, I would expect to see `{"EarlyOrderFtInterceptor","LateOrderFtInterceptor", "asyncGetString", "asyncGetString"}`

If the FT interceptor is configured to **priority 3700**, I would then expect the order to be "FT, Early, Late".

Therefore, in the order queue, I would expect to see `{"EarlyOrderFtInterceptor","LateOrderFtInterceptor", "asyncGetString", "EarlyOrderFtInterceptor","LateOrderFtInterceptor", "asyncGetString"}`

The change is expected here because I now expect the interceptors to be called on each retry attempt.

carlosdlr: Ok @Azquelt Thanks for your clarification

Azquelt:
Thanks @carlosdlr  :)

Issue #342, opened 2018-11-21
CircuitBreaker - clarify how requestVolumeThreshold() and rolling window work


mkouba:
Both the spec and the javadoc mention a "rolling window" but neither explains well how it should work.

If I understand it correcly, only the results of the last `n` requests (where `n=requestVolumeThreshold()`) should be considered when calculating a circuit health. And the results should be reset when a circuit changed its state, i.e. for the following method:
```java
@CircuitBreaker(requestVolumeThreshold = 4, failureRatio = 0.5)
public Connection serviceA() {}
```
REQUEST 1 - ok
REQUEST 2 - ok
REQUEST 3 - ok
REQUEST 4 - fail
REQUEST 5 - fail --> circuit should open, because in the last 4 requests 2 requests failed (REQUEST 1 is ignored).

Is this interpretation correct?

Emily-Jiang:
I'll add some spec clarification for this. yes to your question.

PR #379, opened 2018-12-21
#342 clarify the circuit breaker rolling window

Emily-Jiang:

### DIFF
--- a/spec/src/main/asciidoc/circuitbreaker.asciidoc
+++ b/spec/src/main/asciidoc/circuitbreaker.asciidoc
@@ -38,7 +38,37 @@ succeed before the circuit can be closed. After the specified number of successf
 before the successThreshold is reached the circuit will transition to open. 
 
 Note that circuit state transitions will reset the Circuit Breaker's records. For example, when the circuit transitions to closed a new
-rolling failure window is created with the configured requestVolumeThreshold and failureRatio.
+rolling failure window is created with the configured requestVolumeThreshold and failureRatio. The circuit state will only be assessed when the rolling window reaches the `requestVolumeThreshold`. 
+ 
+[source, java]
+----
+@CircuitBreaker(successThreshold = 10, requestVolumeThreshold = 4, failureRatio=0.5, delay = 1000)
+public Connection serviceA() {
+   Connection conn = null;
+   counterForInvokingServiceA++;
+   conn = connectionService();
+   return conn;
+}
+----
+Below are a couple scenarios to demonstrate when the circuit opens for the above code snippet.
+* Scenario 1
+     Request 1 - success
+     Request 2 - failure
+     Request 3 - failure
+     Request 4 - success
+     Request 5 - CircuitOpenException
+
+The Request 4 will trigger the circuit to open. The Request 5 will hit the exception.
+
+* Scenario 2
+     Request 1 - success
+     Request 2 - failure
+     Request 3 - success
+     Request 4 - success
+     Request 5 - failure
+     Request 6 - CircuitOpenException
+
+The Request 5 will trigger the circuit to open. The Request 6 will hit the exception.
 
 === Circuit Breaker Usage
 
--- a/spec/src/main/asciidoc/circuitbreaker.asciidoc
+++ b/spec/src/main/asciidoc/circuitbreaker.asciidoc
@@ -51,22 +51,23 @@ public Connection serviceA() {
 }
 ----
 Below are a couple scenarios to demonstrate when the circuit opens for the above code snippet.
+
 * Scenario 1
-     Request 1 - success
-     Request 2 - failure
-     Request 3 - failure
-     Request 4 - success
-     Request 5 - CircuitOpenException
+** Request 1 - success
+** Request 2 - failure
+** Request 3 - failure
+** Request 4 - success
+** Request 5 - CircuitBreakerOpenException
 
 The Request 4 will trigger the circuit to open. The Request 5 will hit the exception.
 
 * Scenario 2
-     Request 1 - success
-     Request 2 - failure
-     Request 3 - success
-     Request 4 - success
-     Request 5 - failure
-     Request 6 - CircuitOpenException
+** Request 1 - success
+** Request 2 - failure
+** Request 3 - success
+** Request 4 - success
+** Request 5 - failure
+** Request 6 - CircuitBreakerOpenException 
 
 The Request 5 will trigger the circuit to open. The Request 6 will hit the exception.
 
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/CircuitBreakerTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/CircuitBreakerTest.java
@@ -647,7 +647,7 @@ public class CircuitBreakerTest extends Arquillian {
         for (int i = 1; i < 6; i++) {
 
             try {
-                clientForRW.serviceARollingWindowOpenAfter4();
+                clientForRW.service1RollingWindowOpenAfter4();
             }
             catch (CircuitBreakerOpenException cboe) {
                 // Expected on iteration 5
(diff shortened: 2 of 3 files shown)


tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/circuitbreaker/clientserver/CircuitBreakerClientRollingWindow.java
```diff
@@ -0,0 +1,89 @@
...
+ * A client to exercise Circuit Breaker thresholds, with a SuccessThreshold of 2,
+ * a requestVolumeThreshold of 4, failureRatio of 0.5 and a 1 millisecond delay
+ *
+ * @author <a href="mailto:emijiang@uk.ibm.com">Emily Jiang</a>
+ */
+@RequestScoped
+public class CircuitBreakerClientRollingWindow implements Serializable {
+        
+    private static final long serialVersionUID = 1L;
+    private int counterForInvokingServiceA = 0;
```
Azquelt: Issue #323 requested that we don't reuse circuit breakers between tests. (We were expecting each test to have it's own new request scope and for that to mean we got a new circuit breaker instance).

I wonder if that means we should avoid relying on each test having a fresh request scope? If so we should have one counter for each test method here. (or we should reset the counter to zero at the start of each test, but that seems more prone to error)

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/CircuitBreakerTest.java
```diff
@@ -624,4 +627,98 @@ public void testClassLevelCircuitOverrideNoDelay() {
...
+            catch (Exception ex) {
+                // Not Expected
+                Assert.fail("serviceA should throw a RuntimeException or CircuitBreakerOpenException in testRollingWindowCircuitOpen2 "
+                        + "on iteration " + i);
+            }
+        }
+        int serviceAExecutions = clientForRW.getCounterForInvokingServiceA();
+        Assert.assertTrue(cbExceptionThrown, "The CircuitBreaker exception in testRollingWindowCircuitOpen2 was not thrown correctly.");
+
+        Assert.assertEquals(serviceAExecutions, 6, "The number of executions should be 6");
```
Azquelt: I think we expect 5 executions here, the sixth throws a CircuitBreakerOpenException and the method body is not executed.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/CircuitBreakerTest.java
```diff
@@ -624,4 +627,98 @@ public void testClassLevelCircuitOverrideNoDelay() {
...
+            catch (Exception ex) {
+                // Not Expected
+                Assert.fail("serviceA should throw a RuntimeException or CircuitBreakerOpenException in testRollingWindowCircuitOpen "
+                     + "on iteration " + i);
+            }
+        }
+        int serviceAExecutions = clientForRW.getCounterForInvokingServiceA();
+        Assert.assertTrue(cbExceptionThrown, "The CircuitBreaker exception in testRollingWindowCircuitOpen was not thrown correctly.");
+
+        Assert.assertEquals(serviceAExecutions, 5, "The number of executions should be 5");
```
Azquelt: I think we expect 4 executions here, the fifth throws a CircuitBreakerOpenException and the method body is not executed.

spec/src/main/asciidoc/circuitbreaker.asciidoc
```diff
@@ -38,7 +38,37 @@ succeed before the circuit can be closed. After the specified number of successf
...
+   return conn;
+}
+----
+Below are a couple scenarios to demonstrate when the circuit opens for the above code snippet.
+* Scenario 1
+     Request 1 - success
+     Request 2 - failure
+     Request 3 - failure
+     Request 4 - success
+     Request 5 - CircuitOpenException
```
Azquelt: Exception name is `CircuitBreakerOpenException`. Also below.

Azquelt: This section does not render correctly in the spec pdf

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/circuitbreaker/clientserver/CircuitBreakerClientRollingWindow.java
Emily-Jiang: ah. My c+p mistake. I should have used two different counters. fixed now.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/CircuitBreakerTest.java
Emily-Jiang: yep. thanks.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/CircuitBreakerTest.java
Emily-Jiang: yep. thanks

spec/src/main/asciidoc/circuitbreaker.asciidoc
Emily-Jiang: It seems my asciidoc preview rendered ok but not the generated one. fixed.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/circuitbreaker/clientserver/CircuitBreakerClientRollingWindow.java
```diff
@@ -0,0 +1,92 @@
...
+
+    public int getCounterForInvokingService1() {
+           return counterForInvokingService1;
+    }
+
+    public int getCounterForInvokingService2() {
+        return counterForInvokingService2;
+ }
+
+    @CircuitBreaker(successThreshold = 2, requestVolumeThreshold = 4, failureRatio=0.5, delay = 1)
```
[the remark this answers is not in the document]
Emily-Jiang: removed

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/circuitbreaker/clientserver/CircuitBreakerClientRollingWindow.java
```diff
@@ -0,0 +1,92 @@
...
+        if((counterForInvokingService1 ==2 ) || (counterForInvokingService1 ==3)) {
+                throw new RuntimeException("Connection failed");
+        }
+        return null;
+    
+    
+    }
+    //simulate a backend service
+    //Throw exception for the 2nd and 5th request.
+    @CircuitBreaker(successThreshold = 2, requestVolumeThreshold = 4, failureRatio=0.5, delay = 1)
```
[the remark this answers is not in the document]
Emily-Jiang: fixed. thanks

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/circuitbreaker/clientserver/CircuitBreakerClientRollingWindow.java
brunobat: That private counterForInvokingServiceA attribute is counting calls to the same bean instance. 
Different requests scoped should see different CircuitBreakerClientDefaultSuccessThreshold objects.
In order to be effective that attribute would need to be static.
If we are seeing a non static attribute being incremented, it means its being called from the same context.
This makes me think the all the requests are being made from the same scope making the test ineffective. 
What do you guys think?
In order to be

Azquelt: @brunobat I don't think I'm quite following you. You're correct in that the counter will only count calls to the same instance, but the test is making all of its calls against a single instance, so I'd expect it to work. Within one test method, I'd only expect there to be one request context, so we'd only get one instance of the request scoped bean.

The original code assumed that each test method will have a different request context as it relied on each test using a different instance to ensure that the counter was zero at the start of the test. This is true in most cases, but some embedded containers will run all tests in the same request context.

In the current code, `counterForInvokingService1` counts invocations of `service1RollingWindowOpenAfter4()`. `testRollingWindowCircuitOpen()` makes five requests to this method against the same instance of `CircuitBreakerClientRollingWindow`, and then checks the counter.

brunobat: Ok, I was mixing `@RequestScoped` with `@ApplicationScoped` In this case the test uses `@RequestScope` as we have been doing.

Emily-Jiang:
this  closes #342

Issue #362, opened 2018-12-03
Abandon Future handling and replace it with CompletionStage


rmannibucau:
Today in java futures are most of the time CompletionStage (CompletionFutures), this is what the user can manipulate the most easily, this also enable any framework (MP or not, I'm thinking to DeltaSpike here) to integrate with the pipeline very easily.

Future on the other side does not enable it without reimplementing CompletionStage logic and creating some custom vendor specific API.

To rationalize this effort and promote java 8 I'd like fault-tolerance to replace its Future<?> usage by CompletionStage<?> which does not remove any feature from an user point of view?

Emily-Jiang:
For backward compatibility issues, we are not going to remove Future but adding support for CompletionStage. I image most developers will move towards CompletionStage.

rmannibucau:
@Emily-Jiang hmm, maybe I misunderstood but MP has no backward compatibility concern, metrics API was broken for instance (unit went from plain value to optional recently), is it really something taken into consideration?

Emily-Jiang:
Good idea. We will add a javadoc or spec statement to emphasize the preference for CompletionStage.

brunobat:
During the discussion of #368 the abandonment of Future looks inevitable due to implementation constraints.

Azquelt:
I don't think #368 necessitates abandoning Future.

We can keep the existing behaviour for Future where the fault tolerance logic is applied around the method call, but we can't apply the fault tolerance logic around the completion of the Future, like we can for CompletionStage.

I don't see any reason why we should break backwards compatibility for a feature that people may already be using.

brunobat:
Right, older methods using Future, will not take advantage of most Fault Tolerance functionalities. Something to document.

Emily-Jiang:
Will add some sentences later with a PR to express the preference of CompletionStage.

OndrejM:
Reviewing the spec, `Future` is basically handled only internally by the container and only when an app calls `Future.get()`, which has the same semantics as methods returning synchronously, except the app can choose when to apply those semantics. I think this is fine and covers `Future` enough, especially since most people would use `CompletionStage` instead of `Future` nowadays. Even `@Retry` can be supported like this, because once `Future.get()` is called, the thread is blocked and can wait for downstream futures and retry once they return. We just need to specify this more clearly or specify which annotations don't support the semantics if we decide to make implementers' life easier.

[...]

PR #381, opened 2019-01-01
#246 #362 clarify spec on Future as the reture type

Emily-Jiang:

### DIFF
--- a/spec/src/main/asciidoc/asynchronous.asciidoc
+++ b/spec/src/main/asciidoc/asynchronous.asciidoc
@@ -30,7 +30,7 @@ This thread should have the correct security context or naming context associate
 === Asynchronous Usage
 
 A method or a class can be annotated with `@Asynchronous`, which means the method or the methods under the class will be invoked by a separate thread.
-The method annotated with `@Asynchronous` must return a `Future` or a `CompletionStage` from the `java.util.concurrent` package. Otherwise, a `FaultToleranceDefinitionException` occurs.
+The method annotated with `@Asynchronous` must return a `Future` or a `CompletionStage` from the `java.util.concurrent` package. Otherwise, a `FaultToleranceDefinitionException` occurs. 
 
 When a method annotated with `@Asynchronous` is invoked, it immediately returns a `Future` or `CompletionStage`. The execution of the any remaining interceptors and the method body will then take place on a separate thread.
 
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/asynchronous.asciidoc
```diff
@@ -52,7 +52,15 @@ public CompletionStage<Connection> serviceA() {
...
+In this case, the method invocation and any fault tolerance processing will occur in a different thread. The returned `Future` or `CompletionStage` will be completed with the final result once all other Fault Tolerance processing has been completed. However, the two different return type have some differences, listed below.
+
+==== Return a Future
+If a `Future` completes exceptionally, the exception may not trigger other Fault Tolerance capabilities, even if specified. For instance, if a method returns a `Future` with `Asynchronous` and `Retry` specified, the `Future` that completes exceptionally will not trigger the method to be retried. 
+
+==== Return a CompletionStage 
+When a method returns a `CompletionStage`, if the `CompletionStage` returns exceptionally, the exception will trigger other Fault Tolerance capabilities, e.g. `Retry` if specified. Therefore, the usage of `CompletionStage` is more preferred.
+
+Therefore, it is highly encouraged to use `CompletionStage` instead of `Future` as a method with the return of `Future` behaves similar to synchronous calls.
```
Azquelt: I don't think this section is very clear. It needs to clearly show the differences between the behaviour when using `CompletionStage` and when using `Future`.

I've put together some alternate text, what do you think?

=== Interaction with other fault tolerance annotations

`@Asynchronous` can be used in conjunction with any of the other Fault Tolerance annotations, but the behavior is slightly different depending on whether the method returns a `Future` or a `CompletionStage`.

==== Interactions when returning a Future

If the method returns a `Future`, then the other fault tolerance policies are applied only around the method call itself. This means:
* If the method throws an exception then that exception may trigger other fault tolerance logic.
* If the method returns a `Future` then the method call will be considered to have succeeded. This is the case even if the `Future` that was returned is incomplete, was cancelled or would throw an `ExecutionException` if its `get()` method was called.

==== Interactions when returning a CompletionStage 

If the method returns a `CompletionStage`, then the other fault tolerance policies are applied only around both the method call, and the eventual result of the `CompletionStage`. This means:
* If the method throws an exception then that exception may trigger other fault tolerance logic.
* If the method returns a `CompletionStage` which has completed exceptionally, then that exception may trigger other fault tolerance logic.
* If the method returns a `CompletionStage` which is incomplete, but then later completes exceptionally, then that exception may trigger other fault tolerance logic at the time when the `CompletionStage` completes excpetionally.
* The method call is only considered to have completed successfully if the method returns a `CompletionStage` and that `CompletionStage` has completed successfully.

This behaviour makes it easy to apply fault tolerance logic around a `CompletionStage` which was returned from another component, for example applying `@Retry` and `@Timeout` to the result of a  JAX-RS client call.

Azquelt: On re-reading my suggestion, I don't like they way I've used the word "may".

I'm not sure how best to express the idea that any given call either succeeds, or fails with an exception and that other fault tolerance policies, depending on their configuration, may take some action depending on the exception with which the method failed.

brunobat: @Emily-Jiang I think we should use the text proposed by @Azquelt. It's has more detail.
I'm not really sure about:

"If the method returns a CompletionStage which is incomplete, but then later completes exceptionally, then that exception may trigger other fault tolerance logic at the time when the CompletionStage completes excpetionally."

It's not clear which "other fault tolerance logic" we are talking. We should mention, "the other fault tolerance logic in the method with the original `@Asynchronous` annotation", or something around those lines.

PR #383, opened 2019-01-03
Fix tests where ms had been changed to ns inconsistently

### LINKED ISSUE #373: TCK uses currentTimeMillis
A couple of TCK tests still use `System.currentTimeMillis()` for measuring the time a thread takes. It is recommended that `nanoTime()` is used instead as nanoTime is independent of the system clock and, for example, switching to daylight savings time will not affect the tests.

Azquelt:
I've used Duration where possible to avoid having the same mistake
happen again.

Also fix RetryClientWithDelay so that it records the delay between
subsequent executions correctly.

Fixes #373

I also came across an issue where we were measuring times incorrectly in `testAsyncClassLevelTimeout` so I've fixed that too.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/retry/clientserver/RetryClientWithDelay.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/retry/clientserver/RetryClientWithDelay.java
@@ -20,6 +20,7 @@
 package org.eclipse.microprofile.fault.tolerance.tck.retry.clientserver;
 
 import java.sql.Connection;
+import java.time.Duration;
 import java.util.HashSet;
 import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/retry/clientserver/RetryClientWithDelay.java
```diff
@@ -35,30 +36,35 @@
...
     //there should be at least 4 retries
     @Retry(delay = 400, maxDuration= 3200, jitter= 400, maxRetries = 10)
     public Connection serviceA() {
         return connectionService();
     }
+    
     //simulate a backend service
     private Connection connectionService() {
-       // the time delay between each invocation should be 0-800ms
-       if (timestampForConnectionService != 0) {
```
Azquelt: Note that in the old revision, this condition would never be met because `timestampForConnectionService` would only be set to a non-zero value within this if-block.

PR #387, opened 2019-01-08
Add test to validate proper execution when jitter = 0

doychin:
Incorrectly written code can lead to error during execution if jitter = 0 is not properly processed.

Signed-off-by: Doychin Bondzhev <doychin@dsoft-bg.com>

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/ZeroRetryJitterTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/ZeroRetryJitterTest.java
@@ -30,7 +30,6 @@ import org.jboss.shrinkwrap.api.spec.JavaArchive;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
 import org.testng.annotations.Test;
 
-import static org.junit.Assert.fail;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.assertEquals;
 
@@ -56,14 +55,8 @@ public class ZeroRetryJitterTest extends Arquillian {
      */
     @Test
     public void test() {
-        try {
-            zeroJitterClient.serviceA();
-            assertEquals("Incorrect number of retires", 3, zeroJitterClient.getRetries());
-            assertTrue("It took too much time for 3 retries", zeroJitterClient.getTotalRetryTime() < 3 * 200);
-        }
-        catch (Exception e) {
-            e.printStackTrace();
-            fail("Unexpected exception");
-        }
+        zeroJitterClient.serviceA();
+        assertEquals("Incorrect number of retries", 3, zeroJitterClient.getRetries());
+        assertTrue("It took too much time for 3 retries", zeroJitterClient.getTotalRetryTime() < 3 * 200);
     }
 }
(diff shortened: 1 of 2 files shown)


genie-microprofile:

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/ZeroRetryJitterTest.java
```diff
@@ -0,0 +1,66 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2017 Contributors to the Eclipse Foundation
```
Emily-Jiang: should be 2019

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/ZeroRetryJitterTest.java
```diff
@@ -0,0 +1,66 @@
...
+            .addClasses(RetryClientForZeroJitter.class)
+            .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml")
+            .as(JavaArchive.class);
+
+        return ShrinkWrap
+            .create(WebArchive.class, "ftZeroTestJitter.war")
+            .addAsLibrary(testJar);
+    }
+    /**
+     * Test that checks that jitter = 0 does not generate error during method call.
```
Emily-Jiang: It is better to refactor this test to verify the backend service was retries 3 times.

brunobat:

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/ZeroRetryJitterTest.java
```diff
@@ -62,6 +62,7 @@ public void test() {
             assertTrue("It took too much time for 3 retries", zeroJitterClient.getTotalRetryTime() < 3 * 200);
         }
         catch (Exception e) {
+            e.printStackTrace();
```
Azquelt: Don't print the stack trace like this.

Either don't catch the exception (and allow it to be thrown from the test method), or include it as the second argument to `fail()`. Both of these methods should ensure that the exception is included in the test failure report, without having to dig into the server logs to find the stack trace.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/retry/clientserver/RetryClientForZeroJitter.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private int retries = -1; // first call is normal call
+
+    @Retry(maxRetries = 3, jitter = 0)
+    public void serviceA() {
+        long currentTime = System.currentTimeMillis();
+        totalRetryTime += previousTime > 0 ? currentTime - previousTime : 0;
+        previousTime = currentTime;
+        if (++retries == 3) {
+            return;
```
Emily-Jiang: I don't quite get why you do this. Why cannot you just do increment retries?

doychin: I tought an extra check on how long it will take to run the whole retry test will be a bonus. But now I see that I also need to add another parameter to @Retry to grantee that max delay is always 200.

If you think that is not needed I will remove it and will leave only retry counter.

Emily-Jiang: ah. I see. We have other tests already to cover the max delay.

Emily-Jiang:
I will merge this one and run against our implementation and will put more fix if needed.

doychin:
I've tested it against safeguard 1.2.1-SNAPSHOT

PR #389, opened 2019-01-08
Add test to validate proper execution when jitter = 0 - backport to 1.1.x branch

doychin:
Incorrectly written code can lead to error during execution if jitter = 0 is not properly processed.

Signed-off-by: Doychin Bondzhev <doychin@dsoft-bg.com>

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/retry/clientserver/RetryClientForZeroJitter.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/retry/clientserver/RetryClientForZeroJitter.java
@@ -20,7 +20,6 @@
 package org.eclipse.microprofile.fault.tolerance.tck.retry.clientserver;
 
 import javax.enterprise.context.ApplicationScoped;
-import java.sql.Connection;
 
 import org.eclipse.microprofile.faulttolerance.Retry;
 
@@ -39,12 +38,14 @@ public class RetryClientForZeroJitter {
 
     private int retries = -1; // first call is normal call
 
-    @Retry(jitter = 0)
-    public Connection serviceA() {
+    @Retry(maxRetries = 3, jitter = 0)
+    public void serviceA() {
         long currentTime = System.currentTimeMillis();
         totalRetryTime += previousTime > 0 ? currentTime - previousTime : 0;
         previousTime = currentTime;
-        retries++;
+        if (++retries == 3) {
+            return;
+        }
         throw new RuntimeException();
     }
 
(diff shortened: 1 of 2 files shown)


genie-microprofile:

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/retry/clientserver/RetryClientForZeroJitter.java
```diff
@@ -0,0 +1,58 @@
...
+@ApplicationScoped
+public class RetryClientForZeroJitter {
+
+    private long totalRetryTime = 0;
+
+    private long previousTime = 0;
+
+    private int retries = -1; // first call is normal call
+
+    @Retry(jitter = 0)
```
brunobat: Hi @doychin can you please set the method to retry at least 3 times?
The 2 first retries should fail and the second should pass.
You can push the changes to this branch and the PR will be automatically updated, no need to open a new PR. 
Thanks!

brunobat:

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/retry/clientserver/RetryClientForZeroJitter.java
doychin: isn't maxRetries by default 3? Anyway I will add it as parameter to the annotation. So I add check and on the third retry it will just return without exception. The test code will change to consider exception as fail and when there is no exception it will check for number of retries and total time to execute.

brunobat: yes, the default is 3, but the test should be explicit and resilient to defaults change.

Azquelt:
I don't think we add new TCK tests to released versions, we only remove or fix broken tests. @Emily-Jiang ?

doychin:
Basically this test does not check anything new that is not checked by other tests except when jitter is 0. The whole reason for that is I found a test in one app server that tests something related to @Retry and provides jitter = 0 as parameter. In that case the FT library crashes because jitter was provided as parameter to random generator and 0 is invalid value.

Emily-Jiang:
@Azquelt was right. @doychin we are not going to add any new TCKs to 1.x branch even though the tests were for previous behavior as we don't want to break any compliance status by implementors in the service stream.

Emily-Jiang:
Please address the comments made by @brunobat on your other PR. I will close this one for now.

doychin:
I already made the requested modifications

Issue #399, opened 2019-01-16
Make timeouts in tests configurable


Azquelt:
Some of the tests unavoidably require artificial waits and checking the time it takes for something to happen.

Cases which require these time-sensitive checks:
* Timeout (check that the timeout is fired at the correct time)
* Retry (check that delay and jitter are applied correctly)
* Metrics (check that execution time and queue wait time metrics are applied correctly)
* CircuitBreaker (check that an open breaker half-closes after the configured time)

In addition, some tests require a timeout to avoid hanging if an implementation is not correct. For example:
* waiting for an asynchronous task to finish

We should make these times configurable so that the tests are flexible and can run in environments where test reliability is important but where performance is poor or unpredictable. (e.g. on a busy CI system)

For the tests which expect an action to take a specific time, I would suggest the following:
* a standard time should be configurable via a system property
* tests should specify any delay or timeout as a multiple of this property
  * E.g. a test to check that a TimeoutException is thrown when a method takes longer than the configured timeout may specify the timeout as one time unit, and have the method body sleep for two time units.
* tests should check that the actual value is within a certain percentage of the expected value, to ensure that checks remain valid even if a very large timeout is configured
  * E.g. a test which checks that a TimeoutException is thrown after the configured duration may check that it's thrown after the configured time +-10%
* where a timeout value is specified in an annotation, we should dynamically generate a config file to change the value
  * to ensure that implementations function correctly without any configuration, we may need additional tests with generous timeouts which cannot be configured

Any tests which are not specifically testing one of the areas in the spec noted above should not use any time measurement and should not rely on sleeps to ensure correctness (this may also improve the performance of the tests).

spericas:
We run into this issue pretty frequently in our pipelines. In 2.0, there are also some tests that have some pretty tight constraints, e.g. `ZeroRetryJitterTest::test` and `Retry::testRetryWithNoDelayAndJitter`. The latter also adds a somewhat arbitrary 100 ms when checking acceptable delay ranges. Making some of these delays configurable would be great.

Azquelt:
Changes to allow timeouts in tests to be configured were made under #458 and #476.

I've also opened #477 to backport those changes to the 2.0.x branch.

Azquelt:
#477 is now merged as well.

PR #400, opened 2019-01-17
Widen the timing bounds on TimeoutUninterruptableTest

Azquelt:
Unnecessarily tight time bounds can cause spurious failures in case of
unexpected pauses (e.g. CG, jit, CPU time starvation)

TimeoutUninterruptableTest had quite tight bounds on how quickly the
TimeoutException is returned. However, what's important in this test is
that the TimeoutException is returned before the method has returned so
the bounds can be loosened without invalidating the test.

This sort of problem should be resolved properly by making the timeouts configurable under #399, this is a stop-gap measure to reduce the chance of spurious failures when running the 2.0 TCK.

Also change calls which would wait forever so that the test doesn't hang if the implementation does not work correctly.


brunobat:
3 ok. Thanks @Azquelt !

PR #403, opened 2019-02-06
Make timeouts in tests configurable - Part I

brunobat:

This is related to issue #399 

First approach to the configurable timeouts problem. 
Missing: 
- Not all classes were changed
- Make the configuration timeout more fine grained by using regular expressions. 
- Fix checkstyle issues
- explore to use a TestNG runner to override the `@Test` timeout attributes

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -127,26 +127,6 @@ public class FaultToleranceBean {
 }
 }
 ```
-##### Timeout Configs
-Since v2.1 we can configure the existing test timeouts. With the following configurations, timeouts can be greatly 
-reduced for local usage or in fast CI environments. 
-
-The properties can be set either with MP Config or System Properties. 
-
-.TCK Timeout configs
-|===
-|Property |Default |unit
-
-|org.eclipse.microprofile.fault.tolerance.basetimeout
-|100
-|ms
-
-|org.eclipse.microprofile.fault.tolerance.basemultiplier
-|10
-|Dimensionless
-|===
-
-
 ## Impact on existing code
 
 n/a
--- a/README.adoc
+++ b/README.adoc
@@ -127,6 +127,26 @@ public class FaultToleranceBean {
 }
 }
 ```
+##### Timeout Configs
+Since v2.1 we can configure the existing test timeouts. With the following configurations, timeouts can be greatly 
+reduced for local usage or in fast CI environments. 
+
+The properties can be set either with MP Config or System Properties. 
+
+.TCK Timeout configs
+|===
+|Property |Default |unit
+
+|org.eclipse.microprofile.fault.tolerance.basetimeout
+|100
+|ms
+
+|org.eclipse.microprofile.fault.tolerance.basemultiplier
+|10
+|Dimensionless
+|===
+
+
 ## Impact on existing code
 
 n/a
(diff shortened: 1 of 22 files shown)


README.adoc
```diff
@@ -127,6 +127,26 @@ public class FaultToleranceBean {
...
+
+The properties can be set either with MP Config or System Properties. 
+
+.TCK Timeout configs
+|===
+|Property |Default |unit
+
+|org.eclipse.microprofile.fault.tolerance.basetimeout
+|100
+|ms
```
Emily-Jiang: We discussed this on today's call and decided not to introduce this property as the multiplier is efficient.

README.adoc
```diff
@@ -127,6 +127,26 @@ public class FaultToleranceBean {
...
+
+|org.eclipse.microprofile.fault.tolerance.basemultiplier
+|10
```
Emily-Jiang: On today's call, we agreed to change this to a double value.

README.adoc
Emily-Jiang: Add instruction to running_the_tck doc.

brunobat: done

README.adoc
brunobat: done

brunobat:
Will start on a new PR for the remaining test changes and for getting the annotation attributes using reflection.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/util/MetricComparator.java
```diff
@@ -35,10 +36,12 @@ private MetricComparator() {}
...
      * Useful for checking the results from Histograms.
      * 
-     * @param millis the expected time in milliseconds
+     * @param originalMillis the expected time in milliseconds
      * @return a {@link Matcher} which matches against a time in milliseconds
      */
-    public static Matcher<Double> approxMillis(long millis) {
+    public static Matcher<Double> approxMillis(final long originalMillis) {
+
+        final long millis = TCKConfig.getConfig().getTimeoutInMillis(originalMillis);
```
Azquelt: We should also scale the margin here as well.

brunobat: Hi @Azquelt, I don't know if I follow. What do you mean by the margin?

Azquelt: Sorry, my comment was really unclear :(

`closeTo` takes two parameters, the expected value and the allowed margin of error. I think we should be scaling both values rather than just one.

Possibly we should even change this method so that rather than having a fixed margin of error of +/- 100ms, it uses something like +/- 20% of the expected value.

brunobat: Thanks Andrew. I'm going to do the change

brunobat:
... Need to file the new paperwork

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/util/MetricComparator.java
brunobat: done

brunobat:
Closing PR to address license issues

PR #412, opened 2019-02-19
Async Test Refactor using Awaitility

carlosdlr:


tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/awaitility/poc/AsyncClientTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+public class AsyncClientTest  extends Arquillian {
+
+    @Inject
+    private AsyncClient client;
+
+    @Deployment
+    public static WebArchive deploy() {
+        File[] libs = Maven.resolver().loadPomFromFile("pom.xml")
+            .resolve("org.awaitility:awaitility", "org.assertj:assertj-core").withTransitivity()
```
Ladicek: You could create an Arquillian extension that would add Awaitility to all tested archives, similarly to e.g. this: https://github.com/rhoar-qe/thorntail-test-suite/blob/master/common/src/main/java/org/wildfly/swarm/ts/common/AssertjArquillianExtension.java

carlosdlr: Thanks, @Ladicek for share this info, ok I will include the extension with the refactoring that I'm currently working on.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncClient.java
```diff
@@ -44,92 +40,21 @@
...
-    public CompletionStage<Connection> serviceCS(Future<?> waitCondition, boolean throwException) {
-        return serviceCS(waitCondition, false, null);
-    }
-
-    @Asynchronous
-    public CompletionStage<Connection> serviceCS(Future<?> waitCondition) {
-        return serviceCS(waitCondition, false);
+    public CompletionStage<Task> serviceCS(Task task) {
+        new AsyncBridge(task).perform(1000,"service DATA");
+        return CompletableFuture.completedFuture(task);
```
Azquelt: Similar the other async test client, the previous code was preferable. Avoid sleeps where possible.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/common/AsyncBridge.java
```diff
@@ -0,0 +1,56 @@
...
+        this.task = task;
+    }
+
+    /**
+     * method that performs or executes the task
+     * @param taskTime time that will take the task to be executed
+     * @param expectedTaskResult the expected task result string after the process
+     */
+    public void perform(long taskTime, String expectedTaskResult) {
+        Thread thread = new Thread(new Runnable() {
```
Azquelt: You must not create new `Thread`s like this if the test may run on a Java EE server.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/awaitility/poc/AsyncClientTest.java
carlosdlr: The extension was added and is working now with the asynchronous tests. Thank you

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousCSTest.java
```diff
@@ -144,205 +121,60 @@ public void testClassLevelAsyncIsNotFinished() {
...
-        try {
-            waitUntilCompleted(resultFuture);
-        }
-        catch (CompletionException e) {
-            handleCompletionException(e);
-        }
+        Task taskToPerform = new ServiceTask();
+        CompletionStage<Task> taskResult = clientClass.serviceCS(taskToPerform);
+        await().untilAsserted(() -> Assertions.
+            assertThat(taskResult.toCompletableFuture().isDone()).isTrue());
```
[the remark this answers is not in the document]
carlosdlr: Thanks, @Azquelt for this catch is not necessary, awaitility works well with other assert APIs like testng assertions, the dependency was removed and the tests were refactored accordingly now is working with testng asserts instead.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncClassLevelClient.java
```diff
@@ -45,70 +42,18 @@
...
-        return CompletableFuture.completedFuture(new Connection() {
-            @Override
-            public String getData() {
-                return "service DATA";
-            }
-        });
-
+    public CompletionStage<Task> serviceCS(Task task) {
+        new AsyncBridge(task).perform(1000,"service DATA");
+        return CompletableFuture.completedFuture(task);
```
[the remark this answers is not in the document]
carlosdlr: done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/common/AsyncBridge.java
carlosdlr: done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncClient.java
carlosdlr: done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncClassLevelClient.java
```diff
@@ -45,70 +42,18 @@
...
-            }
-
-            @Override
-            public String getData() {
-                return "service DATA";
-            }
-        };
-
-        return CompletableFuture.completedFuture(conn);
+    public Future<Task> service(Task task) {
```
brunobat: There is a parameter missing on the Javadoc

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncClassLevelClient.java
```diff
@@ -45,70 +42,18 @@
...
-        }
-
-        return CompletableFuture.completedFuture(new Connection() {
-            @Override
-            public String getData() {
-                return "service DATA";
-            }
-        });
-
+    public CompletionStage<Task> serviceCS(Task task) {
```
brunobat: Parameter also missing on the Javadoc

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncClient.java
```diff
@@ -44,92 +40,21 @@
...
-            }
-
-            @Override
-            public String getData() {
-                return "service DATA";
-            }
-        };
-
-        return CompletableFuture.completedFuture(conn);
+    public Future<Task> service(Task task) {
```
brunobat: Javadoc parameter missing

tck/pom.xml
```diff
@@ -109,11 +109,27 @@
...
+        <dependency>
+            <groupId>org.jboss.shrinkwrap.resolver</groupId>
+            <artifactId>shrinkwrap-resolver-api-maven</artifactId>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.shrinkwrap.resolver</groupId>
+            <artifactId>shrinkwrap-resolver-impl-maven</artifactId>
+            <scope>compile</scope>
+        </dependency>
```
Azquelt: Are these dependencies still needed?

I think they were only used for the way you were previously adding awaitility to the application.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousCSTest.java
```diff
@@ -78,20 +79,9 @@ public static WebArchive deploy() {
...
-
-        try {
-            waitUntilCompleted(resultFuture);
-        }
-        catch (CompletionException e) {
-            handleCompletionException(e);
-        }
+        Task taskToPerform = new ServiceTask();
+        CompletionStage<Task> taskResult = client.serviceCS(taskToPerform);
+        assertFalse(taskResult.toCompletableFuture().isDone());
```
Azquelt: This `assertFalse(taskResult.isDone())` check is not sufficient to prove correctness.

`serviceCS` is an asynchronous method, so `isDone()` could always be false, even if the implementation is not correct.

The old check ensures that `isDone()` continues to return `false` for 500ms to give the task the opportunity to complete if it's going to.

Azquelt: Annoyingly, I can't find a way to have awaitility do this sort of check

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousCSTest.java
```diff
@@ -78,20 +79,9 @@ public static WebArchive deploy() {
...
-        CompletableFuture<Void> waitingFuture = newWaitingFuture();
-        
-        CompletionStage<Connection> resultFuture = client.serviceCS(waitingFuture);
-
-        Assert.assertFalse(completesQuickly(resultFuture));
-
-        complete(waitingFuture);
-
-        try {
-            waitUntilCompleted(resultFuture);
```
Azquelt: It would be good to keep this cleanup logic (though it could possibly be put into a common `@AfterMethod` method) as otherwise you may be leaving a thread waiting forever.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousCSTest.java
```diff
@@ -144,205 +112,57 @@ public void testClassLevelAsyncIsNotFinished() {
...
+        CompletionStage<Task> taskResult = client.serviceCS(taskToPerform).thenApply(v -> {
+            executionRecord.append("1");
+            return v;
+        }).thenApply(v -> {
+            executionRecord.append("2");
+            return v;
+        }).thenApply(v -> {
+            executionRecord.append("3");
+            return v;
+        });
```
Azquelt: This test is not the same as the original test.

The original test has the first thenApply applied to a future which is completed within the asynchronous method, the result of which is then returned.

Please revert this change.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/common/AsyncBridge.java
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * method that performs or executes the task
+     * @param taskTime time that will take the task to be executed
+     * @param expectedTaskResult the expected task result string after the process
+     */
+    public void perform(long taskTime, String expectedTaskResult) {
+        CompletableFuture<Task> waitingFuture = new CompletableFuture<>();
+        try {
+            waitingFuture.get(taskTime, TimeUnit.MILLISECONDS);
```
Azquelt: I'm not sure what the intention of this is.

As waitingFuture is created just above, this will always time out.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/common/AsyncBridge.java
```diff
@@ -0,0 +1,59 @@
...
+            task.doTask(expectedTaskResult);
+            waitingFuture.completedFuture(task);
```
Azquelt: Not sure what this does either. `completedFuture` is a static method which creates a new `CompletableFuture`.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousCSTest.java
```diff
@@ -144,205 +112,57 @@ public void testClassLevelAsyncIsNotFinished() {
...
-        catch (ExecutionException ex) {
-            Assert.assertTrue(causeClazz.isInstance(ex.getCause()), "Cause of ExecutionException was " + ex.getCause());
-        }
-        catch (Throwable ex) {
-            Assert.fail("Unexpected exception thrown", ex);
-        }
-    }
+            with().pollThread(Thread::new).await()
+                .atMost(120, TimeUnit.MILLISECONDS).until(
+                () -> taskToPerform.getTaskResult() == "service DATA");
```
Azquelt: Use `.equals()` rather than `==` to compare strings

`"service DATA".equals(taskToPerform.getTaskResult())` is probably the safest option here to avoid an exception if `getTaskResult()` returns `null`.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousCSTest.java
```diff
@@ -144,205 +112,57 @@ public void testClassLevelAsyncIsNotFinished() {
...
 
-    private static class SimulatedException extends RuntimeException {
+            assertTrue(taskResult.toCompletableFuture().isCompletedExceptionally());
 
-        public SimulatedException() {
+            fail("Should throw ConditionTimeoutException");
```
Azquelt: If we're expecting a `ConditionTimeoutException` why is there an assert on the previous line?

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousCSTest.java
```diff
@@ -144,205 +112,57 @@ public void testClassLevelAsyncIsNotFinished() {
...
-        assertThrowsExecutionExceptionWithCause(SimulatedException.class, CompletableFutureHelper.toCompletableFuture(resultFuture)::get);
-    }
-
-    /**
-     * Ensure that any waiting futures get completed at the end of each test
-     * <p>
-     * Important in case tests end early due to an exception or failure.
-     */
-    @AfterTest
-    public void completeWaitingFutures() {
```
Azquelt: This method was important to ensure that any waiting futures were cleaned up at the end of the test.

Ideally we should be using waitingFutures (or some other sort of synchronization) instead of sleeps as this allows test tasks to run only as long as their needed, rather than for a fixed length of time. In this case, I would expect to see some cleanup logic similar to this method to ensure that in the case of failure, we don't leave any threads waiting indefinitely for a signal from the test which will never arrive.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/common/AsyncBridge.java
Azquelt: Given the rest of the code, this means that `task.doTask()` will never be called which I expect will cause test failures.

tck/running_the_tck.asciidoc
```diff
@@ -44,6 +44,11 @@ To enable the tests in your project you need to add the following dependency to
     <version>6.9.9</version>
     <scope>test</scope>
 </dependency>
+<dependency>
+    <groupId>org.awaitility</groupId>
+    <artifactId>awaitility</artifactId>
+    <version>3.1.6</version>
+</dependency>
```
Azquelt: Not sure this is needed. Awaitility should be pulled in as a transitive dependency of the TCK, so the user shouldn't have to declare it themselves (unless I'm missing something here?)

carlosdlr:
Hello, @Azquelt the purpose of the Task or the AsyncBridge classes is to get a centralized component to handle all the asynchronous calls or process that you can pass the time for the task and the string expected as a result and the task interface is intending to have a common behavior for another task that can be used for other tests.  

I think is better discard this pull request and review in more detail this in a new one due to all the findings looks that i misunderstand the tests approach.

Thank you

Azquelt:
Ok, makes sense. Thanks @carlosdlr

PR #421, opened 2019-03-28
clarify circuit breaker metrics

### LINKED ISSUE #419: need clarification on CircuitBreaker.failOn + metrics
If I have a `@CircuitBreaker(failOn = {IllegalArgumentException.class})` method and e.g. `IllegalStateException` is thrown, which circuit breaker metric should be incremented?

The MP Fault Tolerance spec is a little vague in this regard:

1. `circuitbreaker.callsSucceeded.total` is defined as "Number of calls allowed to run by the circuit breaker that returned successfully"
1. `circuitbreaker.callsFailed.total` is defined as "Number of calls allowed to run by the circuit breaker that then failed"
1. the `@CircuitBreaker` javadoc says "A method call will be considered a failure if it throws an exception and the type of that exception is assignable to any of the types listed in `failOn`."

From item 1 follows that in case of any exception, it should be the `circuitbreaker.callsFailed.total` metric that is incremented, because throwing an exception isn't a "successful return". However, from items 2+3 follows that it should be `circuitbreaker.callsSucceeded.total`.

The TCK doesn't have a test for this.

My opinion is that the 2nd interpretation is correct, because that's how everything else in the circuit breaker works. WDYT?

Ladicek:
Resolves #419

In case `@CircuitBreaker(failOn = {IllegalArgumentException.class})`
is defined and the guarded method throws e.g. `IllegalStateException`,
the circuit breaker treats it as success and therefore the
`circuitbreaker.callsSucceeded.total` metric needs to be incremented.

This commit adjusts the spec to make this clear, and also adds
a TCK test for this behavior.

### DIFF
--- a/spec/src/main/asciidoc/metrics.asciidoc
+++ b/spec/src/main/asciidoc/metrics.asciidoc
@@ -102,15 +102,15 @@ Note that the sum of `ft.<name>.timeout.callsTimedOut.total` and `ft.<name>.time
 
 |`ft.<name>.circuitbreaker.callsSucceeded.total`
 | Counter | None
-| Number of calls allowed to run by the circuit breaker that returned successfully
+| Number of calls which were allowed to run and were considered a success by the circuit breaker
 
 |`ft.<name>.circuitbreaker.callsFailed.total`
 | Counter | None
-| Number of calls allowed to run by the circuit breaker that then failed
+| Number of calls which were allowed to run and were considered a failure by the circuit breaker
 
 |`ft.<name>.circuitbreaker.callsPrevented.total`
 | Counter | None
-| Number of calls prevented from running by an open circuit breaker
+| Number of calls which were prevented from running by the circuit breaker
 
 |`ft.<name>.circuitbreaker.open.total`
 | Gauge<Long> | Nanoseconds
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/CircuitBreakerMetricTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/CircuitBreakerMetricTest.java
@@ -18,6 +18,7 @@
  */
 package org.eclipse.microprofile.fault.tolerance.tck.metrics;
 
+import static org.eclipse.microprofile.fault.tolerance.tck.util.Exceptions.expect;
 import static org.eclipse.microprofile.fault.tolerance.tck.util.Exceptions.expectCbOpen;
 import static org.eclipse.microprofile.fault.tolerance.tck.util.Exceptions.expectTestException;
 import static org.hamcrest.MatcherAssert.assertThat;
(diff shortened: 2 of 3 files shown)


genie-microprofile:

Azquelt:

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/CircuitBreakerMetricTest.java
```diff
@@ -122,9 +123,17 @@ public void testCircuitBreakerMetric() throws Exception {
...
+
+        assertThat("circuitbreaker calls succeeded", m.getCircuitBreakerCallsSucceededDelta(), is(3L));
+        assertThat("circuitbreaker calls failed", m.getCircuitBreakerCallsFailedDelta(), is(2L));
+        assertThat("circuitbreaker calls prevented", m.getCircuitBreakerCallsPreventedDelta(), is(1L));
+        assertThat("circuit breaker times opened", m.getCircuitBreakerOpenedDelta(), is(1L));
+
         // General metrics should be updated
-        assertThat("invocations", m.getInvocationsDelta(), is(5L));
+        assertThat("invocations", m.getInvocationsDelta(), is(6L));
         assertThat("failed invocations", m.getInvocationsFailedDelta(), is(3L));
```
Azquelt: I'd want to run it to check, but I think this counter should also be incremented.

This is the counter for `ft.<name>.invocations.failed.total`, defined as "The number of times the method was called and, after all Fault Tolerance actions had been processed, threw a Throwable"

Ladicek: Ah right! Didn't notice that, I was still thinking in terms of what circuit breaker considers failure. Will fix.

Ladicek: Fixed.

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -102,11 +102,11 @@ Note that the sum of `ft.<name>.timeout.callsTimedOut.total` and `ft.<name>.time
 
 |`ft.<name>.circuitbreaker.callsSucceeded.total`
 | Counter | None
-| Number of calls allowed to run by the circuit breaker that returned successfully
+| Number of calls allowed to run by the circuit breaker that were considered a success
```
Ladicek: I was thinking if this should read "that were considered a success _by the circuit breaker_" (dtto for the failure case below), but I thought that's perhaps too repetitive. WDYT?

Azquelt: Hmm, I see what you mean but it is a bit repetitive. How about:
* Number of calls which were allowed to run and were considered a success by the circuit breaker
* Number of calls which were allowed to run and were considered a failure by the circuit breaker
* Number of calls which were prevented from running by the circuit breaker _(slight change to the current wording)_

Azquelt:

spec/src/main/asciidoc/metrics.asciidoc
Ladicek: That's nice, thanks!

Azquelt:

PR #422, opened 2019-04-01
Enhanced circuit breaker API docs

sdaschner:
The JavaDoc of `@CircuitBreaker` is IMO lacking a bit clarity. I saw some confusion around the delay period when exactly that is taken into consideration. I've added the documentation of the circuit breaker states, clarified, and cleaned-up the individual annotation method docs.


genie-microprofile:

Azquelt:
Forgot to run a build before merging this and it's introduced a checkstyle error which I'm fixing under #428

PR #426, opened 2019-04-12
Assume transparent Future implementation

jbee:
In the spec section on _Asynchronous Usage_ it is stated that the returned `Future` or `CompletionStage` should _delegate_ to the result obtained asynchronously from the annotated method. 

I think this is a overly specific demand on the actual implementation that defies the purpose of abstraction.  The caller should only care about semantically correct behaviour. 

In this particular tests it was assumed that `Future.get()` and `Future.get(long, TimeUnit)` have different results. While it is possible to implement a `Future` like this it is semantically incorrect to do so and an implementation should not be limited by such an assumption. In the same way we need to make sure that `a.equals(b)` is consistent with `b.equals(a)` it should also be true that `future.get(Long.MAX_VALUE, TimeUnit.DAYS)` is practically the same as and consistent with `future.get()`. Given that being true there are other ways than delegation that provide the caller with a `Future` that makes the value available as soon as it is computed. Which way is chosen should be transparent.

The suggested change to the spec is only provided as a basis for improvement and for lack of a better phrasing. I think the overall passage on _execution_ is unclear to the extend that it does not make clear if execution or computation means computation of the methods result or the value of the results future.
Some sentences only make sense when applied to the lazily computed value, others seem to refer to the `Future` wrapper as a vehicle of computation. 

Instead of asserting a particular result the tests now only assert a non-null result which is plenty good for the purpose of the test as indicated by their names.

In the PR I also changed the catch blocks to not catch `Throwable` as this would also catch the `AssertionError` thrown within the try block what makes the provided error message less helpful than it could be.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadFutureTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadFutureTest.java
@@ -19,6 +19,11 @@
  *******************************************************************************/
 package org.eclipse.microprofile.fault.tolerance.tck.bulkhead;
 
+import static org.junit.Assert.assertNotNull;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertTrue;
+import static org.testng.Assert.fail;
+
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 
@@ -50,7 +55,6 @@ import org.testng.annotations.Test;
 public class BulkheadFutureTest extends Arquillian {
 
     private static final int SHORT_TIME = 100;
-    private static final int LONG_TIME = 3000;
     @Inject
     private BulkheadMethodAsynchronousDefaultBean bhBeanMethodAsynchronousDefault;
     @Inject
@@ -87,19 +91,18 @@ public class BulkheadFutureTest extends Arquillian {
             result = bhBeanMethodAsynchronousDefault.test(fc);
         }
         catch (InterruptedException e1) {
-            Assert.fail("Unexpected interruption", e1);
+            fail("Unexpected interruption", e1);
         }
 
-        Assert.assertFalse(result.isDone(), "Future reporting Done when not");
+        assertFalse(result.isDone(), "Future reporting Done when not");
         try {
-            String r;
-            Assert.assertTrue("GET.".equals(r = result.get()), r);
-            Assert.assertTrue("GET.GET_TO.".equals(r = result.get(1, TimeUnit.SECONDS)), r);
+            assertNotNull(result.get());
+            assertNotNull(result.get(1, TimeUnit.SECONDS));
         }
-        catch (Throwable t) {
-            Assert.assertNull(t);
+        catch (Exception e) {
+            throw new AssertionError("Unexpected exception: ", e);
         }
-        Assert.assertTrue(result.isDone(), "Future done not reporting true");
+        assertTrue(result.isDone(), "Future done not reporting true");
     }
 
     /**
@@ -116,10 +119,10 @@ public class BulkheadFutureTest extends Arquillian {
             result = bhBeanMethodAsynchronousDefault.test(fc);
         }
         catch (InterruptedException e1) {
-            Assert.fail("Unexpected interruption", e1);
+            fail("Unexpected interruption", e1);
         }
 
-        Assert.assertFalse(result.isDone(), "Future reporting Done when not");
+        assertFalse(result.isDone(), "Future reporting Done when not");
         try {
             Thread.sleep(SHORT_TIME + SHORT_TIME);
             // Usually, we will not enter the loop, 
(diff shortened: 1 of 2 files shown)


genie-microprofile:

Ladicek:
Occurences of `throw new AssertionError()` could be replaced by `Assert.fail()` for consistency, but that's just cosmetics.

Not sure if asserting that `Future.get` just returns non-null value; I think it is valuable to assert that the claim "providing the same value". Not sure how easy or hard would it be with current TCK structure.

jbee:
If I am not mistaken there is no variant of `fail` that also takes a cause exception. Therefore I did the explicit  `new AssertionError`.

I think for the test scenario the actual value returned by `get` is irrelevant. What the tests want to verify is that the `isDone()` state is reflected properly. Therefore it figured it is good enough to just `assertNotNull`.

Azquelt:
Had a further thought. If we're removing the language that requires delegating to the returned object, then we need something extra to say what happens if the execution has completed and the user calls `cancel()`.

I would expect this would result in `cancel()` being called on the future that was returned by the method body.

jbee:
You mean something like _"calling `cancel()` on a `Future` returned to the caller should cancel the `Future` returned by the annotated method"_?

Azquelt:
Yes basically, but only in the case where the method has completed. We should probably also define what happens if the user cancels the future, but the method finishes anyway (should they also cancel the returned future)

jbee:
With `@Asynchronous`, `@Retry` and `@Timeout` this becomes especially interesting since there could be multiple concurrent attempts running trying to get a result in case some time out which all should be cancelled I guess in case the future returned to the called is cancelled by the caller.

jbee:
I did sign the agreement for my work email and added it as one of the email connected to my GH account. Hope this works.

Azquelt:
Not sure. When I click through to the details, I see this:
> Anonymous (jaanbernitt@****.com) is not covered by the necessary legal agreements in order to proceed.

Based on that, I guess it's basing it on the author e-mail address on your commit. If that's the case, I think you'd either need to recreate the commit with your work e-mail or sign the ECA under your personal e-mail.

jbee:
Yeah, I was suspecting that. Didn't think about it enough when I signed the agreement. Will have to resign for my other email.

Emily-Jiang:
@jbee will you be able to fix the sign off email soon so that this PR is in the position to be merged?

jbee:
@Emily-Jiang I thought I did that month ago but it apparently does not work.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadFutureTest.java
```diff
@@ -87,19 +91,18 @@ public void testBulkheadMethodAsynchFutureDoneAfterGet() {
...
+            fail("Unexpected interruption", e1);
         }
 
-        Assert.assertFalse(result.isDone(), "Future reporting Done when not");
+        assertFalse(result.isDone(), "Future reporting Done when not");
         try {
-            String r;
-            Assert.assertTrue("GET.".equals(r = result.get()), r);
-            Assert.assertTrue("GET.GET_TO.".equals(r = result.get(1, TimeUnit.SECONDS)), r);
+            assertNotNull(result.get());
```
Emily-Jiang: Please revert this line to make sure the result is the same.

jbee: This is the crux of this PR so if you want it reverted we can just close it.

Ladicek: Could you please explain more? The future must still be completed with the same value, no?

jbee: The asserts only make sense and hold true if delegation is used to implement the feature. Other ways of implementation will not show the behaviour expected by the asserts because the underlying solution does not use delegation to link to the asynchronously completing computation which is correct to uphold the semantics of `Future` but is not sufficient for the expectations of the test. The test makes an assumption that is beyond `Future` semantics forcing a particular solution. See also the PR description.

Ladicek: So what does `Future.get()` return? `null`? Arbitrary `Object`? Anything it wants?

Azquelt: Rather than changing the test to not check the result, would it make more sense to change `FutureChecker` so that it returns consistent results.

Currently it changes its result every time you call either `get()` method, which doesn't fulfil the contract of the `Future` interface.

jbee: Whatever it returns must be consistent for `get()` and get with arguments. They cannot be different because that breaks abstraction.

Ladicek: Ah, I just noticed the expected difference. That's indeed wrong. But I still think that the TCK shoudl test that the `Future` returns correct result, not just that it returns "something".

jbee:
@Emily-Jiang My suggestion was to weaken the assert since it is not essential to the verification of the particular case. I can do this but if I read the comments correctly this is not what is asked for. So either let me know what exactly the assert should verify and how to change the setup. I am also perfectly fine with someone else doing those changes as it might be simpler to do them then to explain how this should end up.

Emily-Jiang:
I'm closing this one as #471 handles this. @jbee please comment to see whether you are happy with the suggestion there.

PR #429, opened 2019-05-16
add TCK test for the @Asynchronous @Fallback combination

Ladicek:

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsyncFallbackTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsyncFallbackTest.java
@@ -1,0 +1,101 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


genie-microprofile:

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsyncFallbackTest.java
```diff
@@ -0,0 +1,101 @@
...
+        catch (ExecutionException expected) {
+            assertThat("Future-returning method that returns failing future should not fallback",
+                expected.getCause(), instanceOf(IOException.class));
+        }
+    }
+
+    @Test
+    public void testAsyncCSFallbackSuccess() throws ExecutionException, InterruptedException {
+        assertThat("CompletionStage-returning method that returns successfully should not fallback",
+            client.serviceCS1().toCompletableFuture().get(), equalTo("Success"));
```
Azquelt: `CompletionStage.toCompletableFuture()` is technically an optional method. Could you use the static method on  `CompletableFutureHelper` instead?

Azquelt:

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsyncFallbackTest.java
Ladicek: I've seen the helper method and I thought it's superfluous, because I believed that all `CompletionStage`s support it. I read the javadoc again and you're right, it says "A CompletionStage implementation that does not choose to interoperate with others may throw UnsupportedOperationException." I'll fix this.

Ladicek: Fixed.

Azquelt:

Issue #73, opened 2017-07-17
Consider to use awaitibility 


Emily-Jiang:
John brought up our tck might benefit from [awaitibility](https://github.com/awaitility/awaitility).  Use this one to prototype and then refactor the current tck.

derekm:
I wrote some simple wait loops that I use... I think they have several benefits over awaitility.

They look a little awkward in usage, but they work & they're tested unlike most adhoc wait loops:

```
package core.helper.waitloop;

import java.util.function.Predicate;

import core.helper.lambdathrows.ConsumerThrows;
import core.helper.lambdathrows.FunctionThrows;

public class WaitLoopWhile<T> implements ConsumerThrows<T> {

	private FunctionThrows<T,T> whileBody;
	private Predicate<T> whileTest;
	private int count = 10;
	private int counter = 0;

	public WaitLoopWhile(Predicate<T> whileTest, FunctionThrows<T,T> whileBody) {
		this.whileTest = whileTest;
		this.whileBody = whileBody;
	}

	public WaitLoopWhile(int count, Predicate<T> whileTest, FunctionThrows<T,T> whileBody) {
		this(whileTest, whileBody);
		this.count = count;
	}

	@Override
	public void accept(T t) throws Throwable {
		while (whileTest.test(t) && ++counter <= count) {
			t = whileBody.apply(t);
		}
		if (whileTest.test(t) && counter > count) {
			throw new WaitLoopExceededException("Quit loop after " + count + " iterations.");
		}
	}
}
```
...and...
```
package core.helper.waitloop;

import java.util.function.Predicate;

import core.helper.lambdathrows.RunnableThrows;
import core.helper.lambdathrows.SupplierThrows;

public class WaitLoopDoWhile<T> implements RunnableThrows {

	private SupplierThrows<T> doBody;
	private Predicate<T> whileTest;
	private int count = 10;
	private int counter = 0;

	public WaitLoopDoWhile(SupplierThrows<T> doBody, Predicate<T> whileTest) {
		this.doBody = doBody;
		this.whileTest = whileTest;
	}

	public WaitLoopDoWhile(int count, SupplierThrows<T> doBody, Predicate<T> whileTest) {
		this(doBody, whileTest);
		this.count = count;
	}

	@Override
	public void run() throws Throwable {
		T t;
		do {
			t = doBody.get();
		} while (whileTest.test(t) && ++counter <= count);
		if (whileTest.test(t) && counter > count) {
[...]

PR #431, opened 2019-06-04
Adding Awaitility to MP TCK

carlosdlr:
Issue https://github.com/eclipse/microprofile-fault-tolerance/issues/73

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousTest.java
@@ -81,7 +81,8 @@ public class AsynchronousTest extends Arquillian {
     public void testAsyncIsFinished() {
         CompletableFuture<Void> waitingFuture = newWaitingFuture();
         Future<Connection> future = client.service(waitingFuture);
-        await().atLeast(1000, TimeUnit.MILLISECONDS).untilAsserted(()-> Assert.assertTrue(future.isDone()));
+        await().atLeast(1000, TimeUnit.MILLISECONDS).atMost(2000, TimeUnit.MILLISECONDS)
+            .untilAsserted(()-> Assert.assertTrue(future.isDone()));
     }
 
 
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousTest.java
@@ -76,7 +76,6 @@ public class AsynchronousTest extends Arquillian {
         CompletableFuture<Void> waitingFuture = newWaitingFuture();
         Future<Connection> future = client.service(waitingFuture);
         Assert.assertFalse(future.isDone());
-        waitingFuture.complete(null);
     }
 
     /**
@@ -86,9 +85,9 @@ public class AsynchronousTest extends Arquillian {
     public void testAsyncIsFinished() {
         CompletableFuture<Void> waitingFuture = newWaitingFuture();
         Future<Connection> future = client.service(waitingFuture);
-        await().atMost(2000, TimeUnit.MILLISECONDS)
-            .untilAsserted(()-> Assert.assertTrue(future.isDone()));
         waitingFuture.complete(null);
+        await().atMost(30, TimeUnit.SECONDS)
+            .untilAsserted(()-> Assert.assertTrue(future.isDone()));
     }
 
 
@@ -100,7 +99,6 @@ public class AsynchronousTest extends Arquillian {
         CompletableFuture<Void> waitingFuture = newWaitingFuture();
         Future<Connection> future = clientClass.service(waitingFuture);
         Assert.assertFalse(future.isDone());
-        waitingFuture.complete(null);
     }
 
     /**
@@ -110,9 +108,9 @@ public class AsynchronousTest extends Arquillian {
     public void testClassLevelAsyncIsFinished() {
         CompletableFuture<Void> waitingFuture = newWaitingFuture();
         Future<Connection> future = clientClass.service(waitingFuture);
-        await().atMost(2000, TimeUnit.MILLISECONDS)
-            .untilAsserted(()-> Assert.assertTrue(future.isDone()));
         waitingFuture.complete(null);
+        await().atMost(30, TimeUnit.SECONDS)
+            .untilAsserted(()-> Assert.assertTrue(future.isDone()));
     }
 
     /**
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousTest.java
@@ -19,13 +19,11 @@
  *******************************************************************************/
 package org.eclipse.microprofile.fault.tolerance.tck;
 
-
-import java.util.concurrent.Future;
-
 import javax.inject.Inject;
 
 import org.eclipse.microprofile.fault.tolerance.tck.asynchronous.AsyncClassLevelClient;
 import org.eclipse.microprofile.fault.tolerance.tck.asynchronous.AsyncClient;
+import org.eclipse.microprofile.fault.tolerance.tck.asynchronous.CompletableFutureHelper;
 import org.eclipse.microprofile.fault.tolerance.tck.util.Connection;
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.testng.Arquillian;
@@ -35,6 +33,13 @@ import org.jboss.shrinkwrap.api.spec.JavaArchive;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
 import org.testng.Assert;
 import org.testng.annotations.Test;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+
+import static org.awaitility.Awaitility.await;
 
 /**
  * Verify the asynchronous invocation
@@ -51,7 +56,7 @@ public class AsynchronousTest extends Arquillian {
     public static WebArchive deploy() {
         JavaArchive testJar = ShrinkWrap
             .create(JavaArchive.class, "ftAsynchronous.jar")
-            .addClasses(AsyncClient.class, AsyncClassLevelClient.class, Connection.class)
+            .addClasses(AsyncClient.class, AsyncClassLevelClient.class, Connection.class, CompletableFutureHelper.class)
             .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml")
             .as(JavaArchive.class);
 
@@ -65,16 +70,10 @@ public class AsynchronousTest extends Arquillian {
      */
     @Test
     public void testAsyncIsNotFinished() {
-        Future<Connection> future = null;
-        try {
-            future = client.service();
-        }
-        catch (InterruptedException e) {
-            throw new AssertionError("testAsync: unexpected InterruptedException calling service");
-        }
-
-        Assert.assertFalse(future.isDone());
-
+        CompletableFuture<Void> waitingFuture = newWaitingFuture();
+        CompletionStage<Connection> resultFuture = client.serviceCS(waitingFuture);
+        Future<Connection> future = resultFuture.toCompletableFuture();
+        await().atMost(200, TimeUnit.MILLISECONDS).untilAsserted(()-> Assert.assertFalse(future.isDone()));
     }
 
     /**
@@ -82,17 +81,10 @@ public class AsynchronousTest extends Arquillian {
      */
     @Test
     public void testAsyncIsFinished() {
-        Future<Connection> future = null;
-        try {
-            future = client.service();
-            Thread.sleep(1500);
-        }
-        catch (InterruptedException e) {
-            throw new AssertionError("testAsync: unexpected InterruptedException calling service");
-        }
-
-        Assert.assertTrue(future.isDone());
-
+        CompletableFuture<Void> waitingFuture = newWaitingFuture();
+        CompletionStage<Connection> resultFuture = client.serviceCS(waitingFuture);
+        Future<Connection> future = resultFuture.toCompletableFuture();
+        await().atLeast(1000, TimeUnit.MILLISECONDS).untilAsserted(()-> Assert.assertTrue(future.isDone()));
     }
 
 
@@ -101,16 +93,10 @@ public class AsynchronousTest extends Arquillian {
      */
     @Test
     public void testClassLevelAsyncIsNotFinished() {
-        Future<Connection> future = null;
-        try {
-            future = clientClass.service();
-        }
-        catch (InterruptedException e) {
-            throw new AssertionError("testAsync: unexpected InterruptedException calling service");
-        }
-
-        Assert.assertFalse(future.isDone());
-
+        CompletableFuture<Void> waitingFuture = newWaitingFuture();
+        CompletionStage<Connection> resultFuture = clientClass.serviceCS(waitingFuture);
+        Future<Connection> future = resultFuture.toCompletableFuture();
+        await().atMost(400, TimeUnit.MILLISECONDS).untilAsserted(()-> Assert.assertFalse(future.isDone()));
     }
 
     /**
@@ -118,16 +104,20 @@ public class AsynchronousTest extends Arquillian {
      */
     @Test
     public void testClassLevelAsyncIsFinished() {
-        Future<Connection> future = null;
-        try {
-            future = clientClass.service();
-            Thread.sleep(1500);
-        }
-        catch (InterruptedException e) {
-            throw new AssertionError("testAsync: unexpected InterruptedException calling service");
-        }
+        CompletableFuture<Void> waitingFuture = newWaitingFuture();
+        CompletionStage<Connection> resultFuture = clientClass.serviceCS(waitingFuture);
+        Future<Connection> future = resultFuture.toCompletableFuture();
+        await().atLeast(1000, TimeUnit.MILLISECONDS).untilAsserted(()-> Assert.assertTrue(future.isDone()));
+    }
 
-        Assert.assertTrue(future.isDone());
-
+    /**
+     * Use this method to obtain futures for passing to methods on
+     * {@link AsyncClient}
+     * <p>
+     * Using this factory method ensures they will be completed at the end of
+     * the test if your test fails.
+     */
+    private CompletableFuture<Void> newWaitingFuture() {
+        return new CompletableFuture<>();
     }
 }
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousTest.java
@@ -76,6 +76,7 @@ public class AsynchronousTest extends Arquillian {
         CompletableFuture<Void> waitingFuture = newWaitingFuture();
         Future<Connection> future = client.service(waitingFuture);
         Assert.assertFalse(future.isDone());
+        waitingFuture.complete(null);
     }
 
     /**
@@ -87,6 +88,7 @@ public class AsynchronousTest extends Arquillian {
         Future<Connection> future = client.service(waitingFuture);
         await().atMost(2000, TimeUnit.MILLISECONDS)
             .untilAsserted(()-> Assert.assertTrue(future.isDone()));
+        waitingFuture.complete(null);
     }
 
 
@@ -98,6 +100,7 @@ public class AsynchronousTest extends Arquillian {
         CompletableFuture<Void> waitingFuture = newWaitingFuture();
         Future<Connection> future = clientClass.service(waitingFuture);
         Assert.assertFalse(future.isDone());
+        waitingFuture.complete(null);
     }
 
     /**
@@ -109,6 +112,7 @@ public class AsynchronousTest extends Arquillian {
         Future<Connection> future = clientClass.service(waitingFuture);
         await().atMost(2000, TimeUnit.MILLISECONDS)
             .untilAsserted(()-> Assert.assertTrue(future.isDone()));
+        waitingFuture.complete(null);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncClassLevelClient.java
```diff
@@ -41,28 +41,6 @@
...
-
-            @Override
-            public String getData() {
-                return "service DATA";
-            }
-        };
-
-        return CompletableFuture.completedFuture(conn);
-    }
-
```
Ladicek: Unless we're about to remove support for `Future` as a return type of `@Asynchronous` methods (which I'd totally agree with :-) ), I think we need both `service` and `serviceCS`. It should be possible to rewrite `service` to accept a `waitCondition` similarly to `serviceCS`.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncClient.java
```diff
@@ -40,29 +40,6 @@
...
-
-            @Override
-            public String getData() {
-                return "service DATA";
-            }
-        };
-
-        return CompletableFuture.completedFuture(conn);
-    }
-
```
Ladicek: Same here.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousTest.java
```diff
@@ -65,34 +70,21 @@ public static WebArchive deploy() {
...
-        catch (InterruptedException e) {
-            throw new AssertionError("testAsync: unexpected InterruptedException calling service");
-        }
-
-        Assert.assertFalse(future.isDone());
-
+        CompletableFuture<Void> waitingFuture = newWaitingFuture();
+        CompletionStage<Connection> resultFuture = client.serviceCS(waitingFuture);
+        Future<Connection> future = resultFuture.toCompletableFuture();
+        await().atMost(200, TimeUnit.MILLISECONDS).untilAsserted(()-> Assert.assertFalse(future.isDone()));
```
Ladicek: Assuming there is a `service` method taking `waitingFuture` and returning `Future`, I think this test should look roughly like this:

```java
CompletableFuture<Void> waitingFuture = newWaitingFuture();
Future<Connection> future = client.service(waitingFuture);
assertFalse(future.isDone());
```

There's no need for awaiting the `!isDone()` state.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousTest.java
```diff
@@ -65,34 +70,21 @@ public static WebArchive deploy() {
...
-        catch (InterruptedException e) {
-            throw new AssertionError("testAsync: unexpected InterruptedException calling service");
-        }
-
-        Assert.assertTrue(future.isDone());
-
+        CompletableFuture<Void> waitingFuture = newWaitingFuture();
+        CompletionStage<Connection> resultFuture = client.serviceCS(waitingFuture);
+        Future<Connection> future = resultFuture.toCompletableFuture();
+        await().atLeast(1000, TimeUnit.MILLISECONDS).untilAsserted(()-> Assert.assertTrue(future.isDone()));
```
Ladicek: Assuming there is a `service` method taking `waitingFuture` and returning `Future`, I think this test should look roughly like this:

```java
CompletableFuture<Void> waitingFuture = newWaitingFuture();
Future<Connection> future = client.service(waitingFuture);
assertFalse(future.isDone());
waitingFuture.complete(null);
await().atMost(1000, TimeUnit.MILLISECONDS).untilAsserted(()-> Assert.assertTrue(future.isDone()));
```

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncClassLevelClient.java
```diff
@@ -41,28 +41,6 @@
...
-            public String getData() {
-                return "service DATA";
-            }
-        };
-
-        return CompletableFuture.completedFuture(conn);
-    }
-    
     /**
      * Service an operation until waitCondition is completed or 1 second timeout.
```
Ladicek: I'm actually not sure why we use the 1 second timeout. All tests should be able to complete the `waitCondition` future at the right point in time. But that's definitely not for this PR to change.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousTest.java
Ladicek: Alternatively, if we don't want to complete the wait condition and we're fine with useless 1 second wait inside the service method, this would be equivalent I think:

```java
CompletableFuture<Void> waitingFuture = newWaitingFuture();
Future<Connection> future = client.service(waitingFuture);
assertFalse(future.isDone());
await().atLeast(1000, TimeUnit.MILLISECONDS).atMost(2000, TimeUnit.MILLISECONDS).untilAsserted(()-> Assert.assertTrue(future.isDone()));
```

Ladicek: And the `assertFalse(future.isDone());` is not 100% necessary, that's covered by the other test. But it shouldn't hurt :-)

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncClassLevelClient.java
carlosdlr: Well, we have ongoing something that allows using the config spec to pass a timeout as a parameter for all the TCKS the last pull request related to this:  https://github.com/eclipse/microprofile-fault-tolerance/pull/403. but looks that were not finished. i think we will good retake this.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncClient.java
carlosdlr: done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousTest.java
carlosdlr: done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousTest.java
carlosdlr: done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousTest.java
```diff
@@ -101,33 +91,31 @@ public void testAsyncIsFinished() {
...
-            future = clientClass.service();
-        }
-        catch (InterruptedException e) {
-            throw new AssertionError("testAsync: unexpected InterruptedException calling service");
-        }
-
-        Assert.assertFalse(future.isDone());
-
+        CompletableFuture<Void> waitingFuture = newWaitingFuture();
+        CompletionStage<Connection> resultFuture = clientClass.serviceCS(waitingFuture);
```
Ladicek: The original test called `service()`, so this should also call `service()`.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousTest.java
```diff
@@ -101,33 +91,31 @@ public void testAsyncIsFinished() {
...
-            future = clientClass.service();
-            Thread.sleep(1500);
-        }
-        catch (InterruptedException e) {
-            throw new AssertionError("testAsync: unexpected InterruptedException calling service");
-        }
-
-        Assert.assertTrue(future.isDone());
+        CompletableFuture<Void> waitingFuture = newWaitingFuture();
+        CompletionStage<Connection> resultFuture = clientClass.serviceCS(waitingFuture);
```
Ladicek: Same here, should call `service()`, not `serviceCS()`.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncClassLevelClient.java
```diff
@@ -41,28 +41,16 @@
...
-            @Override
-            public String getData() {
-                return "service DATA";
-            }
-        };
-
-        return CompletableFuture.completedFuture(conn);
+    @Asynchronous
+    public Future<Connection> service(Future<?> waitCondition) {
+        return serviceCS(waitCondition, false).toCompletableFuture();
```
Ladicek: Yea, this is probably the easiest impl :-) But we should use the `CompletableFutureHelper` to convert `CompletionStage` to `CompletableFuture`, as `CompletionStage.toCompletableFuture()` doesn't have to be supported.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncClient.java
```diff
@@ -41,26 +41,13 @@
...
-            @Override
-            public String getData() {
-                return "service DATA";
-            }
-        };
-
-        return CompletableFuture.completedFuture(conn);
+    public Future<Connection> service(Future<?> waitCondition) {
+        return serviceCS(waitCondition, false).toCompletableFuture();
```
Ladicek: Same here, should use `CompletableFutureHelper`.

carlosdlr: done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncClassLevelClient.java
carlosdlr: done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousTest.java
carlosdlr: done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousTest.java
carlosdlr: done

carlosdlr:
This test group was tested against liberty implementation

carlosdlr:
> 'd much rather get rid of the 1 sec timeout entirely and always complete the `waitCondition` futures explicitly. But perhaps that's just me :-)

Thanks, @Ladicek for your feedback this is the beginning of this refactor so these considerations can be taking into account in the next pull requests to gradually improve the tcks.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousTest.java
```diff
@@ -65,34 +68,20 @@ public static WebArchive deploy() {
...
-        Future<Connection> future = null;
-        try {
-            future = client.service();
-        }
-        catch (InterruptedException e) {
-            throw new AssertionError("testAsync: unexpected InterruptedException calling service");
-        }
-
+        CompletableFuture<Void> waitingFuture = newWaitingFuture();
+        Future<Connection> future = client.service(waitingFuture);
```
Azquelt: Can we ensure that the waitingFuture is completed, otherwise the asynchronous task may run forever?

Azquelt: We can use the same pattern we use in other tests to ensure that waitingFutures are closed at the end of the test, I'll find an example.

Azquelt: See these parts of AsynchronousCSTest:
* https://github.com/eclipse/microprofile-fault-tolerance/blob/master/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousCSTest.java#L61
* https://github.com/eclipse/microprofile-fault-tolerance/blob/master/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousCSTest.java#L250
* https://github.com/eclipse/microprofile-fault-tolerance/blob/master/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousCSTest.java#L264

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousTest.java
```diff
@@ -65,34 +68,20 @@ public static WebArchive deploy() {
...
-        }
-        catch (InterruptedException e) {
-            throw new AssertionError("testAsync: unexpected InterruptedException calling service");
-        }
-
-        Assert.assertTrue(future.isDone());
-
+        CompletableFuture<Void> waitingFuture = newWaitingFuture();
+        Future<Connection> future = client.service(waitingFuture);
+        await().atLeast(1000, TimeUnit.MILLISECONDS).atMost(2000, TimeUnit.MILLISECONDS)
```
Emily-Jiang: @Ladicek suggested not to wait for 1s ` .atLeast(1000, TimeUnit.MILLISECONDS)` is to be removed.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousTest.java
carlosdlr: done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousTest.java
carlosdlr: done

carlosdlr:
Changes related to your comments were done and tested against liberty implementation

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousTest.java
```diff
@@ -109,6 +112,7 @@ public void testClassLevelAsyncIsFinished() {
         Future<Connection> future = clientClass.service(waitingFuture);
         await().atMost(2000, TimeUnit.MILLISECONDS)
             .untilAsserted(()-> Assert.assertTrue(future.isDone()));
+        waitingFuture.complete(null);
```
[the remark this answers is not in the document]
Ladicek: Indeed the way you added the `future.complete()` calls is equivalent to the `@AfterTest` method and is completely unnecessary.

What I'm saying is that the `test*AsyncIsFinished` tests should complete the waiting future _before_ the assert. If you don't do that, the test will needlessly wait one second.

In fact, I also think that your should revert the timeout in the `serviceCS` method back to 1000 seconds, to make sure all callers actually complete the waiting future instead of relying on the timeout.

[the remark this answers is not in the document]
Azquelt: Except for `testAsyncIsNotFinished` which doesn't need to complete the future (or indeed wait for anything)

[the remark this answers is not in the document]
carlosdlr: Thanks @Ladicek I will make the changes accordingly

Ladicek:
@Azquelt is of course right -- I specifically said `test*AsyncIsFinished`, which doesn't include the `*NotFinished` tests. Those should not complete the future, and indeed don't have to wait for anything.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsynchronousTest.java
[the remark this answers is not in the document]
Ladicek: @Azquelt is of course right -- I specifically said `test*AsyncIsFinished`, which doesn't include the `*NotFinished` tests. Those should not complete the future, and indeed don't have to wait for anything.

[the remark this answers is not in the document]
Azquelt: Doh, I should read more carefully :wink:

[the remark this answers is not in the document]
carlosdlr: Done guys, thanks for your feedback

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncClassLevelClient.java
```diff
@@ -63,7 +63,7 @@
 
         Throwable exception = null;
         try {
-            waitCondition.get(1000, TimeUnit.MILLISECONDS);
+            waitCondition.get(1000, TimeUnit.SECONDS);
```
Emily-Jiang: @carlosdlr can you add a comment to say "This timeout will normally be ignored as the tests will time out before this timeout. The reason for this large value is to ensure test timeout kicks in before this one."?

carlosdlr: Done.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncClassLevelClient.java
```diff
@@ -52,8 +52,12 @@
     }
 
     /**
-     * Service an operation until waitCondition is completed or 1 second timeout.
-     *
+     * Service an operation until waitCondition is completed or 1000 second timeout.
+     * NOTE: The timeout used in the waiting condition (1000 SECONDS) will simulate a large service processing time, to allow use in the
+     * test cases a strategy to improve the waiting time and avoid has process waiting for ever the strategy is described below:
```
Emily-Jiang: I suggest to delete this:
_, to allow use in the test cases a strategy to improve the waiting time and avoid has process waiting for ever the strategy is described below: 
n the cases that are necessary test if the @{@link Future} is finished we can me make this finish immediately and use awaitility library to control at most how much time will wait for the test before to finish in cases that the service takes more time due to external factors like infrastructure in a continuous integration context or stress testing.
_
as it does not read well and does not add much.

Please just add:

This 1000 second timeout is to ensure test timeout kicks in before the operation timeout for a better test error to be displayed.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncClient.java
```diff
@@ -73,7 +73,20 @@
...
         return serviceCS(waitCondition, false, stageToReturn);
     }
-    
+
+    /**
+     * Helper method that allows to complete a future in an exceptional or a non exceptional way also simulates a
+     * service processing waiting time.
+     * NOTE: The timeout used in the waiting condition (1000 SECONDS) will simulate a large service processing time, to allow use in the
+     * test cases a strategy to improve the waiting time and avoid has process waiting for ever the strategy is described below:
+     * In the cases that are necessary test if the @{@link Future} is finished we can me make this finish immediately and use
```
Emily-Jiang: Please see my previous comments.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncClassLevelClient.java
carlosdlr: done.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncClient.java
carlosdlr: done. thanks @Emily-Jiang

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncClassLevelClient.java
```diff
@@ -41,31 +41,19 @@
 @Asynchronous
 public class AsyncClassLevelClient {
 
-
     /**
-     * service 1 second in normal operation.
-     * @return the result as a Future
-     * @throws InterruptedException the interrupted exception
+     * Service an operation until waitCondition is completed or 1 second timeout.
```
Ladicek: This should probably read the same as the `serviceCS` javadoc. Or perhaps we should just say `@see #serviceCS`.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncClient.java
```diff
@@ -41,26 +41,13 @@
 public class AsyncClient {
 
     /**
-     * service 1 second in normal operation.
-     *
-     * @return the result as a Future
-     * @throws InterruptedException the interrupted exception
+     * Service an operation until waitCondition is completed or 1 second timeout.
+     * @param waitCondition Execution of this method will delay until the condition is finished
+     * @return the result as a Future.
```
Ladicek: Same here, javadoc is incorrect. (It wasn't correct before either.)

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncClassLevelClient.java
carlosdlr: done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncClient.java
carlosdlr: done

PR #432, opened 2019-06-07
switch back to MP Metrics 1.1 and use reflection to also be compatible with MP Metrics 2.0

Ladicek:


genie-microprofile:

Azquelt:

PR #438, opened 2019-08-14
Add a clarifying note on running retries on the same thread

### LINKED ISSUE #424: Improve spec on use of threads for retry and asynchronous
7.1 says
> If `@Retry` is used with `@Asynchronous` and a retry is required, the new retry attempt may be run on the same thread as the previous attempt, or on a different thread.

This suggests that doing all retries on the same thread is a valid strategy.

However, 6.1 says
> If `@Timeout` is used with `@Retry`, a `TimeoutException` may trigger a retry, depending on the values of `retryOn` and `abortOn` of the `@Retry` annotation. The timeout is restarted for each retry. If `@Asynchronous` is also used and the retry is the result of a `TimeoutException`, the retry starts after any delay period, even if the original attempt is still running.

This effectively precludes doing all retries on the same thread in cases where a `TimeoutException` occurs.

@jbee pointed out that it's unhelpful to suggest that either strategy is valid in one part of the spec and then invalidate one of those strategies in another part.

Azquelt:
Avoid giving the impression that running retries on the same thread is
valid in all cases.

Fixes #424

### DIFF
--- a/spec/src/main/asciidoc/retry.asciidoc
+++ b/spec/src/main/asciidoc/retry.asciidoc
@@ -86,4 +86,4 @@ The `@Retry` annotation can be used together with `@Fallback`, `@CircuitBreaker`
 
 A `@Fallback` can be specified and it will be invoked if the method still fails after any retires have been run.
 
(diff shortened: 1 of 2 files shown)


carlosdlr:
Thanks @Azquelt and @Ladicek

Issue #418, opened 2019-03-13
Exception-Handling in CircuitBreaker


mfrembs:
Hi,

this is an extra issue for the same use-case I talked about here [1] but this time for the CircuitBreaker and not Fallback.
[1] https://github.com/eclipse/microprofile-fault-tolerance/issues/417

I implemented a MP-Rest-client with a ResponseExceptionMapper. The ResponseExceptionMapper throws "MyException" if the HTTP-status is >= 300. If the GET-request succeeds I get "MyPojo" filled with the provided information.

Now I want to make that call resilient. So I annotate the method with CircuitBreaker. The problem I have is, that the CircuitBreaker is getting also "activated" when the client throws MyException. But this Exception should not be "catched" by the CircuitBreaker. Right now there is only the possibility of a whitelist with "failOn".

Do I oversee here something? I would like to get a blacklist with exceptions when the CircuitBreaker is not getting activated.

Thanks!

Emily-Jiang:
Are you just playing with the annotations or this is your real business scneario?

Emily-Jiang:
We discussed today, we could add another method skipOn() . Here are some thoughts?
- failOn()
 and skipOn() have no overlapping - easy, use failOn

- failOn() and skipOn() overlap, 
failOn(MyException1.class, MyException2.class)
skipOn(MyException1.class) 
Should we join then or they are mutually exclusive?
Thoughts?

def456:
Right now I am building a reference application for a customer. But this will be sooner or later become a real business scenario. I'm a bit confused that no one run into this issue before me ;-)

My first thought was: Why would you define a whitelist and a blacklist together. But my further thought was: that could make sense. Imagine the following use case:

Global "MyServiceException"
With following Exceptions inheriting: "MyConnectionException", "MyValidationException", "MyIOException".

Now I want to fail on MyServiceException (and therefor also on the other 3) but I want to skip on MyValidationException. So I would annotate it with `failOn = MyServiceException.class, skipOn = MyIOException.calls`. Can you follow me?

This way it would be more flexible and you should have every use case.

For my use case the easy one (mutually exclusive) would be sufficient, but I would recommend to do the more flexible option.

michalszynkiewicz:
If we allow to have both, I think it should work as follows:
```java
class E0 extends Exception {...}

class E1 extends E0 {...}

class E2 extends E1 {..}

[...]

PR #440, opened 2019-08-19
test @Retry behavior with various exceptions in an inheritance hierarchy

Ladicek:

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryExceptionHierarchyTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryExceptionHierarchyTest.java
@@ -1,0 +1,271 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+package org.eclipse.microprofile.fault.tolerance.tck;
+
+import org.eclipse.microprofile.fault.tolerance.tck.retry.exception.hierarchy.E0;
+import org.eclipse.microprofile.fault.tolerance.tck.retry.exception.hierarchy.E0S;
+import org.eclipse.microprofile.fault.tolerance.tck.retry.exception.hierarchy.E1;
+import org.eclipse.microprofile.fault.tolerance.tck.retry.exception.hierarchy.E1S;
+import org.eclipse.microprofile.fault.tolerance.tck.retry.exception.hierarchy.E2;
+import org.eclipse.microprofile.fault.tolerance.tck.retry.exception.hierarchy.E2S;
+import org.eclipse.microprofile.fault.tolerance.tck.retry.exception.hierarchy.RetryService;
+import org.eclipse.microprofile.fault.tolerance.tck.retry.exception.hierarchy.RetryStatus;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.testng.Arquillian;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.EmptyAsset;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.testng.annotations.Test;
+
+import javax.inject.Inject;
+
+import static org.testng.Assert.assertEquals;
+
+/**
+ * Verifies behavior of {@code @Retry} for various exceptions in an inheritance hierarchy.
+ * The asserted logic is:
+ * <ol>
+ *     <li>Abort if the exception matches abortOn</li>
+ *     <li>Otherwise retry if the exception matches retryOn</li>
+ *     <li>Otherwise abort</li>
+ * </ol>
+ */
+public class RetryExceptionHierarchyTest extends Arquillian {
+    @Deployment
+    public static WebArchive deploy() {
+        JavaArchive jar = ShrinkWrap.create(JavaArchive.class, "ftRetryExceptionHierarchy.jar")
+            .addClasses(E0.class, E0S.class, E1.class, E1S.class, E2.class, E2S.class, RetryStatus.class, RetryService.class)
+            .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml")
+            .as(JavaArchive.class);
+
+        return ShrinkWrap
+            .create(WebArchive.class, "ftRetryExceptionHierarchy.war")
+            .addAsLibrary(jar);
+    }
+
+    @Inject
+    private RetryService service;
+
+    // the <: symbol denotes the subtyping relation (Foo <: Bar means "Foo is a subtype of Bar")
+    // note that subtyping is reflexive (Foo <: Foo)
+
+    // E0  <: Exception
+    // E1  <: E0
+    // E2  <: E1
+    // E2S <: E2
+    // E1S <: E1, but not E1S <: E2
+    // E0S <: E0, but not E0S <: E1
+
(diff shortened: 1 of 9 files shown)


genie-microprofile:

Azquelt:

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryExceptionHierarchyTest.java
```diff
@@ -0,0 +1,216 @@
...
+    @Test
+    public void serviceBthrowsRuntimeException() {
+        assertEquals(invokeserviceB(new RuntimeException()), RetryStatus.RETRIED_INVOCATION);
+    }
+
+    @Test
+    public void serviceBthrowsThrowable() {
+        // I'd rather expect RetryStatus.FIRST_INVOCATION
+        // BUT: this is probably affected by @Retry being specified to ignore custom Throwable-s, which is stupid
+        assertEquals(invokeserviceB(new Throwable()), RetryStatus.RETRIED_INVOCATION);
```
Azquelt: I don't think this is right, but I was surprised to find that this passes on openliberty.

In our case, it's because the `Throwable` has been wrapped by a `WeldException` by the time it's caught by the Fault Tolerance interceptor. This is presumably because `InvocationContext.proceed()` is only declared to throw `Exception`.

If I change the test to use `Error` instead of `Throwable` then the test fails as I would expect because no retry occurs.

We should decide whether custom `Throwable` is something that we want to test.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryExceptionHierarchyTest.java
```diff
@@ -0,0 +1,216 @@
...
+
+/**
+ * Verifies behavior of {@code @Retry} for various exceptions in an inheritance hierarchy.
+ * Based on https://github.com/eclipse/microprofile-fault-tolerance/issues/418#issuecomment-476112605
+ */
+public class RetryExceptionHierarchyTest extends Arquillian {
+    @Deployment
+    public static WebArchive deploy() {
+        JavaArchive jar = ShrinkWrap.create(JavaArchive.class, "ftRetryExceptionHierarchy.jar")
+            .addClasses(E0.class, E0S.class, E1.class, E1S.class, E2.class, E2S.class, RetryStatus.class, RetryService.class)
```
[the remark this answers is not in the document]
Ladicek: I will pull the classes out of the test, to make them top-level classes. This was just to quickly throw something together :-)

Ladicek:
> the only thing that might make it clearer would be having a comment on each test explaining why it expects the result it does. I found it a little difficult to keep in my head the annotations for serviceA and serviceB while I was reading each test

Very good point, I'll try. BUT! I originally wrote the tests to conform to these tables: https://github.com/eclipse/microprofile-fault-tolerance/issues/418#issuecomment-476112605 I can easily explain the expectations there. And then, I changed the asserts to let all the tests pass, and I'm no longer sure if there's any kind of logic there.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryExceptionHierarchyTest.java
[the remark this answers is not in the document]
Ladicek: I will pull the classes out of the test, to make them top-level classes. This was just to quickly throw something together :-)

Azquelt:
Given the assertions in your PR. I think the currently logic that's asserted is just:
1) Abort if the exception matches `abortOn`
2) Otherwise retry if the exception matches `retryOn`
3) Otherwise abort

Plus the small wrinkle for the test with `Throwable` which I think is caused by the limitations of the `InvocationContext` interface.

Ladicek:
Hmm yea, that's probably the algorithm we use in SmallRye FT, I never really checked :-)

I pushed a changeset that should address your comments.

I wonder: does OpenLiberty behave the same?

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryExceptionHierarchyTest.java
Ladicek: Custom `Throwable` is a corner case, but testing corner cases is IMHO what makes a difference :-)

Azquelt: True, my only concern is whether it would make it impossible to write a portable implementation. I can't see anything in the Interceptors spec about what happens in an `@AroundInvoke` interceptor if a `Throwable` is thrown from a business method.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryExceptionHierarchyTest.java
```diff
@@ -0,0 +1,196 @@
...
+        catch (Throwable expected) {
+        }
+        return service.getStatus();
+    }
+
+    // serviceB: @Retry(retryOn = {Exception.class, E1.class}, abortOn = {E0.class, E2.class})
+
+    @Test
+    public void serviceBthrowsException() {
+        // serviceA directly mentions Exception in retryOn
```
Azquelt: These all say `serviceA` instead of `serviceB`

Ladicek: Aaaah, copy'n'paste is a beast!

Azquelt:
Changes look good but I'm seeing a checkstyle error when running `mvn install`

Azquelt:
Yes the behaviour in openliberty is the same, the tests all pass there without modification.

Ladicek:
Interesting, so we converged on the same behavior without syncing :-)

We could just specify it. Or we could decide we want to break compat and specify something else. I guess we'll probably want to discuss that on the call.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryExceptionHierarchyTest.java
Ladicek: OK, I had to debug our impl to understand, and same thing happens in our case -- `Throwable` is wrapped into `WeldException` by Weld's `AroundInvokeInvocationContext.proceed`.

I'm looking into chapter 2.5 of the Interceptors 1.2 spec, and it doesn't really answer the question for me.

But we probably don't need to hold this work because of this corner case.

Ladicek: So technically, Interceptors spec, version 1.2, chapter 2.5 says:

> Interceptor methods are allowed to throw runtime exceptions or any checked exceptions that the associated target method or constructor allows within its `throws` clause.
>
> ...
>
> The invocation of the `InvocationContext.proceed` method throws the same exception as any thrown by the associated target method ...

And the Java Language Specification, version 11, chapter 11.1.1 says:

> The _unchecked exception_ classes are the run-time exception classes and the error classes.
>
> The _checked exception_ classes are all exception classes other than the unchecked exception classes. That is, the checked exception classes are _Throwable_ and all its subclasses other than _RuntimeException_ and its subclasses and _Error_ and its subclasses.

So `Throwable` is a checked exception, and interceptors should handle it transparently.

At the same time, JLS also says:

> Exception is the superclass of all the exceptions from which ordinary programs may wish to recover.

So I'm not quite sure what to think of it.

Shall we comment the `Throwable` tests out for now? I'm probably making too big of a deal from this :-)

Azquelt: Yeah, I'd comment it out for now and we can discuss it on the call.

Ladicek: Just out of curiosity, I'm looking at what OpenWebBeans do:

- https://github.com/apache/openwebbeans/blob/master/webbeans-impl/src/main/java/org/apache/webbeans/intercept/InterceptorInvocationContext.java#L51
- https://github.com/apache/openwebbeans/blob/master/webbeans-impl/src/main/java/org/apache/webbeans/intercept/AbstractInvocationContext.java#L118
- https://github.com/apache/openwebbeans/blob/master/webbeans-impl/src/main/java/org/apache/webbeans/util/ExceptionUtil.java#L34

It seems they would rethrow "sneakily" even a `Throwable`, but I didn't try it :-)

In any case, gonna disable these tests for now.

Azquelt:
@genie-microprofile please test this

Ladicek:
Added the `serviceC: @Retry(retryOn = {E1.class, E2.class}, abortOn = E0.class)` case.

Azquelt:

Azquelt:
These tests are great. I'm unsure whether we want to wait until we decide which behaviour we want, or whether we should merge these tests now and update them if we decide on a different behaviour.

Ladicek:
This shall be merged after #446.

PR #444, opened 2019-09-05
specify how custom throwables are treated

### LINKED ISSUE #441: clarify the behavior in case intercepted methods throw custom Throwable
We agreed on the call last week (2019-08-20) that:
- throwing custom throwables from business methods intercepted by MP FT results in a non-portable behavior from the MP FT perspective;
- custom throwables specified in the MP FT annotations (such as the `retryOn` or `abortOn` attributes of `@Retry`, or the `failOn` attribute of `@CircuitBreaker`) should always be ignored. It is currently specified that they are ignored for the `@Retry` attributes, but not for `@CircuitBreaker`. And if we add similar attributes to the `@Fallback` annotations, they should also ignore custom throwables.

By "custom throwable", we mean any class that is a subtype of `Throwable` but isn't a subtype neither of `Error` nor `Exception`.

I agreed to draft a spec change and submit a PR, but didn't have time for it yet. I still intend to do it, but if someone else beats me to it, that's fine too :-)

Ladicek:
Resolves #441.

### DIFF
--- a/spec/src/main/asciidoc/fault-tolerance-interceptor.asciidoc
+++ b/spec/src/main/asciidoc/fault-tolerance-interceptor.asciidoc
@@ -33,25 +33,32 @@ For instance, in the following example, `MyLogInterceptor` will be invoked first
 [source, java]
 ----
 @Retry
-@MyLog 
+@MyLog
 @MyPrint
-public void myInvok() {
-
- //do something
+public void myInvoke() {
+  // do something
 }
+
 @Priority(3000)
 @MyLog
-public class MyLogInterceptor{
- //do logging
+public class MyLogInterceptor {
+  // do logging
 }
 
 @Priority(5000)
 @MyPrint
-public class MyPrintInterceptor{
- //do printing
+public class MyPrintInterceptor {
+  // do printing
 }
 ----
 
+=== Custom throwables
 
+Throwing custom throwables from business methods annotated with any of the Fault Tolerance interceptor bindings results in non-portable behavior.
+The term "custom throwable" means: any class that is a subtype of `Throwable`, but isn't a subtype of `Error` or `Exception`.
+This includes `Throwable` itself, and _doesn't_ include `Error` and `Exception`.
 
-
+[NOTE]
+Some Fault Tolerance annotations allow configuring a set of exception types for various purposes.
+For example, `@Retry` includes the `retryOn` attribute which configures the set of exceptions on which retry will be performed.
+In these cases, it is possible to specify `Throwable` and it is guaranteed to cover all ``Error``s and ``Exception``s.


genie-microprofile:

Azquelt:

spec/src/main/asciidoc/fault-tolerance-interceptor.asciidoc
```diff
@@ -33,25 +33,32 @@ For instance, in the following example, `MyLogInterceptor` will be invoked first
...
 @Priority(5000)
 @MyPrint
-public class MyPrintInterceptor{
- //do printing
+public class MyPrintInterceptor {
+  // do printing
 }
 ----
 
+=== Treating custom throwables
```
Azquelt: I think either "Treatment of custom throwables" or just "Custom throwables" would work better as the heading on this section.

spec/src/main/asciidoc/fault-tolerance-interceptor.asciidoc
```diff
@@ -33,25 +33,32 @@ For instance, in the following example, `MyLogInterceptor` will be invoked first
...
 
+Throwing custom throwables from business methods annotated with any of the Fault Tolerance interceptor bindings results in non-portable behavior.
+The term "custom throwable" means: any class that is a subtype of `Throwable`, but isn't a subtype of `Error` or `Exception`.
+This includes `Throwable` itself, and _doesn't_ include `Error` and `Exception`.
 
-
+Some Fault Tolerance annotations allow configuring a set of exception types for various purposes.
+For example, `@Retry` includes the `retryOn` attribute which configures the set of exceptions on which retry will be performed.
+In any of these cases, specifying a custom throwable is always ignored.
```
Azquelt: It seems a bit odd to say that throwing a custom `Throwable` results in undefined behaviour, but listing a custom `Throwable` in an annotation parameter results in it being ignored.

My initial thought was that we should a TCK to check that the custom throwable is ignored, but it's hard to write a good test when throwing the custom throwable results in undefined behaviour.

The only test I can think of that we could actually write would be to have something like `@Retry(retryOn=Throwable.class)` and to test that with both an `Exception` and an `Error`.

However, this spec now says that `Throwable.class` should be _ignored_ in the `retryOn` parameter, which means that in this test _nothing_ should be retried, which itself seems highly unintuitive.

I suggest that we don't need to say anything about the types listed in annotation parameters, and merely state that "Throwing a `Throwable` which is not assignable to either `Exception` or `Error` results in undefined behaviour."

Ladicek: So, I did think about this, and my initial decision was to stick to current `@Retry` behavior. Because `@Retry` already specifies that custom throwables are ignored in `retryOn` and `abortOn`. Here, I just changed `@CircuitBreaker` to follow suit. We can do the opposite -- change `@Retry` to be the same as `@CircuitBreaker`, but IMHO we should align them one way or the other.

spec/src/main/asciidoc/fault-tolerance-interceptor.asciidoc
Ladicek: Good point, changed to "Custom throwables".

spec/src/main/asciidoc/fault-tolerance-interceptor.asciidoc
Azquelt: Ah sorry, I misread the diff and thought you'd added that part for `Retry`. You can ignore this comment in that case.

I agree that they should at least be consistent, whichever behaviour we decide on.

I think this is worth more discussion though. With the current wording, I think the only possible test is the one that I outlined and the specified result is (at least to me) unexpected.

Ladicek: Sure, let's discuss this on the call tomorrow.

Ladicek:
Updated. This PR no longer touches `@CircuitBreaker` at all, and removes the language around "custom throwables are ignored" in `@Retry`.

Ladicek:
Previously, this PR also contained some changes to the `@Retry` annotation, but I've removed them, as they are also present in #446. Now, this PR only touches the specification text. It should probably be merged first, and after that, we can merge other PRs that mention "custom throwables are non-portable" in javadocs etc.

PR #446, opened 2019-09-05
specify the meaning of overlapping retryOn and abortOn

### LINKED ISSUE #442: clarify retryOn/abortOn behavior of @Retry
As we discovered, both SmallRye and OpenLiberty behave the same when it comes to overlapping set of throwables in `retryOn` and `abortOn`: blacklist wins over whitelist. We agreed to codify this behavior.

(Where `retryOn` is the "whitelist" and `abortOn` is the "blacklist". The guiding principle here seems to be: the normal [= white] thing to do for `@Retry` in case of an exception is to retry. You can configure an exception [= black] to that behavior.)

### LINKED ISSUE #449: Revisit behaviour when a custom throwable is specified for retryOn or abortOn
In #441, we also said that a method throwing a custom `Throwable` results in non-portable behaviour, and in reviewing that, I found some issues with what happens when the user _specifies_ a custom Throwable in an annotation.

The Javadoc for the `retryOn` and `abortOn` parameters for `@Retry` says

> Only a type assignable from {@link java.lang.Error} or {@link java.lang.Exception} can be specified.
> Specifying a custom java.lang.Throwable will be ignored.

This has some interesting consequences:
1. If the user does `@Retry(retryOn=MyCustomThrowable.class)`, it gets ignored and no Retries occur. Implementations could emit a warning here, but there's no deployment time exception.
1. If the user does `@Retry(retryOn=Throwable.class)`, it also gets ignored (as `Throwable` is not a subclass of `Error` or `Exception`). Again, no deployment error.
   * It's worth noting that doing this is a bad idea. An `Error` is often something fairly fatal (like OOM) so retrying is not likely to be helpful.
1. If we want to be consistent, then we should define the `@CircuitBreaker` `failOn` and `skipOn` parameters in the same way. However, this would mean that the _default_ configuration of `@CircuitBreaker(failOn=Throwable.class)` would be ignored, obviously not the desired behaviour.

Ladicek:

Resolves #442.
Resolves #449.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java
+++ b/api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java
@@ -30,9 +30,17 @@ import javax.enterprise.util.Nonbinding;
 import javax.interceptor.InterceptorBinding;
 
 /**
- * The Retry annotation to define the number of the retries and the fallback method on reaching the
- * retry counts. Any invalid config value causes 
+ * The Retry annotation to define the number of the retries. Any invalid config value causes
  * {@link org.eclipse.microprofile.faulttolerance.exceptions.FaultToleranceDefinitionException}.
+ * <p>
+ * When a method returns and the retry policy is present, the following rules are applied:
+ * <ol>
+ * <li>If the method returns normally (doesn't throw), the result is simply returned.
+ * <li>Otherwise, if the thrown object is assignable to any value in the {@link #abortOn()} parameter, the thrown object is rethrown.
+ * <li>Otherwise, if the thrown object is assignable to any value in the {@link #retryOn()} parameter, the method call is retried.
+ * <li>Otherwise the thrown object is rethrown.
+ * </ol>
+ * If a method throws a {@link Throwable} which is not an {@link Error} or {@link Exception}, non-portable behavior results.
  * @author <a href="mailto:emijiang@uk.ibm.com">Emily Jiang</a>
  * @author John Ament
  */
@@ -101,17 +109,23 @@ public @interface Retry {
 
 
     /**
+     * The list of exception types which should trigger a retry.
+     * <p>
+     * Note that if a method throws a {@link Throwable} which is not an {@link Error} or {@link Exception}, non-portable behavior results.
      *
-     * @return Specify the failure to retry on. Only a type assignable from {@link java.lang.Error} or {@link java.lang.Exception} can be specified.
-     * Specifying a custom java.lang.Throwable will be ignored.
+     * @return the exception types on which to retry
      */
     @Nonbinding
     Class<? extends Throwable>[] retryOn() default { Exception.class };
 
     /**
-     * 
-     * @return Specify the failure to abort on. Only a type assignable from {@link java.lang.Error} or {@link java.lang.Exception} can be specified.
-     * Specifying a custom java.lang.Throwable will be ignored.
+     * The list of exception types which should <i>not</i> trigger a retry.
+     * <p>
+     * This list takes priority over the types listed in {@link #retryOn()}.
+     * <p>
+     * Note that if a method throws a {@link Throwable} which is not an {@link Error} or {@link Exception}, non-portable behavior results.
+     *
+     * @return the exception types on which to abort (not retry)
      */
     @Nonbinding
     Class<? extends Throwable>[] abortOn() default {};


genie-microprofile:

genie-microprofile:

api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java
```diff
@@ -104,14 +104,24 @@
...
     @Nonbinding
     Class<? extends Throwable>[] retryOn() default { Exception.class };
 
     /**
-     * 
+     *
      * @return Specify the failure to abort on. Only a type assignable from {@link java.lang.Error} or {@link java.lang.Exception} can be specified.
      * Specifying a custom java.lang.Throwable will be ignored.
+     * <p>
+     * Whenever the set of retry-on exceptions overlaps with the set of abort-on exceptions, {@code abortOn} takes precedence.
```
Emily-Jiang: retry-on -> `retryOn`, abort-on -> `abortOn`

api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java
```diff
@@ -104,14 +104,24 @@
...
+     * Overlapping means not only that the same type is present in both sets, but also that one set contains a type which
```
Emily-Jiang: I will rephrase to be:
If an exception is thrown, it first matches `{code abortOn}`. If the exception is listed in the `{code abortOn}` or can be assigned to one exception in the list, the exception will be aborted from the retry operation. 
If not, the exception will be matched against `{code retryOn}. If the exception is listed in the `{code retryOn}`  or can be assigned to one exception in the list, the exception will trigger the retry operation.

Ladicek: I did a more significant rewording. Please take another look.

Azquelt:

Azquelt:
Would it make more sense to put this information at the class level rather than on both of the parameters?

Ladicek:
I thought about that and I think that users are more likely to search for details of `retryOn` / `abortOn` in their respective documentation, rather then the general `@Retry` documentation. But we can add `@see #retryOn` or something like that.

Azquelt:

Ladicek:
Updated. This is now similar to #448: the behavior is described as a set of rules in the main annotation class javadoc, not in the `retryOn` and `abortOn` attributes javadoc. As opposed to #448, I left out the constant mentions of "custom throwables result in non-portable behavior", as that's already explained in the specification and doesn't add value here.

api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java
```diff
@@ -30,9 +30,18 @@
...
- * retry counts. Any invalid config value causes 
+ * The Retry annotation to define the number of the retries. Any invalid config value causes
  * {@link org.eclipse.microprofile.faulttolerance.exceptions.FaultToleranceDefinitionException}.
+ * <p>
+ * When a method returns and the retry policy is present, the following rules are applied:
+ * <ol>
+ * <li>If the method returns normally (doesn't throw), the result is simply returned.
+ * <li>Otherwise, if the thrown object is assignable to any value in the {@link #abortOn()} parameter, the thrown object is rethrown.
+ * <li>Otherwise, if the thrown object is assignable to any value in the {@link #retryOn()} parameter, the method call is retried
+ *     (up to configured {@link #maxRetries() number of times}).
```
Emily-Jiang: I suggest to delete the above line as it does not add much info. It could be up to configured maxRetries or maxDuration.

api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java
```diff
@@ -101,17 +110,19 @@
 
 
     /**
+     * The list of exception types which should trigger a retry.
      *
-     * @return Specify the failure to retry on. Only a type assignable from {@link java.lang.Error} or {@link java.lang.Exception} can be specified.
-     * Specifying a custom java.lang.Throwable will be ignored.
```
Emily-Jiang: Why did you delete this sentence?

Emily-Jiang: I sensed the reason when I saw your comments on the PR from @Azquelt. I advise to put the info back on as many people told me that they don't read spec and rely on JavaDoc. For some important info, we need to put on JavaDoc, which is why JDK classes have long javadocs. In Config, I have PR to add some important info from spec to JavaDoc.

api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java
Ladicek: Good point.

api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java
Ladicek: I propose to _allow_ specifying `Throwable`.

api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java
Ladicek: Done.

Azquelt: Should we extend this list to also cover `maxRetries` and `maxDuration`? Possibly in another PR.

api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java
Emily-Jiang: What do you mean @Ladicek ! I prefer the sentences are not deleted.

Ladicek: What I mean is that it doesn't make any sense to ignore `@Retry(retryOn = Throwable.class)`.

Azquelt: I agree with @Ladicek, the current doc specifies that `@Retry(retryOn = Throwable.class)` will be ignored, which is very unintuitive.

We already specify that if a method throws an object which isn't a subclass of `Error` or `Exception` then the behaviour is undefined, we don't also need to have a restriction here that states it must be _ignored_.

I think that `@Retry(retryOn = Throwable.class` should retry on `Exception`, `Error` and all their subclasses. If a throwable that doesn't extend `Exception` or `Error` is thrown, then non-portable behaviour results (as specified elsewhere).

Emily-Jiang: @Ladicek I talked with @Azquelt . The bit mentioned in his last paragraph needs to be in the JavaDoc. The current Javadoc does not convene that info, as I think this info is very important.

Ladicek: Updated.

I actually consider this a very rare corner case that _doesn't_ warrant 3 mentions in a javadoc of this annotation, but so be it.

PR #448, opened 2019-09-09
Add skipOn to CircuitBreaker

Azquelt:
Add the `skipOn` parameter to Circuit Breaker

Update the spec and javadoc to specify the meaning of `skipOn` and how `failOn` and `skipOn` interact.

We still need TCKs for this. I'll likely base them on @Ladicek's tests in #440.

For #418

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/faulttolerance/CircuitBreaker.java
+++ b/api/src/main/java/org/eclipse/microprofile/faulttolerance/CircuitBreaker.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017 Contributors to the Eclipse Foundation
+ * Copyright (c) 2017-2019 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
@@ -58,6 +58,15 @@ import javax.interceptor.InterceptorBinding;
  * </li>
  * </ul>
  * Circuit state transitions will reset the circuit breaker's records.
+ * <p>
+ * When a method returns a result, the following rules are applied to determine whether the result is a success or a failure:
+ * <ul>
+ * <li>If the method does not throw a {@link Throwable}, it is considered a success
+ * <li>Otherwise, if the thrown object is assignable to any value in the {@link #skipOn()} parameter, is is considered a success
+ * <li>Otherwise, if the thrown object is assignable to any value in the {@link #failOn()} parameter, it is considered a failure
+ * <li>Otherwise it is considered a success
+ * </ul>
+ * If a method throws a {@link Throwable} which is not an {@link Error} or {@link Exception}, non-portable behavior results.
  *
  * @author <a href="mailto:emijiang@uk.ibm.com">Emily Jiang</a>
  */
@@ -69,15 +78,28 @@ import javax.interceptor.InterceptorBinding;
 public @interface CircuitBreaker {
 
     /**
-     * Defines the failure criteria.
+     * The list of exception types which should be considered failures
      * <p>
-     * A method call will be considered a failure if it throws an exception and the type of that exception is assignable
-     * to any of the types listed in {@code failOn}.
+     * Note that if a method throws a {@link Throwable} which is not an {@link Error} or {@link Exception}, non-portable behavior
+     * results.
      *
      * @return the exception types which should be considered failures
      */
     @Nonbinding
     Class<? extends Throwable>[] failOn() default {Throwable.class};
+    
+    /**
+     * The list of exception types which should not be considered failures
+     * <p>
+     * This list takes priority over the types listed in {@link #failOn}
+     * <p>
+     * Note that if a method throws a {@link Throwable} which is not an {@link Error} or {@link Exception}, non-portable behavior results.
+     *
+     * @return the exception types which should not be considered failures
+     */
+    @Nonbinding
+    Class<? extends Throwable>[] skipOn() default {};
+
 
     /**
      * The delay after which an open circuit will transitions to half-open state.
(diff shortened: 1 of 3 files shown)


spec/src/main/asciidoc/circuitbreaker.asciidoc
```diff
@@ -38,7 +38,16 @@ succeed before the circuit can be closed. After the specified number of successf
...
+rolling failure window is created with the configured requestVolumeThreshold and failureRatio. The circuit state will only be assessed when the rolling window reaches the `requestVolumeThreshold`.
+
+When a method returns a result, the following rules are applied to determine whether the result is a success or a failure:
+
+* If the method does not throw a `Throwable`, it is considered a success
+* Otherwise, if the thrown object is assignable to any value in the `skipOn` parameter, is is considered a success
+* Otherwise, if the thrown object is assignable to any value in the `failOn` parameter, it is considered a failure
+* Otherwise it is considered a success
+
+If a method throws a `Throwable` which is not a subclass of either `Error` or `Exception`, non-portable behavior results.
```
Ladicek: I personally consider the Javadoc text to be part of the specification, so I'd say we don't have to duplicate this info in the spec text.

Emily-Jiang: @Ladicek I think for this kind of info, we need to put on JavaDoc to ensure the developers see this. The reason is that with the javadoc, IDE can display it, which is the point you need the info. Sure, someone might read the spec and they might have forgotten when they code.

Ladicek: Personally, I think that custom throwables are sufficiently rare case that we don't have to add the info into Javadoc.

On the other hand, I'd be totally fine with moving most of the spec text to the Javadocs (because, as I said, I see Javadoc as part of the spec); I just don't see the point of duplicating the info.

On a third hand :-), I don't really mind _that much_.

Emily-Jiang: I prefer the Javadoc way. I'll vote to add more text to Javadoc. It does not hurt as the Javadoc directly tells the end users what to do.

Ladicek: I don't oppose, but we should probably discuss elsewhere (separate issue?) the relationship between the spec text and the javadoc. Some people might see them as separate (spec text is canonical source of truth, while javadoc is end user documentation), while some might see them as one thing. Perhaps this should be even discussed on the MicroProfile Architecture level.

Azquelt:
I've updated the copyright year on a few files where I'd missed it and increased the package minor version because I'm adding an extra method to the annotation.

api/src/main/java/org/eclipse/microprofile/faulttolerance/CircuitBreaker.java
```diff
@@ -69,15 +78,29 @@
 public @interface CircuitBreaker {
 
     /**
-     * Defines the failure criteria.
+     * The list of exception types which should be considered failures
      * <p>
-     * A method call will be considered a failure if it throws an exception and the type of that exception is assignable
-     * to any of the types listed in {@code failOn}.
+     * Note that if a method throws a {@link Throwable} which is not a subclass of either {@link Error} or {@link Exception}, non-portable behavior
```
Ladicek: I'm stealing this sentence for the `@Retry` javadoc :-) and I found that it relies on people understanding that subclassing is reflexive. In the spec text, I mention explicitly that `Throwable` itself is custom throwable, while `Error` and `Exception` are not. Here, I don't want to get unnecessarily verbose, and I want the text to be accessible. Perhaps we can say

```
Note that if a method throws a {@link Throwable} which is not an {@link Error} or {@link Exception}, non-portable behavior results.
```

or

```
Note that if a method throws a {@link Throwable} which is not an instance of {@link Error} or {@link Exception}, non-portable behavior results.
```

(In the second case, relying on people knowing how the `instanceof` operator behaves.)

WDYT?

Azquelt: That's a good point.

Another possibility I've seen in other specs would be
```
Note that if a method throws a {@link Throwable} which is not assignable to either {@link Error} or {@link Exception}, non-portable behavior results.
```

On balance, I think I prefer using your first suggestion in the javadoc as I find it easier to understand, even if the "assignable to" language is more precise.

Ladicek: I'd be fine with using the "assignable to" variant as well.

Azquelt: Updated this to match #446

Ladicek: Perhaps GitHub is flaky, but it only shows me that it's updated on 1 place out of 3.

Azquelt: err, yeah, sorry about that, should be fixed now.

Ladicek:
I pulled your branch from GitHub to look at it locally, because I'm still seeing the "subclass of" wording. Locally, I'm looking at the commit `5d3348f3a087b6ef805c538c4774c7c5244f90c1`, and it still has the "subclass of" wording. Otherwise LGTM.

spec/src/main/asciidoc/circuitbreaker.asciidoc
Ladicek: Just wondering -- in the spec text, we already specify that throwing custom throwables is non-portable in the _Fault Tolerance Interceptor_ section. Shall we repeat it again for each FT policy?

Azquelt:
I hadn't pushed correctly :man_facepalming:

spec/src/main/asciidoc/circuitbreaker.asciidoc
Ladicek: Yea. I'll consider it a service to spec readers :-)

Ladicek: And I did that in my PR to add the retryOn/abortOn algorithm to the spec (#453), so we should be good.

PR #450, opened 2019-09-26
Make Utils.handleResults() public

kenfinnigan:
Signed-off-by: Ken Finnigan <ken@kenfinnigan.me>


Azquelt:
This change doesn't seem unreasonable. What was it needed for?

kenfinnigan:
For Quarkus the Test and this class were in different classloaders, which meant that package isn't actually the same

Azquelt:
Ah, ok

PR #453, opened 2019-10-08
specify the meaning of overlapping retryOn and abortOn

Ladicek:
This is already specified in the `@Retry` javadoc. This commit
adds the same wording to the specification text.

### DIFF
--- a/spec/src/main/asciidoc/retry.asciidoc
+++ b/spec/src/main/asciidoc/retry.asciidoc
@@ -1,7 +1,8 @@
 //
-// Copyright (c) 2016-2018 Eclipse Microprofile Contributors:
+// Copyright (c) 2016-2019 Eclipse Microprofile Contributors:
 // Emily Jiang
 // Andrew Rouse
+// Ladislav Thon
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


Ladicek:
This is a follow-up to #446 which only added this text to the javadoc. Now, it's also in the spec.

genie-microprofile:

spec/src/main/asciidoc/retry.asciidoc
```diff
@@ -39,6 +39,29 @@ If applied to a class, it means the all methods in the class will have the `@Ret
...
 If the `@Retry` policy applied on a class level and on a method level within that class, the method level `@Retry` will override the class-level `@Retry` policy for that particular method.
 
+When a method returns and the retry policy is present, the following rules are applied:
+
+* If the method returns normally (doesn't throw), the result is simply returned.
+* Otherwise, if the thrown object is assignable to any value in the `abortOn` parameter, the thrown object is rethrown.
+* Otherwise, if the thrown object is assignable to any value in the `retryOn` parameter, the method call is retried.
+* Otherwise the thrown object is rethrown.
+
+For example, to retry on all exceptions except of IO exceptions, one would write:
```
Azquelt: Grammar: replace "except of" with "except for"

Azquelt:
Just spotted something else, please update the copyright date on the file. Thanks.

spec/src/main/asciidoc/retry.asciidoc
Ladicek: This sounds weird to me, but I'll trust your English more than mine :-)

Ladicek:
Done. I also added myself to the list of contributors, but my copyright-fu is really weak, so that might be a wrong thing to do :-)

spec/src/main/asciidoc/retry.asciidoc
Azquelt: Heh, that's fair enough. For what it's worth, I've never heard "except of" used, but any of the following would be fine:
* all exceptions *except* IO exceptions
* all exceptions *except for* IO exceptions
* all exceptions *other than* IO exceptions
* all exceptions *apart from* IO exceptions
* all exceptions *with the exception of* IO exceptions

Issue #417, opened 2019-03-13
Exception-handling in Fallback


mfrembs:
Hi,

I have an issue with the @Fallback procedure. I cannot throw Exceptions even if I like to.

I implemented a MP-Rest-client with a ResponseExceptionMapper. The ResponseExceptionMapper throws "MyException" if the HTTP-status is >= 300. If the GET-request succeeds I get "MyPojo" filled with the provided information.

Now I want to make that call resilient. So I annotate the method with @Fallback. The problem I have is, that the fallback-method is getting also called when the client throws MyException. But this Exception should not be "catched" by the fallback-method. The fallback-method itself has no clue why it is called. That means I don't know if it is getting called because there was an connection-error or something like that, or it is getting called because of the "expected" MyException. Therefore I cannot rethrow MyException.

The other option is to use a FallbackHandler. Then I have to implement a handle-Method with an ExecutionContext. The ExecutionContext tells me which exception lead to calling the FallbackHandler. But I cannot rethrow the exception unless it is a RuntimeException.

Do I oversee here something? I would like to get a blacklist with exceptions when the fallback (method or handler) is not getting called. Or at least a handle-method which throws exception.

Is it possible to add this in a future version?

Thanks!

Emily-Jiang:
@def456 Thanks for the use case! Based on your description, you would like to selectively trigger fallback. I think we could add a default method like
    ```
@Nonbinding
Class<? extends Throwable>[] abortOn() default {};
```
For your use case, you can build up a blacklist.

I think blacklist might be more useful than a whitelist. Thoughts?

def456:
Thanks for your fast reply @Emily-Jiang . Yes, I also think a blacklist would be the right thing. I'm not sure about the name "abortOn()". It's a bit confusing. Will the fallback be aborted or will the fallback be called, when the underlying method is aborted by the named exceptions. I suggest "ignoreExceptions()" or something like that.

def456:
By the way: I have the same issue with the CircuitBreaker. I can only define a whitelist (failOn) but not a blacklist. Should I create an extra issue for this use case?

Emily-Jiang:
You can create an issue for discussion on CircuitBreaker as I am not sure whether it is a must to add it as yet.

def456:
I'd prefer `skipOn()`. This sounds good to me.

Do you have a roadmap until when this could be a) in the microprofile spec. and b) in the wlp / olp?

[...]

PR #457, opened 2019-10-13
#417 add fallback enhancement with whitelistand blacklist

Emily-Jiang:

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
+++ b/api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
@@ -36,6 +36,17 @@ import javax.interceptor.InterceptorBinding;
  * <li>If fallbackMethod is specified, invoke the method specified by the fallbackMethod on the same class.</li>
  * <li>If both are specified, the {@link org.eclipse.microprofile.faulttolerance.exceptions.FaultToleranceDefinitionException} must be thrown.</li>
  * </ol>
+ * <p>
+ * When a method returns and the Fallback policy is present, the following rules are applied:
+ * <ol>
+ * <li>If the method returns normally (doesn't throw), the result is simply returned.
+ * <li>Otherwise, if the thrown object is assignable to any value in the {@link #skipOn()} parameter, the thrown object will be rethrown.
+ * <li>Otherwise, if the thrown object is assignable to any value in the {@link #applyOn()} parameter, 
+ * the Fallback policy, detailed above, will be applied.
+ * <li>Otherwise the thrown object will be rethrown.
+ * </ol>
+ * If a method throws a {@link Throwable} which is not an {@link Error} or {@link Exception}, non-portable behavior results.
+
  * 
  * @author <a href="mailto:emijiang@uk.ibm.com">Emily Jiang</a>
  * 
@@ -78,4 +89,30 @@ public @interface Fallback {
     */
     @Nonbinding
     String fallbackMethod() default "";
+
+    /**
+     * The list of exception types which should trigger Fallback
+     * <p>
+     * Note that if a method throws a {@link Throwable} which is not an {@link Error} or {@link Exception}, 
+     * non-portable behavior results.
+     *
+     * @return the exception types which should trigger Fallback
+     */
+    @Nonbinding
+    Class<? extends Throwable>[] applyOn() default {Throwable.class};
+    
+    /**
+     * The list of exception types which should not be trigger Fallback
+     * <p>
+     * This list takes priority over the types listed in {@link #applyOn}
+     * <p>
+     * Note that if a method throws a {@link Throwable} which is not an {@link Error} or {@link Exception}, 
+     * non-portable behavior results.
+     *
+     * @return the exception types which should not trigger Fallback
+     */
+    @Nonbinding
+    Class<? extends Throwable>[] skipOn() default {};
+
+
 }
--- a/api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
+++ b/api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
@@ -102,7 +102,7 @@ public @interface Fallback {
     Class<? extends Throwable>[] applyOn() default {Throwable.class};
     
     /**
-     * The list of exception types which should not be trigger Fallback
+     * The list of exception types which should <i>not</i> trigger Fallback
      * <p>
      * This list takes priority over the types listed in {@link #applyOn}
      * <p>
--- a/api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
+++ b/api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017 Contributors to the Eclipse Foundation
+ * Copyright (c) 2017-2019 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
--- a/spec/src/main/asciidoc/fallback.asciidoc
+++ b/spec/src/main/asciidoc/fallback.asciidoc
@@ -94,5 +94,28 @@ When `fallbackMethod` is used a `FaultToleranceDefinitionException` will be thro
 The parameter `value` and `fallbackMethod` on `@Fallback` cannot be specified at the same time.
 Otherwise, the `FaultToleranceDefinitionException` exception will be thrown.
 
-The fallback should be triggered when an exception occurs.
-For instance, `BulkheadException`, `CircuitBreakerOpenException`, `TimeoutException` should trigger the fallback.
+The fallback might be triggered when an exception occurs (e.g. `BulkheadException`, `CircuitBreakerOpenException`, `TimeoutException`, etc), detailed below.
+When a method returns and the Fallback policy is present, the following rules are applied:
+
+* If the method returns normally (doesn't throw), the result will be simply returned.
+* Otherwise, if the thrown object is assignable to any value in the `skipOn` parameter, the thrown object will be rethrown.
+* Otherwise, if the thrown object is assignable to any value in the `applyOn` parameter, the specified fallback will be triggered.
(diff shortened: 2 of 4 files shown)


spec/src/main/asciidoc/fallback.asciidoc
```diff
@@ -94,5 +94,28 @@ When `fallbackMethod` is used a `FaultToleranceDefinitionException` will be thro
 The parameter `value` and `fallbackMethod` on `@Fallback` cannot be specified at the same time.
 Otherwise, the `FaultToleranceDefinitionException` exception will be thrown.
 
-The fallback should be triggered when an exception occurs.
-For instance, `BulkheadException`, `CircuitBreakerOpenException`, `TimeoutException` should trigger the fallback.
+The fallback might be triggered when an exception occurs (e.g. `BulkheadException`, `CircuitBreakerOpenException`, `TimeoutException`, etc), detailed below.
```
Azquelt: This hasn't changed in this pull request, but I don't like that all the examples here are Fault Tolerance exceptions, it might give the impression that fallback applies only to fault tolerance exceptions when it applies equally to any exception.

Given that you've now added more detail about how fallback works when an exception is thrown, maybe we could take these examples out. We could also add a sentence in an appropriate place to say that exceptions thrown by other fault tolerance policies can be handled by a fallback on the same method as that's what I think we were trying to hint at with these examples.

api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
```diff
@@ -36,6 +36,17 @@
  * <li>If fallbackMethod is specified, invoke the method specified by the fallbackMethod on the same class.</li>
  * <li>If both are specified, the {@link org.eclipse.microprofile.faulttolerance.exceptions.FaultToleranceDefinitionException} must be thrown.</li>
  * </ol>
+ * <p>
+ * When a method returns and the Fallback policy is present, the following rules are applied:
+ * <ol>
+ * <li>If the method returns normally (doesn't throw), the result is simply returned.
```
carlosdlr: In this part "(doesn't throw)" we can add (doesn't throw an exception)

api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
```diff
@@ -36,6 +36,17 @@
...
+ * <li>Otherwise, if the thrown object is assignable to any value in the {@link #skipOn()} parameter, the thrown object will be rethrown.
```
carlosdlr: maybe we can rewrite this to "Otherwise if the thrown exception is part of the parameter list {@link #skipOn()}, the thrown exception will not be rethrown."

api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
```diff
@@ -36,6 +36,17 @@
...
+ * <li>Otherwise, if the thrown object is assignable to any value in the {@link #applyOn()} parameter,
```
carlosdlr: maybe we can rewrite this to "Otherwise if the thrown exception is part of the parameter list {@link #applyOn()} parameter, the Fallback policy, detailed above, will be applied."

api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
Ladicek: I think the original text is clearer and more precise.

api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
Ladicek: dtto

api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
```diff
@@ -78,4 +89,30 @@ public Void handle(ExecutionContext context) {
...
+     * Note that if a method throws a {@link Throwable} which is not an {@link Error} or {@link Exception}, 
+     * non-portable behavior results.
+     *
+     * @return the exception types which should trigger Fallback
+     */
+    @Nonbinding
+    Class<? extends Throwable>[] applyOn() default {Throwable.class};
+    
+    /**
+     * The list of exception types which should not be trigger Fallback
```
Ladicek: "should not be trigger" -> "should not trigger" (or perhaps "should \<i>not\</i> trigger")

spec/src/main/asciidoc/fallback.asciidoc
```diff
@@ -94,5 +94,28 @@ When `fallbackMethod` is used a `FaultToleranceDefinitionException` will be thro
...
+When a method returns and the Fallback policy is present, the following rules are applied:
```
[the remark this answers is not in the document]
Emily-Jiang: Added a new section

spec/src/main/asciidoc/fallback.asciidoc
```diff
@@ -94,5 +94,28 @@ When `fallbackMethod` is used a `FaultToleranceDefinitionException` will be thro
...
+* If the method returns normally (doesn't throw), the result will be simply returned.
+* Otherwise, if the thrown object is assignable to any value in the `skipOn` parameter, the thrown object will be rethrown.
+* Otherwise, if the thrown object is assignable to any value in the `applyOn` parameter, the specified fallback will be triggered.
+* Otherwise the thrown object will be rethrown.
+In the following example, all exceptions assignable to `ExceptionA` and `ExceptionB`, except the ones assignable to `ExceptionBSub` will trigger the fallback operation.
+[source, java]
+----
+
+    @Retry(maxRetries = 2)
+    @Fallback(applyOn={ExceptionA.class, ExceptionB.class}, skipOn=ExceptionBSub.class, fallbackMethod= "fallbackForServiceB")
```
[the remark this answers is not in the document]
Emily-Jiang: Deleted the line. As for demonstrating the criteria, I prefer to specify the custom exception instead of relying developers knowing the hierarchy of FileNotFoundException is a subclass of IOException. For usage wise, the exceptions will be mostly user defined exception. I think the sentence above the code snippet explained the intent.

spec/src/main/asciidoc/fallback.asciidoc
Emily-Jiang: I've rephrased the sentence.

api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
Emily-Jiang: fixed

api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
Emily-Jiang: @carlosdlr the important part is the assignable and we need to clearly spell them out. Hence the long sentence :o.

api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
Emily-Jiang: See the above comments

spec/src/main/asciidoc/fallback.asciidoc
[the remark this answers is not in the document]
Ladicek: There are 2 sections under `=== Fallback usage`, called `==== Specify a FallbackHandler class` and `==== Specify the fallbackMethod`. If we make this one `==== Specify the criteria for triggering Fallback`, it looks like it's a 3rd method of specifying fallback.

I'd call it `=== Criteria for triggering fallback`, so that:

1. it's on the same level as `=== Fallback usage`;
2. the name doesn't look like it's something the user needs to do.

api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
carlosdlr: is ok

api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
Emily-Jiang: done

spec/src/main/asciidoc/fallback.asciidoc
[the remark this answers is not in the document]
Emily-Jiang: I agree with @Azquelt ! I think it looks better as it stands and still fall into the usage category as these are new parameters on the annotation of fallback.

Emily-Jiang:
I am going to merge this one now as I have resolved all issues.

api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java
```diff
@@ -78,4 +89,30 @@ public Void handle(ExecutionContext context) {
...
+    /**
+     * The list of exception types which should trigger Fallback
+     * <p>
+     * Note that if a method throws a {@link Throwable} which is not an {@link Error} or {@link Exception}, 
+     * non-portable behavior results.
+     *
+     * @return the exception types which should trigger Fallback
+     */
+    @Nonbinding
+    Class<? extends Throwable>[] applyOn() default {Throwable.class};
```
jbee: May I ask to add a sentence on what semantics an empty list has? I assume it means no white-list is used that any exception triggers the fallback.

Ladicek: Well, the default is "any exception [Throwable] triggers fallback", so presumably if the user specifies an empty list, they want no fallback. Sounds crazy/useless, but perhaps there might be some usecases for overriding a non-empty list specified in an annotation with an empty list specified in a config file.

Ladicek: But I agree this corner case should be explicitly specified for each annotation/attribute.

jbee: > But I agree this corner case should be explicitly specified for each annotation/attribute.

That is what I was after. Ideally reading the javadoc should spare people reading the spec.

PR #458, opened 2019-10-14
Configure timeouts Part I

Azquelt:
This is the change from @brunobat #403 

I've updated the commits with his new e-mail address (which does have an ECA on file) and re-merged with the current master branch to see if that passes all the checks.


Azquelt:
@brunobat I've recreated your old pull requests with the commits changed so that the author e-mail matches the e-mail address in the `Signed-off-by` line.

If you're happy with this, would you like to approve it and then we can merge it?

Azquelt:
Looking through this, I've found a couple of places where the timeout is adjustable via config but only half of the test is affected. E.g. in `CircuitBreakerRetryTest` the timeouts in the test code are now configurable, but they're supposed to match the timeouts configured in the annotations which have not been adjusted.

I don't remember seeing that before, so either I missed it when I reviewed this PR originally, or I broke it in the merge with the latest changes from master. Either way it can't be merged until those bits are fixed since it would be easy to configure the timeouts so that the test was no longer valid.

Azquelt:
I've gone through and fixed up the places where timeouts were adjusted in part of a test but not in the whole test.

tck/running_the_tck.asciidoc
```diff
@@ -115,3 +115,21 @@ Once you've added the dependency, you don't need a `tck-suite.xml` you can just
...
+|===
+|Property |Default |Unit
+
+|org.eclipse.microprofile.fault.tolerance.tck.timeout.multiplier
+|1.0
+|Dimensionless double
+|===
+
+The property can be set either with MP Config or System Properties.
+Example:
```
Ladicek: Missing example? :-) (I don't think we absolutely need one.)

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/CircuitBreakerRetryTest.java
```diff
@@ -105,7 +118,7 @@ public void testCircuitOpenWithMoreRetries() {
         catch (Exception ex) {
             // Not Expected
             invokeCounter = clientForCBWithRetry.getCounterForInvokingServiceA();
-            Assert.fail("serviceA should retry or throw a CircuitBreakerOpenException in testCircuitOpenWithMoreRetries on iteration "
+            Assert.fail("serviceA should retry or throw a CircuitthenBreakerOpenException in testCircuitOpenWithMoreRetries on iteration "
```
Ladicek: Typo, should probably be reverted to previous version.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryConditionTest.java
```diff
@@ -396,7 +397,7 @@ private void assertCompleteExceptionally(final CompletionStage<String> future,
                                              final Class<? extends Throwable> exceptionClass,
                                              final String exceptionMessage) {
         try {
-            CompletableFutureHelper.toCompletableFuture(future).get(1000, TimeUnit.MILLISECONDS);
+            CompletableFutureHelper.toCompletableFuture(future).get(TCKConfig.getConfig().getTimeoutInMillis(), TimeUnit.MILLISECONDS);
```
Ladicek: Is this call, and the one below, missing a `getTimeoutInMillis()` argument?

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryTimeoutTest.java
```diff
@@ -45,12 +49,21 @@
...
+            .setValue(RetryTimeoutClient.class,"serviceA", Timeout.class,getConfig().getTimeoutInStr(500))
+            .setValue(RetryTimeoutClient.class,"serviceWithoutRetryOn", Timeout.class,getConfig().getTimeoutInStr(500))
+            .setValue(RetryTimeoutClient.class,"serviceWithAbortOn", Timeout.class,getConfig().getTimeoutInStr(500));
+
         JavaArchive testJar = ShrinkWrap
                 .create(JavaArchive.class, "ftRetryTimeout.jar")
                 .addClasses(RetryTimeoutClient.class)
-                .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml")
+            .addAsManifestResource(config, "microprofile-config.properties")
+            .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml")
```
Ladicek: Looks like this will break formatting? :-)

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/TimeoutTest.java
```diff
@@ -48,6 +53,12 @@
 
     @Deployment
     public static WebArchive deploy() {
+        final Asset config = new ConfigAnnotationAsset()
+            .setValue(TimeoutClient.class,"serviceA", Timeout.class, getConfig().getTimeoutInStr())
```
Ladicek: Again, is this missing an argument?

tck/src/main/resources/META-INF/services/org.jboss.arquillian.core.spi.LoadableExtension
```diff
@@ -1,2 +1 @@
-org.eclipse.microprofile.fault.tolerance.tck.extension.HamcrestLoadableExtension
-org.eclipse.microprofile.fault.tolerance.tck.extension.AwaitilityLoadableExtension
+org.eclipse.microprofile.fault.tolerance.tck.extension.TckLoadableExtension
```
Ladicek: +1

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryConditionTest.java
Ladicek: I see there's an argument-less variant which defaults to 1000, but I'd lean towards being explicit.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/TimeoutTest.java
Ladicek: I see there's an argument-less variant which defaults to 1000, but I'd lean towards being explicit.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/CircuitBreakerRetryTest.java
Azquelt: Not sure how that slipped in

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryConditionTest.java
Azquelt: Yeah, I agree. I think there was the idea that most tests should use the same timeouts, but I think that's a larger, more difficult, thing to accomplish. I've removed the default timeout method.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryTimeoutTest.java
Azquelt: :anguished:

Azquelt:
Thanks for the review @Ladicek, I think I've addressed all of your comments.

Ladicek:
Looks fine to me now.

PR #459, opened 2019-10-14
Tck fixes

kenfinnigan:


kenfinnigan:
Ok, all TCK fixes should be in this PR now.

Good to review/merge, and then we can release

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/fallbackmethod/beans/FallbackMethodSubclassBeanB.java
```diff
@@ -22,11 +22,15 @@
 
 import org.eclipse.microprofile.faulttolerance.Fallback;
 
+import javax.enterprise.inject.Vetoed;
+
+// bean is vetoed to avoid accidentally picking it up and execution validation
+@Vetoed
```
Emily-Jiang: What is the consequence if this bean is picked up? I understand this bean was picked previously as there is a beans.xml. This change turns a previous bean to non-bean.

kenfinnigan: I don't recall. @manovotn might know.

All these fixes are already in master, this PR just adds them to the 2.0.x branch for a patch release

manovotn: The consequence is that if you pick this as a bean and try to run it through FT validation, then possibly you blow up on this bean instead of `FallbackMethodSubclassBeanA` which is what the test expects.

This makes sure that the test actually validates and throws an exception from the correct bean. See the discussion here for more info - https://github.com/eclipse/microprofile-fault-tolerance/pull/455#discussion_r333919938

Azquelt: This class is invalid because it has an `@Fallback` which references a non-existent fallback method, so if this class is not vetoed, it will cause a deployment failure.

The test which uses this class which is checking for a _different_ deployment failure. (It's testing that `FallbackMethodSubclassBeanA` which extends this class and provides a method named `fallback` is rejected as invalid because you cannot reference a fallback method defined in a subclass)

Emily-Jiang: ok. With the explanation from @manovotn and @Azquelt , I think this is good.

PR #460, opened 2019-10-15
IncompatibleFallbackPolicies should @Inject FallbackClientWithBothFal…

manovotn:
…lbacks.

I think the test is wrong in trying to inject `FallbackClient` but to the `@Deployment` it adds `FallbackClientWithBothFallbacks` instead. Changing it like i did in this PR gave me the correct exception.


Azquelt:
Yeah, that seems fair.

I'm not sure why the test needs to inject the bean at all, but as it's injecting a bean which doesn't exist in the application, you could get an exception for that before validating the `@Fallback` annotation.

PR #463, opened 2019-10-16
Make CircuitBreakerClientWithSyncBulkhead Dependent

Azquelt:
This bean is called asynchronously and there's no requirement for the
request scope to be active on an async thread.


Emily-Jiang:
Can you change to Depended scoped to stay the same as the previous test instead of changing the meaning of this test?

Azquelt:
@Emily-Jiang done

Emily-Jiang:
@Ladicek please review to ensure you are happy with the change.

Ladicek:
I'll check. @manovotn could you also take a look?

kenfinnigan:
I've verified SmallRye and Quarkus, so it looks good

Ladicek:
OK from Thorntail perspective too.

PR #466, opened 2019-10-25
Add CB exception hierarchy tests

### LINKED ISSUE #418: Exception-Handling in CircuitBreaker

Azquelt:
Check that `failOn` and `skipOn` are implemented correctly.

Fixes #418

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/CircuitBreakerExceptionHierarchyTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/CircuitBreakerExceptionHierarchyTest.java
@@ -1,0 +1,332 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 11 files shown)


Azquelt:
I've based these on the similar `@Retry` tests.  I haven't passed them yet as I haven't done the OpenLiberty implementation.

Azquelt:
Previously (#323, #363) we've tried to avoid re-using methods annotated with `@CircuitBreaker` and on relying on getting a new instance for each test method with `@RequestScoped`.

In this PR, I've tried to ensure that we create a new dependent instance at the beginning of each test and destroy it at the end.

I'd value feedback on whether that's sufficient as it would require a lot of repetition to use a separate method for each test case.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/CircuitBreakerExceptionHierarchyTest.java
```diff
@@ -0,0 +1,332 @@
...
+        catch (CircuitBreakerOpenException ex) {
+            return CircuitState.OPEN;
+        }
+        catch (Throwable ex) {
+            assertEquals(ex, exception);
+            return CircuitState.CLOSED;
+        }
+    }
+
+    // serviceC: @Retry(retryOn = {E1.class, E2.class}, abortOn = E0.class)
```
Ladicek: Forgot to replace retry stuff with circuit breaker stuff.

Azquelt: Well spotted, thanks. Now fixed.

Issue #271, opened 2018-05-09
Improve speed and perhaps reliability of the Bulkhead TCK tests.


hutchig:
Many of the Bulkhead TCK tests 'count' how many threads are in the bulkhead at
once. This is often done by starting them all at the same time and each method
taking a perion of time using, for example, a sleep. 

Very rarely I see a test fail due to not 'enough' parallelism, (for example in the current TCK  testBulkheadClassSemaphore3 seems to be the most common when the whole suite is run
but can be run on it's own many many times with no problem). [ACTUALLY - I DID MORE WORK ON THAT ISSUE I SAW AND BELIEVE IT WAS A PROBLEM IN OUR PROTOTYPE METRICS INTEGRATION IMPLEMENTATION BUT THIS CHANGE WOULD STILL BE AN IMPROVEMENT AS IT WOULD SPEED UP THE TESTS REF: https://github.com/OpenLiberty/open-liberty/issues/3100]

It would be better to insert some sort of join/latch.await to ensure a level of parallelism
that can be robustly checked for 100% - this will also enable the tests to run faster as well as more
reliably.

hutchig:
Having put in tighted concurrency controls, I now have a working theory that I am testing that
this may be occurring due to the Bulkhead3ClassSemaphoreBean being application
scoped and there being leftover 'workers' in the Bulkhead from previous tests.

Azquelt:
When we look at these tests, we should see if we can make it easier to understand how the tests work and what they are checking for. See #420

Emily-Jiang:
@carlosdlr will modify more tests.

Azquelt:
The Bulkhead tests take about 25% of the total runtime when running on liberty and 58% of the total runtime on smallrye. (Liberty uses a managed arquillian container so there's more overhead deploying applications).

Bulkhead doesn't have any inherent dependency on time passing (unlike Timeout, Retry and CircuitBreaker) so it's hard to justify having the tests take this long.

It's still the area of the tests that we should be targeting if we want to reduce the time it takes to run the TCK. We don't necessarily need to do any more work on this for 2.1, but there's certainly more that could be done to reduce the time that the tests take.

Azquelt:
@jbee pointed out on gitter that many of the bulkhead tests throw a lot of tasks at a bulkhead method and then check that at some point the bulkhead was filled. This can occasionally be a problem on a busy CI system, or if the runtime limits the amount of concurrent tasks in another way.

[...]

PR #469, opened 2019-11-11
Adding awaitility when to the first test refactor

carlosdlr:
Adding awaitility when to the first test refactor

issue https://github.com/eclipse/microprofile-fault-tolerance/issues/271

Signed-off-by: Carlos Andres De La Rosa <kusanagi12002@gmail.com>

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/ConfigTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/ConfigTest.java
@@ -141,7 +141,7 @@ public class ConfigTest extends Arquillian {
      * Test the configuration of maxDuration on a specific method. 
      * 
      * The serviceA is annotated with maxDuration=3000 but a configuration property overrides it with a value of 1000,
-     * so serviceA should be executed less than 11 times.
+     * so serviceA should be executed 11 or less times.
      * 
      * The test assumes that the container has been configured with the property,
      * org.eclipse.microprofile.fault.tolerance.tck.config.clientserver.ConfigClient/serviceC/Retry/maxDuration=1000
@@ -157,16 +157,17 @@ public class ConfigTest extends Arquillian {
         }
 
         //The writing service invocation takes 100ms plus a jitter of 0-200ms with the max duration of 1000ms, 
-        //the max invocation should be less than 10
+        //the max invocation should be 11 or less
+        //(minimum time for 10 invocations is 1000ms, that leaves the possibility of starting the 11th invocation just as we hit the maxDuration)
         int count = clientForConfig.getRetryCountForWritingService();
-        Assert.assertTrue(count< 11, "The max retry counter should be less than 11");
+        Assert.assertTrue(count <= 11, "The max retry counter should be 11 or less");
     }
 
     /**
      * Test the configuration of maxDuration on a class. 
      * 
      * The class is annotated with maxDuration=3000 but a configuration property overrides it with a value of 1000
-     * so serviceA should be executed less than 11 times.
+     * so serviceA should be executed 11 or less times.
      * 
      * The test assumes that the container has been configured with the property,
      * org.eclipse.microprofile.fault.tolerance.tck.config.clientserver.ConfigClassLevelMaxDurationClient/Retry/maxDuration=1000
@@ -182,8 +183,9 @@ public class ConfigTest extends Arquillian {
         }
 
         //The writing service invocation takes 100ms plus a jitter of 0-200ms with the max duration of 1000ms, 
-        //the max invocation should be less than 10
-        int retryCountforWritingService = clientForClassLevelMaxDurationConfig.getRetryCountForWritingService();        
-        Assert.assertTrue(retryCountforWritingService< 11, "The max retry counter should be less than 11");
+        //the max invocation should be 11 or less
+        //(minimum time for 10 invocations is 1000ms, that leaves the possibility of starting the 11th invocation just as we hit the maxDuration)
+        int retryCountforWritingService = clientForClassLevelMaxDurationConfig.getRetryCountForWritingService();
+        Assert.assertTrue(retryCountforWritingService <= 11, "The max retry counter should be 11 or less");
     }
 }
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryTest.java
@@ -157,9 +157,6 @@ public class RetryTest extends Arquillian {
         Assert.assertTrue(retryCountForConnectionService > 8 && retryCountForConnectionService < 100,
             "The max number of execution should be between 8 and 100 but it was " + retryCountForConnectionService +
                 ". Too many retries mean jitter is not being applied.");
-
-        Assert.assertTrue(retryClientWithNoDelayAndJitter.isDelayInRange(),
-            "The delay between each retry should be 0-400ms");
     }
 
     /**
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/TimeoutTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/TimeoutTest.java
@@ -21,23 +21,18 @@ package org.eclipse.microprofile.fault.tolerance.tck;
 
 import javax.inject.Inject;
 
-import org.eclipse.microprofile.fault.tolerance.tck.config.ConfigAnnotationAsset;
 import org.eclipse.microprofile.fault.tolerance.tck.timeout.clientserver.DefaultTimeoutClient;
 import org.eclipse.microprofile.fault.tolerance.tck.timeout.clientserver.ShorterTimeoutClient;
 import org.eclipse.microprofile.fault.tolerance.tck.timeout.clientserver.TimeoutClient;
-import org.eclipse.microprofile.faulttolerance.Timeout;
 import org.eclipse.microprofile.faulttolerance.exceptions.TimeoutException;
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.testng.Arquillian;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
-import org.jboss.shrinkwrap.api.asset.Asset;
 import org.jboss.shrinkwrap.api.asset.EmptyAsset;
 import org.jboss.shrinkwrap.api.spec.JavaArchive;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
 import org.testng.Assert;
 import org.testng.annotations.Test;
-
-import static org.eclipse.microprofile.fault.tolerance.tck.util.TCKConfig.getConfig;
 
 /**
  * Tests to exercise Fault Tolerance Timeouts.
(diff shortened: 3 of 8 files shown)


carlosdlr:
Hello @Ladicek yes that is the idea but is various classes that used this constructor in the bulkhead tests, so I wanted to do it first in this class to allow you to review the root change in an easy way I'll add the rest of classes affected today. thanks for the quick review

Ladicek:
OK, sure, makes sense.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadFutureTest.java
```diff
@@ -77,126 +83,124 @@ public void beforeTest(final ITestContext testContext) {
...
-            // but we are prepared to wait up to 100 SHORT_TIMES (10sec)
             int loops = 0;
+            final int waitingInitialCondition = loops;
+
+            await().atMost(SHORT_TIME * 2, MILLISECONDS)
+                .untilAsserted(() -> assertEquals(waitingInitialCondition, 0));
+
+            // Usually, we will not enter the loop,
+            // but we are prepared to wait up to 100 SHORT_TIMES (10sec)
             while(!result.isDone() && loops++ < 100) {
```
Ladicek: Should remove the loop and instead do a single `await()`.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/Checker.java
```diff
@@ -82,41 +85,38 @@ public Checker(int sleepMillis, TestData td, int fails) {
...
                 max = td.getMaxSimultaneousWorkers().get();
             }
+
             if (fails > 0) {
-                Thread.sleep(millis / 2);
                 fails--;
-                RuntimeErrorException e = new RuntimeErrorException(new Error("fake error for Retry Testing"));
-                Utils.log(e.toString());
+                await().atMost(millis / 2, MILLISECONDS).
+                    untilAsserted(() -> assertEquals(fails, fails - 1));
```
Ladicek: This looks wrong. When can `fails == fails - 1`?

Ladicek:
Two comments, otherwise LGTM.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadFutureTest.java
```diff
@@ -77,126 +83,121 @@ public void beforeTest(final ITestContext testContext) {
...
+     * @param result future to simulate the waiting time
+     */
+    private void waitingSimulation(Future<String> result) {
         try {
-            Thread.sleep(SHORT_TIME + SHORT_TIME);
-            // Usually, we will not enter the loop, 
+            final int waitingInitialCondition = 0;
+
+            await().atMost(SHORT_TIME * 2, MILLISECONDS)
+                .untilAsserted(() -> assertEquals(waitingInitialCondition, 0));
```
Ladicek: Actually, why do we even do this? We set a variable to 0 and then we wait up to 200 millis, until the variable is 0?

Ladicek: So, we replaced this

```java
        Assert.assertFalse(result.isDone(), "Future reporting Done when not");
        try {
            Thread.sleep(SHORT_TIME + SHORT_TIME);
            // Usually, we will not enter the loop, 
            // but we are prepared to wait up to 100 SHORT_TIMES (10sec)
            int loops = 0;
            while(!result.isDone() && loops++ < 100) {
                Thread.sleep(SHORT_TIME);
            }
        }
        catch (Throwable t) {
            Assert.assertNull(t);
        }
        Assert.assertTrue(result.isDone(), "Future done not reporting true");
```

with this

```java
        assertFalse(result.isDone(), "Future reporting Done when not");
        waitingSimulation(result);
        assertTrue(result.isDone(), "Future done not reporting true");
```

I think the replacement should be

```
        assertFalse(result.isDone());
        await().atMost(100 * SHORT_TIME, MILLISECONDS).untilAsserted(() -> assertTrue(result.isDone());
```

It's short and conveys intent well, and doesn't need a `waitingSimulation` method (which still contains nonsensical code like `catch (Throwable t) { assertNull(t); }`).

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/Checker.java
Ladicek: I don't think this was properly addressed. `await().atMost(TIME).untilAsserted(something that is never true)` is IMHO a very poor replacement for `Thread.sleep(TIME)`.

Ladicek:
Sorry for being so stubborn, I think this is generally a move in the right direction, but some pieces are still a bit suboptimal.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadFutureTest.java
carlosdlr: yes for me make sense and with that, we can remove the waitingSimulation method and make this more readable, I'll make the change and run the tests again thank you @Ladicek

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/Checker.java
carlosdlr: actually here is all the opposite always is true because and putting this after decreasing the counter so we are ensuring that always will wait, but if you think that is so poor what do you suggest? thank you

Ladicek: I think `assertEquals(fails, fails - 1)` can never be true (unless `fails` is a floating point number, then there are options :-) ).

Frankly, if we need to wait for certain amount of time, then `Thread.sleep` is best. `await().atMost(TIME).untilAsserted(() -> Assert.fail())` is exactly the same as `Thread.sleep(TIME)` (modulo interruptions), but it's IMHO much harder to read.

carlosdlr: well, I don't understand why you say that is not true the condition, the fail variable is an integer and after decreasing the value `fails --` I'm checking if the variable was decreased by one so always will wait `millis / 2` or maybe I can change the assertion for an `asserTrue` to make it clear. if is not enough for you I can let the Thread wait like was before.

carlosdlr:
yes is not a problem always we can learn something new

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/Checker.java
Ladicek: But you're doing `assertEquals(fails, fails - 1)`. That can't possibly be true. You'd have to store the old value into a different variable to be able to verify that `fails--` really decrements by 1 -- but I don't think such assert would make a lot of sense, we're not in the business of testing how Java operators work :-)

carlosdlr: ok i'll revert the change here. thank you

Ladicek: Thanks!

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadFutureTest.java
```diff
@@ -77,126 +83,121 @@ public void beforeTest(final ITestContext testContext) {
...
         }
 
-        Assert.assertFalse(result.isDone(), "Future reporting Done when not");
+        assertFalse(result.isDone(), "Future reporting Done when not");
         try {
             String r;
-            Assert.assertTrue("GET.".equals(r = result.get()), r);
-            Assert.assertTrue("GET.GET_TO.".equals(r = result.get(1, TimeUnit.SECONDS)), r);
+            assertEquals(r = result.get(), "GET.", r);
+            assertEquals(r = result.get(1, TimeUnit.SECONDS), "GET.GET_TO.", r);
```
Azquelt: I don't like this `assertEquals(r = result.get(), "GET", r)` structure.

Can we change it to `assertEquals(result.get, "GET");` and get rid of `r`?

I'd hope that `assertEquals` gives a sensible error message like `Expected "GET" but was "something else"` so passing the result as the message shouldn't be necessary.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/Checker.java
```diff
@@ -82,41 +85,38 @@ public Checker(int sleepMillis, TestData td, int fails) {
...
+            await().atMost(millis, MILLISECONDS).untilAsserted(() -> assertTrue(fails < 0));
 
+            log("Task " + taskId + " woke.");
             // We will countDown the latch in the finally block
-
         }
-        catch (InterruptedException e) {
-            Utils.log(e.toString());
+        catch (RuntimeException e) {
+            log(e.toString());
```
Azquelt: Is it possible to remove this catch block so that an exception here fails the test?

In general an unexpected exception should fail the test, not be caught and logs. No-one looks at test logs unless the test fails.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/Checker.java
carlosdlr: done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadFutureTest.java
carlosdlr: done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadFutureTest.java
carlosdlr: done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/Checker.java
carlosdlr: This try-catch is not possible to remove at this moment due this method has 2 different flows base on how many failed threads the test wants to simulate. I think this can be improved later after review all the other tests that use this method. This is just the first refactoring.

Ladicek:
This PR now contains commits that shouldn't be there?

carlosdlr:
@Emily-Jiang @Ladicek @Azquelt, please refer to this new PR [Adding awaitility to Bulkhead Tests First refactor #478](https://github.com/eclipse/microprofile-fault-tolerance/pull/478), due to some issues with my fork all the review fixes that were here are there. Please approve if it is ok for you. Thank you

PR #471, opened 2019-11-12
relax the requirements on Future implementations

### LINKED ISSUE #425: Implementation of Future should be transparent
Some tests (e.g. `org.eclipse.microprofile.fault.tolerance.tck.bulkhead.BulkheadFutureTest.testBulkheadClassAsynchFutureDoneAfterGet`) make the assumption that a `Future` returned from a method with FT behaviour does delegate to the original future instance returned by the wrapped method which is observable by `get()` and `get(long timeout, TimeUnit unit)` actually returning different values.

Delegation as solution is also asked for in the spec:

> If the execution completes normally and returns a value, the Future or CompletionStage will then delegate to the returned value.

This is only one way to implement it but since the result of `get` with or without timeout should be consistent any other solution that behaves correctly should be valid. After all this is the point of using abstractions.

Ladicek:

Resolves #425.

Supersedes #426.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadFutureTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadFutureTest.java
@@ -96,8 +96,8 @@ public class BulkheadFutureTest extends Arquillian {
 
         assertFalse(result.isDone(), "Future reporting Done when not");
         try {
-            assertEquals(result.get(), "GET.");
-            assertEquals(result.get(1, TimeUnit.SECONDS), "GET.GET_TO.");
+            assertEquals(result.get(), "RESULT");
+            assertEquals(result.get(1, TimeUnit.SECONDS), "RESULT");
         }
         catch (Throwable t) {
             fail("Unexpected exception", t);
@@ -143,8 +143,8 @@ public class BulkheadFutureTest extends Arquillian {
 
         assertFalse(result.isDone(), "Future reporting Done when not");
         try {
-            assertEquals(result.get(1, TimeUnit.SECONDS), "GET_TO.");
-            assertEquals(result.get(), "GET_TO.GET.");
+            assertEquals(result.get(1, TimeUnit.SECONDS), "RESULT");
+            assertEquals(result.get(), "RESULT");
 
         }
         catch (Throwable t) {
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/FutureChecker.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/FutureChecker.java
@@ -19,17 +19,15 @@
  *******************************************************************************/
 package org.eclipse.microprofile.fault.tolerance.tck.bulkhead.clientserver;
 
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.eclipse.microprofile.fault.tolerance.tck.bulkhead.Utils;
 
 /**
  * A backend that is used in the tests of the operations on the Future returned
  * by the user.
- * 
+ *
  * @author Gordon Hutchison
  */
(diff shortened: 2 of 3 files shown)


Ladicek:
We discussed this on the call today and decided to keep the text a bit more vague. Instead of defining how results from the original future are propagated to the resulting one, we'll just say that the futures are "behaviorally equivalent", or something like that.

Ladicek:
Pushed a change per above.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/FutureChecker.java
```diff
@@ -79,7 +78,7 @@ public FutureChecker(int sleepMillis) {
     public final class TestFuture implements Future<String> {
 
         private AtomicBoolean done = new AtomicBoolean(false);
-        private String result = "";
+        private String result = "RESULT";
```
Azquelt: This could be `static final` as it's now a constant.

Ladicek: Good point, will fix!

Ladicek:
> I wouldn't be sure what "behaviourally equivalent" meant

Well, exactly!

I could possibly add a `[NOTE]` explaining what it means for two futures A and B to be behaviorally equivalent, but it would contain the same text that was in this PR originally :-)

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/FutureChecker.java
Ladicek: I just inlined the constant into the two `get` methods.

Ladicek:
Hi @jbee, we'd be interested in your feedback on this PR :-)

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadFutureTest.java
```diff
@@ -93,11 +93,11 @@ public void testBulkheadMethodAsynchFutureDoneAfterGet() {
         Assert.assertFalse(result.isDone(), "Future reporting Done when not");
         try {
             String r;
-            Assert.assertTrue("GET.".equals(r = result.get()), r);
-            Assert.assertTrue("GET.GET_TO.".equals(r = result.get(1, TimeUnit.SECONDS)), r);
+            Assert.assertTrue("RESULT".equals(r = result.get()), r);
+            Assert.assertTrue("RESULT".equals(r = result.get(1, TimeUnit.SECONDS)), r);
```
Ladicek: I think this (and perhaps other parts of this PR?) will conflict with #469. I'm fine with letting #469 in and then rebasing.

Ladicek:
@michalszynkiewicz came with a very good question: if the user-defined method declares `CompletionStage` as a return type and actually returns a `CompletableFuture` (or any other `CompletionStage` that supports `.toCompletableFuture`), does the "behaviorally equivalent" `CompletionStage` also support `.toCompletableFuture`?

Ladicek:
I'd be inclined to say yes, but I'm interested in everyone's opinion as well.

jbee:
I'd say the method signature is the factor to consider. If it returns a `CompletionStage` this is what the caller can assume, nothing more. The idea of giving importance to the way something is implemented somewhat breaks the idea of decoupling though the use of interfaces. If the user wants a `CompletableFuture` the method should return it.

Ladicek:
If the method has a return type of `CompletableFuture`, it is not a valid `@Asynchronous` method declaration per my understanding of the spec.

Azquelt:
Correct, which we did because the implementation will need to proxy whatever type is returned from the `@Asynchronous` method.

jbee:
I got this completely wrong. Had not thought about the connection to the spec and `@Asynchronous` and that `CompletionStage` does include `toCompletableFuture` with clear indication that it might throw `UnsupportedOperationException`. Considering this now I'd even go so far to say that a `CompletionStage` returned by FT processing should always return a class that does support the method. It should not be a big deal to do so and it gives the users something they can rely on without looking into how something might be implemented.

Ladicek:
Rebased.

PR #472, opened 2019-11-12
Update the CircuitBreaker spec with more examples

Azquelt:
I've moved around the text a bit so that the examples are next to the
part of the spec that they're demonstrating.

### DIFF
--- a/spec/src/main/asciidoc/circuitbreaker.asciidoc
+++ b/spec/src/main/asciidoc/circuitbreaker.asciidoc
@@ -23,6 +23,7 @@
 == Circuit Breaker
 
 A Circuit Breaker prevents repeated failures, so that dysfunctional services or APIs fail fast.
+
 There are three circuit states:
 
 * Closed: In normal operation, the circuit is closed. If a failure occurs, the Circuit Breaker records the event. In closed 
@@ -40,15 +41,8 @@ before the successThreshold is reached the circuit will transition to open.
 Note that circuit state transitions will reset the Circuit Breaker's records. For example, when the circuit transitions to closed a new
 rolling failure window is created with the configured requestVolumeThreshold and failureRatio. The circuit state will only be assessed when the rolling window reaches the `requestVolumeThreshold`.
 
-When a method returns a result, the following rules are applied to determine whether the result is a success or a failure:
+The following example and scenarios demonstrate when the circuit opens.
 
-* If the method does not throw a `Throwable`, it is considered a success
-* Otherwise, if the thrown object is assignable to any value in the `skipOn` parameter, is is considered a success
-* Otherwise, if the thrown object is assignable to any value in the `failOn` parameter, it is considered a failure
-* Otherwise it is considered a success
-
-If a method throws a `Throwable` which is not a subclass of either `Error` or `Exception`, non-portable behavior results.
- 
 [source, java]
 ----
 @CircuitBreaker(successThreshold = 10, requestVolumeThreshold = 4, failureRatio=0.5, delay = 1000)
@@ -59,18 +53,8 @@ public Connection serviceA() {
    return conn;
 }
 ----
-Below are a couple scenarios to demonstrate when the circuit opens for the above code snippet.
 
 * Scenario 1
-** Request 1 - success
-** Request 2 - failure
-** Request 3 - failure
-** Request 4 - success
-** Request 5 - CircuitBreakerOpenException
-
-The Request 4 will trigger the circuit to open. The Request 5 will hit the exception.
-
-* Scenario 2
 ** Request 1 - success
 ** Request 2 - failure
 ** Request 3 - success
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/circuitbreaker.asciidoc
```diff
@@ -59,26 +53,46 @@ public Connection serviceA() {
...
 ** Request 1 - success
 ** Request 2 - failure
-** Request 3 - success
+** Request 3 - failure
 ** Request 4 - success
-** Request 5 - failure
-** Request 6 - CircuitBreakerOpenException 
+** Request 5 - CircuitBreakerOpenException
+
+In this scenario, request 4 will cause the circuit to open. Request 5 will hit the `CircuitBreakerOpenException`.
```
Ladicek: It seems you just swapped the 2 scenarios, do I get that correctly?

spec/src/main/asciidoc/circuitbreaker.asciidoc
```diff
@@ -59,26 +53,46 @@ public Connection serviceA() {
...
+Note that request 3 does not cause the circuit to open because the rolling window has not yet reached the `requestVolumeThreshold`.
+
+When a method returns a result, the following rules are applied to determine whether the result is a success or a failure:
```
Ladicek: I don't have a solution, but the flow of the text is strange here. We have a section describing when and how the circuit opens, and then without any delimitation, we have "When a method returns a result"... I don't know how to say it, perhaps just ignore me :-)

spec/src/main/asciidoc/circuitbreaker.asciidoc
Azquelt: Yeah, I did (didn't realise the diff would look so odd). I added some further explanation to what was happening in each scenario and I realised that scenario 2 was the "normal" one where the rolling window is full when the failure ratio is reached and scenario 1 was the less usual one where the failure ratio is reached before the rolling window is full but the circuit doesn't open until the window is filled. I swapped them to put the more common and easier to understand scenario first.

spec/src/main/asciidoc/circuitbreaker.asciidoc
Azquelt: Yeah, I agree. I was trying to stick to the same structure as the other annotations but I'm not sure it really works. I'll try adding some headings and see if that makes it flow better.

spec/src/main/asciidoc/circuitbreaker.asciidoc
Ladicek: Agree, it just took me a while to understand the diff, so I figured I'd rather ask :-)

spec/src/main/asciidoc/circuitbreaker.asciidoc
Ladicek: Headings is also the only possible solution that I could come up with.

Azquelt:
I've done some more re-arranging and added headings. When I put things in what I thought was a more logical order, I found there was now quite a bit of repetition. I ended up removing most of what was previously under the *Circuit Breaker usage* heading since it was better covered by the new examples.

spec/src/main/asciidoc/circuitbreaker.asciidoc
```diff
@@ -22,33 +22,38 @@
...
-A Circuit Breaker prevents repeated failures, so that dysfunctional services or APIs fail fast.
+A Circuit Breaker prevents repeated failures, so that dysfunctional services or APIs fail fast. If a service is failing frequently, the circuit breaker opens and no more calls to that service are attempted until a period of time has passed.
+
 There are three circuit states:
 
-* Closed: In normal operation, the circuit is closed. If a failure occurs, the Circuit Breaker records the event. In closed 
-state the requestVolumeThreshold and failureRatio parameters may be configured in order to specify the conditions under which the breaker
-will transition the circuit to open. If the failure conditions are met, the circuit will be opened.
+* *Closed*: In normal operation, the circuit breaker is closed. The circuit breaker records whether each call is a success or failure and
+keep track of the most recent results in a rolling window. Once the rolling window is full, if the proportion of failures in the rolling window
```
michalszynkiewicz: typo `keep` -> `keeps`

Azquelt: Thanks!

spec/src/main/asciidoc/circuitbreaker.asciidoc
```diff
@@ -59,48 +64,53 @@ public Connection serviceA() {
...
+When a method returns a result, the following rules are applied to determine whether the result is a success or a failure:
+
+* If the method does not throw a `Throwable`, it is considered a success
+* Otherwise, if the thrown object is assignable to any value in the `skipOn` parameter, is is considered a success
+* Otherwise, if the thrown object is assignable to any value in the `failOn` parameter, it is considered a failure
+* Otherwise it is considered a success
+
+If a method throws a `Throwable` which is not a subclass of either `Error` or `Exception`, non-portable behavior results.
+
+In the following example, all exceptions assignable to `ExceptionA` and `ExceptionB`, except the ones assignable to `ExeptionBSub` will be considered failures. `ExceptionBSub` and all other exceptions will be considered successes.
```
Emily-Jiang: I would not add the sentence`ExceptionBSub` and all other exceptions will be considered successes.`, because they don't contribute towards circuitbreaker not they are not treated as success.

Azquelt: Hmm, I put that because, as I understand it, the method throwing an exception which is not considered a failure should be treated the same as the method returning without throwing an exception.

I.e.:
* In closed state, a success is recorded in the rolling window
* In half-open state, the trial execution is considered to have "succeeded" and will allow the circuit breaker to close

I thought this behaviour might be considered unintuitive (why should an exception be considered a success) so I thought it would be helpful to call it out explicitly.

Azquelt: If that's not the case and exceptions matching `skipOn` should be ignored completely then I would want to revise the language in the spec and check that we have TCKs to assert that.

Currently the TCK only asserts that an exception matching `skipOn` doesn't cause the breaker to open, but it doesn't check whether it's recorded as a success in the rolling window or whether it will cause the breaker to close when it's in half-open state.

Emily-Jiang: `SkipOn` does not change the method exception. It only means the exception is not counted towards the circuit breaker. Similarly, for `Retry`, the `abortOn` means the exception will not trigger `Retry`. The fact of the exception listed in `abortOn` should not cause the Exception to be swallowed. +1 on adding more tests to mandate this.

Azquelt: Existing tests do assert that an exception listed in `skipOn` is not swallowed.

I'm still not 100% clear on the desired behaviour when an exception listed in `skipOn` is thrown in terms of:
 * what is stored in the rolling window
 * what should happen if a trial execution ends with an exception listed in `skipOn` being thrown

My assumption was that all of the following cases would be treated as a "success" by the rolling window and for a trial execution:
 * method returns normally
 * method throws exception not listed in `failOn`
 * method throws exception listed in `skipOn`

Only in one case is the execution considered a "failure" by the rolling window and for a trial execution:
 * method throws exception listed in `failOn` and not listed in `skipOn`

I'll continue with this assumption unless someone wants to argue it.

I'll try to update the text to avoid giving the impression that an exception is ever swallowed by the circuit breaker.

Azquelt:
@Emily-Jiang I've updated the language slightly with the latest commit. Could you check whether you're happy with it?

spec/src/main/asciidoc/circuitbreaker.asciidoc
```diff
@@ -22,33 +22,38 @@
...
+==== Configuring when the circuit opens and closes
+
+The following parameters control when the circuit breaker opens and closes.
+
+* `requestVolumeThreshold` controls the size of the rolling window used when the circuit breaker is closed
+* `failureRatio` controls the proportion of failures within the rolling window which will cause the circuit breaker to open
+* `successThreshold` controls the number of trial calls which are allowed when the circuit breaker is half-open
+* `delay` and `delayUnit` control how long the circuit breaker stays open
+
+Circuit breaker state transitions will reset the Circuit Breaker's records. For example, when the circuit breaker transitions to closed a new
```
Emily-Jiang: closed to `closed` for a better reading

PR #478, opened 2019-11-26
Adding awaitility to Bulkhead Tests First refactor

carlosdlr:
Adding awaitility when to the first test refactor

issue #271

Signed-off-by: Carlos Andres De La Rosa kusanagi12002@gmail.com

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadAsynchRetryTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadAsynchRetryTest.java
@@ -53,18 +53,24 @@ import org.eclipse.microprofile.fault.tolerance.tck.bulkhead.clientserver.TestDa
 import org.eclipse.microprofile.fault.tolerance.tck.util.Packages;
 import org.eclipse.microprofile.fault.tolerance.tck.util.TestException;
 import org.eclipse.microprofile.faulttolerance.exceptions.BulkheadException;
-import org.hamcrest.MatcherAssert;
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.testng.Arquillian;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
 import org.jboss.shrinkwrap.api.asset.EmptyAsset;
 import org.jboss.shrinkwrap.api.spec.JavaArchive;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
-import org.testng.Assert;
 import org.testng.ITestContext;
 import org.testng.annotations.AfterMethod;
 import org.testng.annotations.BeforeTest;
 import org.testng.annotations.Test;
+
+import static org.eclipse.microprofile.fault.tolerance.tck.bulkhead.Utils.log;
+import static org.eclipse.microprofile.fault.tolerance.tck.bulkhead.Utils.handleResults;
+import static org.eclipse.microprofile.fault.tolerance.tck.bulkhead.Utils.loop;
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.fail;
 
 /**
  * This collection of tests tests that failures, particularly Asynchronous
@@ -72,6 +78,7 @@ import org.testng.annotations.Test;
  * work correctly.
  *
  * @author Gordon Hutchison
+ * @author carlosdlr
  *
  */
 public class BulkheadAsynchRetryTest extends Arquillian {
@@ -96,24 +103,24 @@ public class BulkheadAsynchRetryTest extends Arquillian {
     private Bulkhead55RapidRetry10ClassAsynchBean rrClassBean;
     @Inject
     private Bulkhead55RapidRetry10MethodAsynchBean rrMethodBean;
-    
+
     @Inject
     private BulkheadRetryDelayAsyncBean retryDelayAsyncBean;
-    
+
     @Inject
     private BulkheadRetryQueueAsyncBean retryQueueAsyncBean;
-    
+
     @Inject
     private BulkheadRetryAbortOnAsyncBean retryAbortOnAsyncBean;
-    
+
     private List<AsyncBulkheadTask> tasks = new ArrayList<>();
-    
+
     private AsyncBulkheadTask newTask() {
         AsyncBulkheadTask task = new AsyncBulkheadTask();
         tasks.add(task);
         return task;
     }
-    
+
     @AfterMethod
     public void cleanupTasks() {
         for (AsyncBulkheadTask task : tasks) {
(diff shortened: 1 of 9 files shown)


tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/Checker.java
```diff
@@ -82,41 +85,36 @@ public Checker(int sleepMillis, TestData td, int fails) {
...
 
-            Utils.log("Task " + taskId + " sleeping for " + millis + " milliseconds. " + now
-                    + " workers inside Bulkhead from " + td.getInstances() + " instances " + BAR.substring(0, now));
-
-            Thread.sleep(millis);
+            log("Task " + taskId + " sleeping for " + millis + " milliseconds. " + now
+                + " workers inside Bulkhead from " + td.getInstances() + " instances " + BAR.substring(0, now));
 
-            Utils.log("Task " + taskId + " woke.");
+            await().atMost(millis, MILLISECONDS).untilAsserted(() -> assertTrue(fails < 0));
```
Ladicek: This is weird. The `fails--` statement above is the only place where `fails` is modified. So at this point, either `fails < 0` already, or it isn't and it doesn't make sense to wait for it. Or am I missing something?

Ladicek:
One comment, otherwise LGTM.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadFutureTest.java
```diff
@@ -77,126 +82,92 @@ public void beforeTest(final ITestContext testContext) {
...
+    @Test
     public void testBulkheadMethodAsynchFutureDoneWithoutGet() {
-
         Checker fc = new FutureChecker(SHORT_TIME);
-        Future<String> result = null;
         try {
-            result = bhBeanMethodAsynchronousDefault.test(fc);
+            final Future<String> result = bhBeanMethodAsynchronousDefault.test(fc);
+            assertFalse(result.isDone(), "Future reporting Done when not");
+            await().atMost(SHORT_TIME * SHORT_TIME, MILLISECONDS).untilAsserted(()-> assertTrue(result.isDone()));
```
Azquelt: You've changed an effective wait of `SHORT_TIME * 100` to `SHORT_TIME * SHORT_TIME`

`SHORT_TIME` is actually 100, but it looks weird in the code to just be squaring the timeout time.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadFutureTest.java
```diff
@@ -77,126 +82,92 @@ public void beforeTest(final ITestContext testContext) {
...
+    @Test
     public void testBulkheadClassAsynchFutureDoneWithoutGet() {
-
         Checker fc = new FutureChecker(SHORT_TIME);
-        Future<String> result = null;
         try {
-            result = bhBeanMethodAsynchronousDefault.test(fc);
+            final Future<String> result = bhBeanClassAsynchronousDefault.test(fc);
+            assertFalse(result.isDone(), "Future reporting Done when not");
+            await().atMost(SHORT_TIME * SHORT_TIME, MILLISECONDS).untilAsserted(()-> assertTrue(result.isDone()));
```
Azquelt: `SHORT_TIME * SHORT_TIME` again, see above.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/Checker.java
Azquelt: It looks to me like the original code was sleeping to simulate some work, rather than sleeping to wait for something to happen.

If that's the case, then we'd just want to sleep rather than using awaitility.

If that's not the case, we should change the log line above which says we're going to sleep.

Ladicek: Yea. We definitely shouldn't blindly replace all `Thread.sleep` -- if the code sleeps to simulate work, rewriting it to Awaitility is a bad idea. (It's possible, as shown here, but it will only obscure the meaning of the code and provide no benefit at all.)

Azquelt:
One more thing (sorry to be a pain) can you update the copyright date on each of the files you've updated.

Any files created before 2019 should have
```
 * Copyright (c) (creation year)-2019 Contributors to the Eclipse Foundation
```

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadFutureTest.java
carlosdlr: done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadFutureTest.java
carlosdlr: done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/Checker.java
carlosdlr: I put it back in the sleep part. done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadFutureTest.java
carlosdlr: done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/Checker.java
carlosdlr: done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadFutureTest.java
carlosdlr: done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/Checker.java
```diff
@@ -82,41 +82,40 @@ public Checker(int sleepMillis, TestData td, int fails) {
...
-                    + " workers inside Bulkhead from " + td.getInstances() + " instances " + BAR.substring(0, now));
+            log("Task " + taskId + " sleeping for " + millis + " milliseconds. " + now
+                + " workers inside Bulkhead from " + td.getInstances() + " instances " + BAR.substring(0, now));
 
             Thread.sleep(millis);
 
-            Utils.log("Task " + taskId + " woke.");
+            if (td.getLatch() != null) {
+                td.getLatch().countDown();
+            }
```
Ladicek: Why does this duplicate the `finally` block?

Ladicek:
Just one comment, otherwise LGTM.

One question: @Azquelt didn't you recently add a change to make all timeouts adjustable (by using a multiplier read from a system property)? This PR has code that still uses hardcoded timeouts. Do we want to do anything about that?

Azquelt:
@Ladicek I did, but I don't think I actually changed the bulkhead tests as they weren't causing us problems (my biggest issue has been that they take _too_ long).

It would be a good idea to make these timeouts configurable (and I think that should be fairly easy) but we can do that as a follow up after this PR is merged.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadAsynchRetryTest.java
```diff
@@ -1,6 +1,6 @@
 /*
  *******************************************************************************
- * Copyright (c) 2017-2018 Contributors to the Eclipse Foundation
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
```
Azquelt: 2017-2019

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadAsynchTest.java
```diff
@@ -1,6 +1,6 @@
 /*
  *******************************************************************************
- * Copyright (c) 2017-2018 Contributors to the Eclipse Foundation
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
```
Azquelt: 2017-2019

Ladicek:
Ah, I see. Thanks.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/clientserver/Checker.java
carlosdlr: good catch was a mistake thank you. done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadAsynchRetryTest.java
carlosdlr: done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadAsynchTest.java
carlosdlr: done

carlosdlr:
@Emily-Jiang if you don't have any comment, can I merge this PR? Thank you

Issue #487, opened 2019-12-05
Jenkins build failed with gpg failure


Emily-Jiang:
[INFO] [INFO] --- maven-gpg-plugin:1.6:sign (sign-artifacts) @ microprofile-fault-tolerance-parent ---
[INFO] gpg: signing failed: Not a tty
[INFO] gpg: signing failed: Not a tty
[INFO] [INFO] ------------------------------------------------------------------------
[INFO] [INFO] Reactor Summary for MicroProfile Fault Tolerance 2.0.3-RC1:
[INFO] [INFO] 
[INFO] [INFO] MicroProfile Fault Tolerance ....................... FAILURE [ 56.725 s]
[INFO] [INFO] microProfile-fault-tolerance-api ................... SKIPPED
[INFO] [INFO] microprofile-fault-tolerance-tck ................... SKIPPED
[INFO] [INFO] microprofile-fault-tolerance-spec .................. SKIPPED
[INFO] [INFO] ------------------------------------------------------------------------
[INFO] [INFO] BUILD FAILURE
[INFO] [INFO] ------------------------------------------------------------------------
[INFO] [INFO] Total time:  01:11 min
[INFO] [INFO] Finished at: 2019-12-04T11:20:30Z
[INFO] [INFO] ------------------------------------------------------------------------
[INFO] [ERROR] Failed to execute goal org.apache.maven.plugins:maven-gpg-plugin:1.6:sign (sign-artifacts) on project microprofile-fault-tolerance-parent: Exit code: 2 -> [Help 1]
[INFO] [ERROR] 
[INFO] [ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[INFO] [ERROR] Re-run Maven using the -X switch to enable full debug logging.

Emily-Jiang:
Need to port the change to master branch.

PR #489, opened 2019-12-05
#487 - fix Jenkins build

Emily-Jiang:


Azquelt:
This looks like the same fix that was needed in eclipse/microprofile-metrics#485

Issue #416, opened 2019-03-06
RetryConditionTest uses unmanaged threads


Azquelt:
In several places, `AsyncRetryClient` calls `CompletableFuture.supplyAsync()` without providing an executor, resulting in a task which runs on the common fork join pool.

In a JavaEE environment, this is not valid.

This problem affects the following test methods in `RetryConditionTest`:
* `testRetryChainSuccess`
* `testRetryChainExceptionally`
* `testRetryParallelExceptionally`
* `testRetryParallelSuccess`

Azquelt:
Currently, some tests which need to run things in parallel use the `org.eclipse.microprofile.fault.tolerance.tck.util.AsyncCaller`, which works similarly to an `Executor` but uses the `@Asynchronous` annotation.

PR #490, opened 2019-12-11
#416 - executor for tests

brunobat:
Signed-off-by: Bruno <brunobat@gmail.com>

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/util/AsyncCallerExecutor.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/util/AsyncCallerExecutor.java
@@ -20,11 +20,13 @@
 package org.eclipse.microprofile.fault.tolerance.tck.util;
 
 import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
 import java.util.concurrent.Executor;
 
 @ApplicationScoped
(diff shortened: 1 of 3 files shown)


tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/util/AsyncCallerExecutor.java
```diff
@@ -0,0 +1,34 @@
...
+ *******************************************************************************/
+package org.eclipse.microprofile.fault.tolerance.tck.util;
+
+import javax.enterprise.context.ApplicationScoped;
+import java.util.concurrent.Executor;
+
+@ApplicationScoped
+public class AsyncCallerExecutor implements Executor {
+
+    private AsyncCaller asyncCaller;
```
Ladicek: How is this field initialized? Am I missing something?

brunobat: Nop... you are right. Missing the annotation. 
Which rises the curious question of how the implementation container was successfully passing the test. Will dig tomorrow

brunobat: yup. seems ok now

PR #493, opened 2020-01-09
Specify behaviour of Async Bulkhead with CompletionStage

### LINKED ISSUE #484: Behavior of `Bulkhead` with `CompletionStage` is not tested and seems underspecified
I am working on MP FT reimplementation and noticed there are no tests that test `@Asynchronous` `CompletionStage` bulkhead.

Also, should an async `CompletionStage` invocation leave bulkhead when the `CompletionStage` is done? This does not seem to be said explicitly in the spec. OTOH seems a natural consequence of what is in the spec.

Azquelt:
Update the spec to make clear that the Bulkhead doesn't consider the method to be complete until the returned completion stage completes.

Add TCK tests to assert this.

Fixes #484

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadAsynchTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadAsynchTest.java
@@ -1,6 +1,6 @@
 /*
  *******************************************************************************
- * Copyright (c) 2017-2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2017-2020 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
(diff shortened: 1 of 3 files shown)


tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadAsynchTest.java
```diff
@@ -255,6 +267,41 @@ public void testBulkheadExceptionThrownWhenQueueFullAsync() throws InterruptedEx
...
+            // Because futures 1 & 2 are still not "complete" (even though the method call may have returned), future 5 should be rejected
+            expectBulkheadException(future5);
+            
+            // Complete the CompletionStage which was returned
+            result.complete(null);
+            
+            // After the CompletionStage completes, future 1 & 2 should complete
+            future1.get(getConfig().getTimeoutInMillis(1000), TimeUnit.MILLISECONDS);
+            future2.get(getConfig().getTimeoutInMillis(1000), TimeUnit.MILLISECONDS);
+        }
```
Ladicek: Perhaps also check `future3` and `future4`? With Awaitility I guess :-)

Ladicek:
Otherwise LGTM.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/BulkheadAsynchTest.java
Azquelt: Done

PR #495, opened 2020-01-13
fix the timeout metrics test

### LINKED ISSUE #496: timeout metrics test doesn't propagate timeout multiplier correctly
The `TimeoutMetricTest` puts the `microprofile-config.properties` file, generated by
`ConfigAnnotationAsset`, into `META-INF` in the root of a WAR. That way, the file is not on classpath and therefore isn't read by MicroProfile Config implementations, which means it isn't possible to use the timeout multiplier with this test.

All tests that use `ConfigAnnotationAsset` put everything into a JAR, and that JAR is then put into a WAR. The `TimeoutMetricTest` should do the same.

Ladicek:
When the `microprofile-config.properties` file, generated by
`ConfigAnnotationAsset`, is put into `META-INF`, the archive
needs to be a JAR. That JAR can then be put into a WAR.
All tests that use `ConfigAnnotationAsset` follow this pattern,
except `TimeoutMetricTest`, where the file is put directly
into the WAR. From there, it can't be read, as `META-INF`
inside a WAR is not on classpath.

This commit fixes `TimeoutMetricTest` to follow the same
pattern: put everything into a JAR, and that JAR into a WAR.

Resolves #496.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/CircuitBreakerRetryTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/CircuitBreakerRetryTest.java
@@ -66,7 +66,7 @@ public class CircuitBreakerRetryTest extends Arquillian {
 
     @Deployment
     public static WebArchive deploy() {
-        
+
         ConfigAnnotationAsset config = new ConfigAnnotationAsset()
                 .autoscaleMethod(CircuitBreakerClientWithRetry.class, "serviceC")
                 .autoscaleMethod(CircuitBreakerClientWithRetry.class, "serviceWithRetryOnCbOpen")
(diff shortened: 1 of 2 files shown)


Emily-Jiang:
@Ladicek can you create an issue first? In this way, it is easier to track the fixes.

Ladicek:
I thought for little TCK bugs like this, issues are not needed, but sure, I'll create one.

Azquelt:
Based on the existing code, the `ConfigAnnotationAsset` should end up in `/WEB-INF/classes/META-INF/microprofile-config.properties` which should be on the classpath. Shrinkwrap knows that stuff added as a "resource" should be put under `/WEB-INF/classes` for .wars.

That said, I'm ok with changing it to be the same as the other tests.

Ladicek:
Except the asset is not "added as a resource", it's added as "manifest resource". It 100% ends up in `/META-INF` in the WAR, at least in our impl, I've seen it in a debugger :-)

Azquelt:
Ah, sorry, I looked at your PR title and then looked for the deployment method at the top of the diff and saw you removing an `addAsResource()` method, but that's for the `CircuitBreakerRetryTest` and I didn't spot that you'd changed two files.

The `TimeoutMetricTest` is indeed wrong.

Issue #274, opened 2018-05-10
CDI request context should be active during @Asynchronous method invocation


mkouba:
Just like it's active during an asynchronous method invocation of an EJB session bean.

Emily-Jiang:
@mkouba is this a spec clarification?

mkouba:
Yes, it is.

Emily-Jiang:
Will add clarify to the spec and add a tck with `@RequestScoped` bean to demonstrate this.

PR #497, opened 2020-01-14
Async request scope test

Joseph-Cass:


Emily-Jiang:
This is the tck for #274.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/asynchronous/AsyncApplicationScopeClient.java
```diff
@@ -0,0 +1,42 @@
...
+@ApplicationScoped
+public class AsyncApplicationScopeClient {
+
+    @Inject
+    private AsyncRequestScopeClient requestScopeClient;
+
+    @Asynchronous
+    public Future<Connection> service() {
+        // AsyncRequestScopeClient is a request scope bean
+        return CompletableFutureHelper.toCompletableFuture(requestScopeClient.sampleMethod());
```
Ladicek: I think the return type of `AsyncRequestScopeClient.sampleMethod` could also be `Future<Connection>`, because `CompletableFuture` extends `Future`. This method would then just forward the call.

Ladicek:
The existing test LGTM, barring one comment 

I wonder if we should also have a test for `CompletionStage`-returning `@Asynchronous` methods?

Joseph-Cass:
Writing an additional test using CompletionStage

Emily-Jiang:
@Ladicek Please double check to see whether you are happy for this to be merged.

Issue #407, opened 2019-02-12
We need more microprofile-config tests


brunobat:
I've noticed that we don't have microprofile-config tests for all MP annotations. 
We need to add more to make sure implementations can actually override all values.

Emily-Jiang:
We also need to test the newly added extra skipOn, failOn etc on `Retry`, `CircuitBreaker`, `Fallback`

Emily-Jiang:
This is to test MP FT annotations can be configured via MP Config. e.g in the following example `
package com.acme;
public class Demo {
@Retry(skipOn="IOException.class", maxRetries=9)
public String myGet() {
...


} 
'
If in a config source, the following property is specified, the skipOn will be RuntimeException instead of IOException.
'com.acme.Demo/myGet/Retry/skipOn=RuntimeException.class'

Emily-Jiang:
@carlosdlr  can you produce a list of tests missing for the existing annotations? @Azquelt will add tests for the new annoations.

Azquelt:
Annotations can be applied to a method, or to a class. In addition, users can configure any parameter globally (for the whole application). So that's three places we can potentially test for each annotation: **method**, **class** and **global**

I think we should concentrate on having full test coverage for annotations at the method level, as that's the most common and easiest to write tests for.

This is our current test coverage for annotations at the **method** level:

- [ ] `Retry`
  - [x] `maxRetries` (covered by `ConfigTest`)
  - [ ] `delay` / `delayUnit`
  - [ ] `maxDuration` / `durationUnit` (`maxDuration` only covered by `ConfigTest`)
  - [ ] `jitter` / `jitterDelayUnit`
  - [ ] `retryOn`
  - [ ] `abortOn`
- [ ] `CircuitBreaker` (`CircuitBreakerConfigTest`)
  - [ ] `failOn`
  - [x] `skipOn`
  - [ ] `delay` / `delayUnit`
  - [ ] `requestVolumeThreshold`
  - [ ] `successThreshold`
- [ ] `Timeout`
  - [ ] `value` / `unit` (`value` only covered by `TimeoutMethodConfigTest`)
- [ ] `Bulkhead`
  - [x] `value` (covered by `BulkheadSynchConfigTest`)
  - [ ] `waitingTaskQueue`
- [ ] `Fallback` (`FallbackConfigTest`)
  - [ ] `value` (i.e. defining a `FallbackHandler`)
  - [ ] `fallbackMethod`
[...]

PR #500, opened 2020-01-17
Add configuration tests for parameters new in 2.1

Azquelt:
* Test that Fallback applyOn and skipOn can be configured
* Test that CircuitBreaker skipOn can be configured

For #407


Azquelt:
@Emily-Jiang I've added tests for configuring Circuitbreaker.skipOn, Fallback.applyOn and Fallback.skipOn globally.

PR #501, opened 2020-01-20
Adding more config tests

carlosdlr:
issue https://github.com/eclipse/microprofile-fault-tolerance/issues/407

Signed-off-by: Carlos Andres De La Rosa <kusanagi12002@gmail.com>

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/TimeoutUninterruptableTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/TimeoutUninterruptableTest.java
@@ -1,6 +1,6 @@
 /*
  *******************************************************************************
- * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ * Copyright (c) 2018-2020 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
@@ -203,13 +203,14 @@ public class TimeoutUninterruptableTest extends Arquillian {
      * Test that the timeout timer is started when the execution is added to the queue
      * 
      * @throws InterruptedException if the test is interrupted
+     * @throws ExecutionException 
      */
     @Test
-    public void testTimeoutAsyncBulkheadQueueTimed() throws InterruptedException {
+    public void testTimeoutAsyncBulkheadQueueTimed() throws InterruptedException, ExecutionException {
         CompletableFuture<Void> waitingFutureA = newWaitingFuture();
         CompletableFuture<Void> waitingFutureB = newWaitingFuture();
         
-        client.serviceTimeoutAsyncBulkheadQueueTimed(waitingFutureA);
+        Future<Void> resultA = client.serviceTimeoutAsyncBulkheadQueueTimed(waitingFutureA);
         Thread.sleep(config.getTimeoutInMillis(100));
         
         long startTime = System.nanoTime();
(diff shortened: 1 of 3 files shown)


tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/config/fallback/ConfigPropertyGlobalVsMethodTest.java
```diff
@@ -0,0 +1,76 @@
...
+    @Inject
+    private BeanWithFallback bean;
+
+    @Deployment
+    public static WebArchive deployAnotherApp() {
+        JavaArchive testJar = ShrinkWrap
+            .create(JavaArchive.class, "ftConfigFallback.jar")
+            .addClasses(BeanWithFallback.class)
+            .addAsManifestResource(
+            new StringAsset(
```
Ladicek: Use the `ConfigAnnotationAsset`?

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/config/fallback/ConfigPropertyGlobalVsMethodTest.java
```diff
@@ -0,0 +1,76 @@
...
+     */
+    @Test
+    void propertyGlobalConfAccessTest() {
+        try {
+            bean.serviceA();
+        }
+        catch (IllegalStateException e) {
+            e.printStackTrace();
+        }
+        Assert.assertEquals(bean.getCalledMethod(), "fallbackGlobalConfForServiceA");
```
Ladicek: Better to use a return value, instead of modifying internal state? Also swallowing the exception is not nice :-)

Emily-Jiang:
@cealsair will repurpose this PR to cover the existing methods or parameters instead of the new ones. @Azquelt will comment on the issues on the reminding untested parameters.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/config/fallback/ConfigPropertyGlobalVsMethodTest.java
carlosdlr: @Ladicek sorry but I don't understand this comment, can you explain? thank you

carlosdlr: I understand now what you mean, you refer to the ConfigAnnotationAsset class to build the properties file.

carlosdlr: I was checking the class ConfigAnnotationAsset and looks like that is designed for handling time values, scale timeouts and enable annotations. I don't see in the class a  way to set this kind of values `fallbackMethod=fallbackGlobalConfForServiceA`.

Ladicek: Something like:

```java
new ConfigAnnotationAsset()
    .set(BeanWithFallback.class, "serviceA", Fallback.class, "fallbackMethod", "fallbackGlobalConfForServiceA")
    .set(BeanWithFallback.class, "serviceA", Fallback.class, "applyOn", RuntimeException.class.getName())
```

(Actually just noticed your `applyOn=RuntimeException.class` -- that wrong AFAIK? The code above should fix that.)

carlosdlr: yes that will work thank you @Ladicek

Ladicek:
I don't understand why GitHub shows so many changes now, unrelated to the commits in question. Perhaps it would have been better to rebase instead of merge.

PR #506, opened 2020-01-23
Use separate methods in CircuitBreakerExceptionHierarchyTest

Azquelt:
Change the CircuitBreakerExceptionHierarchyTest to use a different
service method for each test.

This requires the service method to be replicated lots of times, but
allows the lifetime of the circuit breaker to be tied to the method
itself rather than to the bean instance.


Azquelt:
Pushed a tiny correction to the interface name and the assertion messages.

PR #511, opened 2020-02-12
Make timeout for concurrent execution tracker configurable

### LINKED ISSUE #510: Timeout for concurrent execution tracker should be configurable
Currently, in [ConcurrentExecutionTracker.java](https://github.com/eclipse/microprofile-fault-tolerance/blob/master/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/util/ConcurrentExecutionTracker.java) `WAIT_TIMEOUT` is set to `3L * 1000 * 1_000_000`.

This means that anywhere that uses `waitForRunningExecutions()` waits for up to 3 seconds.

This class is used by some tests which have configurable timeouts, e.g. [BulkheadMetricTest.java](https://github.com/eclipse/microprofile-fault-tolerance/blob/master/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/BulkheadMetricTest.java). Can we configure this timeout as well so that it is adjusted in line with the rest of the test?

Joseph-Cass:
Resolves #510

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/util/ConcurrentExecutionTracker.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/util/ConcurrentExecutionTracker.java
@@ -48,7 +48,8 @@ public class ConcurrentExecutionTracker {
     // The Atomicness is not important, this is just an integer holder
     private final AtomicInteger executionCount = new AtomicInteger(0);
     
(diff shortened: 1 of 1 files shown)


genie-microprofile:

Azquelt:

Azquelt:
Your new commit has failed the eca check because it doesn't have the `signed-off-by` line. You can either amend the commit to add it, or squash it into the previous commit.

PR #515, opened 2020-02-26
Typo and grammar corrections

Joseph-Cass:
Some of these changes will be up for discussion, as some are preferences rather than corrections.

I've tried to follow some standards, such as:
* no capitals after @return's
* colons before listed items
*  consistent full stops, excluding:
    *  listed items which are not sentences
    * sentences following @return's


genie-microprofile:

Azquelt:

PR #517, opened 2020-03-05
Fixed TimeoutUninterruptableTest scaling issue with TCKConfig baseMultiplier

### LINKED ISSUE #516: TimeoutUninterruptableTest doesnt scale well with baseMultiplier
[TimeoutUninterruptableTest](https://github.com/eclipse/microprofile-fault-tolerance/blob/master/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/TimeoutUninterruptableTest.java) fails with a large enough [`baseMultiplier`](https://github.com/eclipse/microprofile-fault-tolerance/blob/fbfa129e32380eac64544d6eac8f88f0aa26e79c/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/util/TCKConfig.java#L40).

This is due to the `serviceTimeout`'s in [UninterruptableTimeoutClient](https://github.com/eclipse/microprofile-fault-tolerance/blob/master/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/timeout/clientserver/UninterruptableTimeoutClient.java) waiting for a fixed maximum of 5 seconds. This means that with a large enough `baseMultiplier` then the `waitingFuture.get()` methods are getting `TimeoutException`'s before the expected scaled `@Timeout` duration from `@Timeout(value = 500, unit = ChronoUnit.MILLIS)`. (where the `@Timeout` duration is `value * baseMultiplier`)

The `waitingFuture.get()`'s should be scaled with the `baseMultiplier`.

Joseph-Cass:
Fixes #516

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/timeout/clientserver/UninterruptableTimeoutClient.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/timeout/clientserver/UninterruptableTimeoutClient.java
@@ -1,6 +1,6 @@
 /*
  *******************************************************************************
- * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ * Copyright (c) 2018-2020 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/timeout/clientserver/UninterruptableTimeoutClient.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/timeout/clientserver/UninterruptableTimeoutClient.java
@@ -19,7 +19,7 @@
  *******************************************************************************/
 package org.eclipse.microprofile.fault.tolerance.tck.timeout.clientserver;
 
-import static java.util.concurrent.TimeUnit.SECONDS;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 import java.time.Duration;
 import java.time.temporal.ChronoUnit;
@@ -32,6 +32,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 import javax.enterprise.context.RequestScoped;
 
+import org.eclipse.microprofile.fault.tolerance.tck.util.TCKConfig;
 import org.eclipse.microprofile.faulttolerance.Asynchronous;
 import org.eclipse.microprofile.faulttolerance.Bulkhead;
 import org.eclipse.microprofile.faulttolerance.Fallback;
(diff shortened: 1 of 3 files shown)


genie-microprofile:

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/timeout/clientserver/UninterruptableTimeoutClient.java
```diff
@@ -94,7 +98,7 @@ public void serviceTimeout(long waitMs) {
                 Assert.fail("Waiting future threw exception", e);
             }
             catch (TimeoutException e) {
-                return CompletableFuture.completedFuture(null);
+                Assert.fail("Waiting future timed out", e);
```
Ladicek: I wanted to ask if this is guaranteed to run on the testing thread (otherwise asserts are useless), but I see we do that already in the previous `catch` block.

Joseph-Cass: I did also test that this exception is thrown when it should be and that the failure message is thrown:

I configured a `WAITING_FUTURE_DURATION` less than the scaled `@Timeout value` and the failure message was returned as expected :)

Joseph-Cass:
I found a strangely formatted [copyright header](https://github.com/eclipse/microprofile-fault-tolerance/pull/517/files#diff-68cf39706775452d2f1a9d44cdfe623a) whilst I was updating my own copyright header, which i've resolved.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/timeout/clientserver/UninterruptableTimeoutClient.java
Azquelt: It doesn't run on the testing thread, it results in an `AssertionError` being thrown on the async thread, which is then returned to the testing thread via the `Future`. In almost all cases, the test is asserting the result from the `Future`.

I do spot one case where it isn't in `testTimeoutAsyncBulkheadQueueTimed`. @Joseph-Cass could you add a check to the result of the first call to `serviceTimeoutAsyncBulkheadQueueTimed`. Calling `future.get()` after completing `waitingFutureA` should be sufficient to ensure the failure exception is thrown.

Ladicek: I see. Thanks!

Joseph-Cass: I have added an additional check for `testTimeoutAsyncBulkheadQueueTimed` in commit [a4555ff](https://github.com/eclipse/microprofile-fault-tolerance/pull/517/commits/a4555ff97991119719c3e57e629032dac09954fa), thanks for spotting the missing case @Azquelt

Azquelt:
@genie-microprofile test please

Azquelt:

PR #520, opened 2020-03-27
Updated AllMetricsBean

### LINKED ISSUE #519: Timeout value for metrics.AllMetricsBean should be configurable
[AllMetricsTest](https://github.com/eclipse/microprofile-fault-tolerance/blob/master/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/AllMetricsTest.java) uses  [AllMetricsBean](https://github.com/eclipse/microprofile-fault-tolerance/blob/master/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/AllMetricsBean.java) to test `that metrics are created when all the Fault Tolerance annotations are placed on the same method`.

This bean has a Timeout value of 1 second, which is irrelevant to the test.

This value should be configurable.

Joseph-Cass:
Fixes #519 

Updated AllMetricsBean: 
* Increased the Timeout duration to 1 minute
* Scaled all the FT values  with the the TCKConfig baseMultiplier using the autoscaleMethod

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/AllMetricsTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/AllMetricsTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ * Copyright (c) 2018-2020 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/AllMetricsTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/AllMetricsTest.java
@@ -26,6 +26,7 @@ import java.util.concurrent.ExecutionException;
 
 import javax.inject.Inject;
 
+import org.eclipse.microprofile.fault.tolerance.tck.config.ConfigAnnotationAsset;
 import org.eclipse.microprofile.fault.tolerance.tck.metrics.util.MetricGetter;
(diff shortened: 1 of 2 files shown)


genie-microprofile:

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/AllMetricsTest.java
```diff
@@ -42,11 +43,17 @@
...
+        ConfigAnnotationAsset allMetricsBeanConfig = new ConfigAnnotationAsset()
+                .autoscaleMethod(AllMetricsBean.class, "doWork");
+
         WebArchive war = ShrinkWrap.create(WebArchive.class, "ftMetricAll.war")
                 .addClasses(AllMetricsBean.class)
                 .addPackage(Packages.UTILS)
                 .addPackage(Packages.METRIC_UTILS)
-                .addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml");
+                .addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml")
+                .addAsManifestResource(allMetricsBeanConfig, "microprofile-config.properties");
```
Azquelt: I don't think `addAsManifestResource` works correctly here when the `microprofile-config.properties` is being added to a .war file.

MP Config requires the file to be found on the classpath, whereas I think your code results in the file being located in `META-INF/microprofile-config.properties` which _isn't_ included in the classpath for a .war.

We could place the file elsewhere in the .war, however, could you instead change this deployment method to a a jar and a .war file instead? Most of the tests are packaged this way, it's only some of the metrics tests that aren't. See `TimeoutMetricTest` for an example of a class which does this correctly.

Joseph-Cass: I have added a new commit to resolve this issue

PR #523, opened 2020-03-31
for each annotation, add a listing of all atributes at the end of main javadoc

### LINKED ISSUE #522: add a listing of all attributes to all FT annotations javadoc
We just discussed with @Joseph-Cass that it would be helpful for the main javadoc of each annotation to contain a listing of all attributes in the form of `@see #value`, `@see #failOn` etc. This significantly improves navigation when reading the javadoc in an IDE.

I have done that and will submit a PR for consideration.

Ladicek:

Resolves #522.


Ladicek:
CC @Joseph-Cass -- this is what we discussed.

Joseph-Cass:
Yes I like this and think I will find it useful when developing with MP FT :)

PR #525, opened 2020-03-31
specify lifecycle of stateful fault tolerance strategies

### LINKED ISSUE #479: specify lifecycle of the circuit breaker
Circuit breaker maintains state over time, between various invocations of the guarded method. Therefore, the circuit breaker object needs to have a lifecycle. For example, in SmallRye Fault Tolerance, all circuit breakers are singletons -- no matter the scope of a bean (be it `@Dependent` or `@RequestScoped`), all invocations of a single method are guarded by the same circuit breaker instance.

https://github.com/eclipse/microprofile-fault-tolerance/issues/302 suggests that the circuit breaker should follow the lifecycle of the bean, but to the best of my knowledge, this isn't specified. There are no tests for it in the TCK either.

I only found out because the new `CircuitBreakerExceptionHierarchyTest` test relies on that behavior.

Ladicek:

Resolves #479.

### DIFF
--- a/spec/src/main/asciidoc/bulkhead.asciidoc
+++ b/spec/src/main/asciidoc/bulkhead.asciidoc
@@ -18,12 +18,13 @@
 // Contributors:
 // Emily Jiang
 // Andrew Rouse
+// Ladislav Thon
 
 [[bulkhead]]
 == Bulkhead
 
 The `Bulkhead` pattern is to prevent faults in one part of the system from cascading to the entire system, which might bring down the whole system.
-The implementation is to limit the number of concurrent requests accessing to an instance. Therefore, `Bulkhead` pattern is only effective when applying `@Bulkhead` to a component that can be accessed from multiple contexts.
+The implementation is to limit the number of concurrent requests accessing an instance. Therefore, `Bulkhead` pattern is only effective when applying `@Bulkhead` to a component that can be accessed from multiple contexts.
 
 === Bulkhead Usage
 A method or class can be annotated with `@Bulkhead`, which means the method or the methods under the class will have Bulkhead policy applied correspondingly.
@@ -70,6 +71,17 @@ public Future<Connection> serviceA() {
 
 
 When using the thread pool approach, when a request cannot be added to the waiting queue, `BulkheadException` will be thrown.
+
+==== Lifecycle
+
+Bulkhead needs to maintain some state between invocations: the number of currently running executions, or the queue of waiting executions.
+This state is a singleton, irrespective of the lifecycle of the bean that uses the `@Bulkhead` annotation.
+
+More specifically, the bulkhead state is uniquely identified by the combination of the bean class (`java.lang.Class`) and the method object (`java.lang.reflect.Method`) representing the guarded method.
+
+For example, if there's a guarded method `doWork` on a bean which is `@RequestScoped`, each request will have its own instance of the bean, but all invocations of `doWork` will share the same bulkhead state.
+
+=== Interactions with other annotations
 
 The `@Bulkhead` annotation can be used together with `@Fallback`, `@CircuitBreaker`, `@Asynchronous`, `@Timeout` and `@Retry`.
 
--- a/spec/src/main/asciidoc/circuitbreaker.asciidoc
+++ b/spec/src/main/asciidoc/circuitbreaker.asciidoc
@@ -18,6 +18,7 @@
 // Contributors:
 // Emily Jiang
 // Andrew Rouse
+// Ladislav Thon
 
 [[circuitbreaker]]
 == Circuit Breaker
@@ -71,7 +72,7 @@ public Connection serviceA() {
 ** Request 3 - success
 ** Request 4 - success
 ** Request 5 - failure
-** Request 6 - CircuitBreakerOpenException 
+** Request 6 - CircuitBreakerOpenException
 
 In this scenario, request 5 will trigger the circuit to open because out of the last four requests (the `requestVolumeThreshold`), two failed
 which reaches the `failureRatio` of `0.5`. Request 6 will therefore hit the `CircuitBreakerOpenException`.
@@ -109,6 +110,15 @@ public void service() {
     underlyingService();
 }
 ----
+
+==== Lifecycle
+
+Circuit breaker needs to maintain some state between invocations: the number of recent successful and failed invocations, or how long has the circuit breaker been open.
+This state is a singleton, irrespective of the lifecycle of the bean that uses the `@CircuitBreaker` annotation.
+
+More specifically, the circuit breaker state is uniquely identified by the combination of the bean class (`java.lang.Class`) and the method object (`java.lang.reflect.Method`) representing the guarded method.
+
+For example, if there's a guarded method `doWork` on a bean which is `@RequestScoped`, each request will have its own instance of the bean, but all invocations of `doWork` will share the same circuit breaker state.
 
 === Interactions with other annotations
 
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/lifecycle/BulkheadLifecycleTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/lifecycle/BulkheadLifecycleTest.java
@@ -1,0 +1,216 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+package org.eclipse.microprofile.fault.tolerance.tck.bulkhead.lifecycle;
+
+import org.eclipse.microprofile.fault.tolerance.tck.util.AsyncCaller;
+import org.eclipse.microprofile.fault.tolerance.tck.util.Barrier;
+import org.eclipse.microprofile.faulttolerance.exceptions.BulkheadException;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.testng.Arquillian;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.EmptyAsset;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.testng.annotations.Test;
+
+import javax.enterprise.inject.Instance;
+import javax.inject.Inject;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import static org.awaitility.Awaitility.await;
+import static org.eclipse.microprofile.fault.tolerance.tck.util.Exceptions.expect;
+
+/**
+ * Test that bulkhead is a singleton, even if the bean is not.
+ */
+public class BulkheadLifecycleTest extends Arquillian {
+    @Deployment
+    public static WebArchive deploy() {
+        JavaArchive testJar = ShrinkWrap
+                .create(JavaArchive.class, "ftBulkheadLifecycle.jar")
+                .addPackage(BulkheadLifecycleService1.class.getPackage())
+                .addPackage(AsyncCaller.class.getPackage())
+                .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml")
+                .as(JavaArchive.class);
+
+        return ShrinkWrap
+                .create(WebArchive.class, "ftBulkheadLifecycle.war")
+                .addAsLibrary(testJar);
+    }
+
+    // verify that bulkhead is shared between instances of the same class and for the same method,
+    // but not shared between different classes and different methods of the same class
+
+    @Inject
+    private AsyncCaller async;
+
+    @Inject
+    private Instance<BulkheadLifecycleService1> service1;
+
+    @Inject
+    private Instance<BulkheadLifecycleService2> service2;
+
+    @Inject
+    private Instance<BulkheadLifecycleServiceSubclass1> subclassService1;
+
+    @Inject
+    private Instance<BulkheadLifecycleServiceSubclass2> subclassService2;
+
+    @Inject
+    private Instance<MutlipleMethodsBulkheadLifecycleService> multipleMethodsService;
+
+    @Test
+    public void noSharingBetweenClasses() throws InterruptedException, ExecutionException, TimeoutException {
+        Barrier barrier = new Barrier();
+
+        BulkheadLifecycleService1 service1a = service1.get();
+        BulkheadLifecycleService1 service1b = service1.get();
+
+        BulkheadLifecycleService2 service2a = service2.get();
+        BulkheadLifecycleService2 service2b = service2.get();
+
+        List<Future<Void>> futures = new ArrayList<>();
+        try {
+            for (int i = 0; i < 4; i++) {
+                futures.add(async.run(() -> service1a.service(barrier)));
+                futures.add(async.run(() -> service2a.service(barrier)));
+                futures.add(async.run(() -> service1b.service(barrier)));
+                futures.add(async.run(() -> service2b.service(barrier)));
+            }
+
+            // wait until all submissions have entered the bulkhead
+            // - there are 4 iterations of the loop, each invoking 4 methods that will wait on the barrier
+            // - there are 2 bulkheads, each allowing 8 executions
+            await().atMost(10, TimeUnit.SECONDS).until(() -> barrier.countWaiting() == 16);
+
+            expect(BulkheadException.class, async.run(() -> service1a.service(barrier)));
+            expect(BulkheadException.class, async.run(() -> service2a.service(barrier)));
+            expect(BulkheadException.class, async.run(() -> service1b.service(barrier)));
+            expect(BulkheadException.class, async.run(() -> service2b.service(barrier)));
+        }
+        finally {
+            try {
+                barrier.open();
+
+                for (Future<Void> future : futures) {
+                    future.get(1, TimeUnit.MINUTES);
+                }
+            }
+            finally {
+                service1.destroy(service1a);
+                service1.destroy(service1b);
+                service2.destroy(service2a);
+                service2.destroy(service2b);
+            }
+        }
+    }
+
+    @Test
+    public void noSharingBetweenClassesWithCommonSuperclass() throws InterruptedException, ExecutionException, TimeoutException {
+        Barrier barrier = new Barrier();
+
+        BulkheadLifecycleServiceSubclass1 service1a = subclassService1.get();
+        BulkheadLifecycleServiceSubclass1 service1b = subclassService1.get();
+
+        BulkheadLifecycleServiceSubclass2 service2a = subclassService2.get();
+        BulkheadLifecycleServiceSubclass2 service2b = subclassService2.get();
+
+        List<Future<Void>> futures = new ArrayList<>();
+        try {
+            for (int i = 0; i < 4; i++) {
+                futures.add(async.run(() -> service1a.service(barrier)));
+                futures.add(async.run(() -> service2a.service(barrier)));
+                futures.add(async.run(() -> service1b.service(barrier)));
+                futures.add(async.run(() -> service2b.service(barrier)));
+            }
+
+            // wait until all submissions have entered the bulkhead
+            // - there are 4 iterations of the loop, each invoking 4 methods that will wait on the barrier
+            // - there are 2 bulkheads, each allowing 8 executions
+            await().atMost(10, TimeUnit.SECONDS).until(() -> barrier.countWaiting() == 16);
+
+            expect(BulkheadException.class, async.run(() -> service1a.service(barrier)));
+            expect(BulkheadException.class, async.run(() -> service2a.service(barrier)));
+            expect(BulkheadException.class, async.run(() -> service1b.service(barrier)));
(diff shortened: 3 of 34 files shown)


Ladicek:
This is a draft, because I need to [figure out how to] add some TCK tests. However, feedback about the specification wording is welcome!

spec/src/main/asciidoc/bulkhead.asciidoc
```diff
@@ -71,8 +71,19 @@ public Future<Connection> serviceA() {
 
 When using the thread pool approach, when a request cannot be added to the waiting queue, `BulkheadException` will be thrown.
 
+==== Lifecycle
+
+Unlike some other fault tolerance strategies, bulkhead needs to store some data between invocations: the number of currently running executions, or the queue of waiting executions.
+This state is a singleton, irrespective of the lifecycle of the bean that uses the `@Bulkhead` annotation.
```
Emily-Jiang, line 77: It is associated with the class instead of the instances.

spec/src/main/asciidoc/bulkhead.asciidoc
```diff
@@ -71,8 +71,19 @@ public Future<Connection> serviceA() {
...
+If, for example, the bean is `@RequestScoped` and therefore there is a new instance of the bean for each request, all instances share the same bulkead state.
```
Emily-Jiang, line 78: blukead -> bulkhead

spec/src/main/asciidoc/bulkhead.asciidoc
```diff
@@ -71,8 +71,19 @@ public Future<Connection> serviceA() {
 
 When using the thread pool approach, when a request cannot be added to the waiting queue, `BulkheadException` will be thrown.
 
+==== Lifecycle
+
+Unlike some other fault tolerance strategies, bulkhead needs to store some data between invocations: the number of currently running executions, or the queue of waiting executions.
```
Emily-Jiang, line 76: Delete "Unlike some other fault tolerance strategies," as Circuit Breaker has the same stategy.

spec/src/main/asciidoc/circuitbreaker.asciidoc
```diff
@@ -109,6 +109,15 @@ public void service() {
 }
 ----
 
+==== Lifecycle
+
+Unlike some other fault tolerance strategies, circuit breaker needs to store some data between invocations: the number of recent successful and failed invocations, or how long has the circuit breaker been open.
```
Emily-Jiang, line 114: Delete "Unlike some other fault tolerance strategies," as Bulkhead has the same stategy.

Ladicek:
I tried to reword it slightly, based on your comments. Thanks @Emily-Jiang!

spec/src/main/asciidoc/bulkhead.asciidoc
Ladicek: I've changed it to use both explanations.

spec/src/main/asciidoc/bulkhead.asciidoc
```diff
@@ -71,8 +71,20 @@ public Future<Connection> serviceA() {
...
 
+==== Lifecycle
+
+Bulkhead needs to needs to maintain some state between invocations: the number of currently running executions, or the queue of waiting executions.
+This state is a singleton, irrespective of the lifecycle of the bean that uses the `@Bulkhead` annotation.
+In other words, the state is associated with the bean class, not with its instances.
+For example, if the bean is `@RequestScoped` and therefore there is a new instance of the bean for each request, all instances share the same bulkhead state.
+
+More specifically: each method of each class that uses a bulkhead has its own state.
+If a subclass inherits a `@Bulkhead` method, then the subclass has its own, distinct, bulkhead state.
```
Azquelt, line 82: I found this part surprising.

If a method is inherited by a subclass, why should it receive a distinct bulkhead state for the subclass? I has assumed that each method declaration would have a separate state, rather than each method declaration and subclass combination having a separate state.

Do you see advantages to defining it one way or the other?

I would see having one state per method declaration as consistent with the way we use annotations to apply fault tolerance policies to methods. When deciding which policies to apply, you must find the declaration of the method which is to be run and then consider:
 * annotations applied to the method
 * annotations applied to the class which declares the method
 * annotations applied to superclasses of the class which declares the method (because our annotations have `@Inherited`)

You ignore:
 * annotations applied to any method that the method overrides
 * annotations applied to any subclass of the class which declares the method

Ladicek: I think it was you who pointed out that subclasses should have their own circuit breakers (applies to bulkhead as well): https://github.com/eclipse/microprofile-fault-tolerance/issues/479#issuecomment-561197781

But you raise very good points. I'll try to address them (tomorrow).

Azquelt: Heh, when I was writing this, it did ring a few bells.

Unfortunately, I suspect whichever way round we do it, we might catch someone out :(

I think on balance, I'm leaning towards the state being per method declaration for the reasons set out above. It's also fairly easy to override the method and delegate to `super` if you do want a separate bulkhead state.

Ladicek: I didn't forget about this, I just had a ton of other work. I got back to it today, and changed to your suggestion. I agree that tying the state to the method declaration makes most sense.

I'm wondering if something happens to a `@CircuitBreaker` method if a subclass which _doesn't_ override it adds a `@CircuitBreaker` annotation. I'd say no, but I'd also say that this should be specified somewhere for all annotations. Same thing with configuration overrides -- can I override configuration of a `@CircuitBreaker` method on a subclass, even if that subclass doesn't override the method? This is something that should be specified for all annotations. (I'm not sure if it is or not, didn't check :-) )

Ladicek:
I have started adding TCK tests for this. For now, I started with circuit breaker, and will continue with bulkhead later, when we agree on desired behavior for CB.

The current TCK tests mirror existing tests for usage of `@Retry` in inheritance hierarchy, both in their structure and their expectations. The specification wording hence is _not_ aligned to what the current TCK tests expect.

I present this for discussion, not as a definitive version that is ready for merge.

Azquelt:
> The current TCK tests mirror existing tests for usage of `@Retry` in inheritance hierarchy, both in their structure and their expectations. The specification wording hence is _not_ aligned to what the current TCK tests expect.

Bah, yes, it looks like back in #183 it was decided that annotations on the current class should be considered, rather than just the ones on the declaring class.

I didn't realise that. Obviously we can't have circuitbreakers with different settings sharing the same state so either we'd have to change the existing behaviour, or redefine how the state is shared here.

I don't really like the idea of changing the existing behaviour since I think it's fairly likely that it'll break existing code.

Ladicek:
Yea. I need to re-read the spec to figure out the right wording for "where" is the CB state located.

Emily-Jiang:
We discussed today we agreed the circuit breaker scope is per class per method.

Ladicek:
I have just adjusted the specification wording, it should be now in sync with the TCK tests.

Now I need to add more TCK tests.

Ladicek:
I have added a few more tests for circuit breaker, and added some tests for bulkhead. I don't think it's necessary to test all the inheritance hierarchy scenarios with bulkhead, because they have been tested with retry for a long time, and are now tested with circuit breaker as well.

spec/src/main/asciidoc/bulkhead.asciidoc
```diff
@@ -71,8 +71,21 @@ public Future<Connection> serviceA() {
 
 When using the thread pool approach, when a request cannot be added to the waiting queue, `BulkheadException` will be thrown.
 
+==== Lifecycle
+
+Bulkhead needs to needs to maintain some state between invocations: the number of currently running executions, or the queue of waiting executions.
```
Emily-Jiang, line 76: dup of `needs to`

spec/src/main/asciidoc/bulkhead.asciidoc
```diff
@@ -71,8 +71,21 @@ public Future<Connection> serviceA() {
...
+This state is a singleton, irrespective of the lifecycle of the bean that uses the `@Bulkhead` annotation.
+In other words, the state is associated with the bean class, not with its instances.
+For example, if the bean is `@RequestScoped` and therefore there is a new instance of the bean for each request, all instances share the same bulkhead state.
+
+More specifically: the bulkhead state is bound to the bean class _and_ the guarded method.
+If all the bulkheads were stored in a `Map`, then the key to the `Map` would be a combination of the bean class (`java.lang.Class`) and the method object (`java.lang.reflect.Method`).
```
Emily-Jiang, line 82: I would reword to be:
The combination of the bean class (`java.lang.Class`) and the method object (`java.lang.reflect.Method`) uniquely identify a bulkhead.

spec/src/main/asciidoc/bulkhead.asciidoc
```diff
@@ -71,8 +71,21 @@ public Future<Connection> serviceA() {
...
+Implementations are of course free to use any strategy they wish.
```
Emily-Jiang, line 83: This can be deleted.

spec/src/main/asciidoc/circuitbreaker.asciidoc
```diff
@@ -110,6 +110,17 @@ public void service() {
 }
 ----
 
+==== Lifecycle
+
+Circuit breaker needs to maintain some state between invocations: the number of recent successful and failed invocations, or how long has the circuit breaker been open.
+This state is a singleton, irrespective of the lifecycle of the bean that uses the `@CircuitBreaker` annotation.
+In other words, the state is associated with the bean class, not with its instances.
+For example, if the bean is `@RequestScoped` and therefore there is a new instance of the bean for each request, all instances share the same circuit breaker state.
```
Emily-Jiang, line 118: This line suggests one circuit breaker per class. The following paragraph states it is associated with the class and the methods.

spec/src/main/asciidoc/circuitbreaker.asciidoc
```diff
@@ -110,6 +110,17 @@ public void service() {
...
+
+More specifically: the circuit breaker state is bound to the bean class _and_ the guarded method.
+If all the circuit breakers were stored in a `Map`, then the key to the `Map` would be a combination of the bean class (`java.lang.Class`) and the method object (`java.lang.reflect.Method`).
```
Emily-Jiang, line 121: https://github.com/eclipse/microprofile-fault-tolerance/pull/525/files#r430455078

Emily-Jiang: see above comments

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/lifecycle/MutlipleMethodsBulkheadLifecycleService.java
```diff
@@ -0,0 +1,37 @@
...
+import javax.enterprise.context.Dependent;
+
+@Dependent
+public class MutlipleMethodsBulkheadLifecycleService {
+    @Bulkhead(8)
+    public void service1(Barrier barrier) {
+        barrier.await();
+    }
+
+    @Bulkhead(8)
```
Emily-Jiang, line 33: It works better if you use a different number.

Ladicek: The purpose is to verify that they are different (even though they perhaps have the same number of allowed executions).

spec/src/main/asciidoc/bulkhead.asciidoc
Ladicek: Thanks!

spec/src/main/asciidoc/bulkhead.asciidoc
Ladicek: Done.

spec/src/main/asciidoc/bulkhead.asciidoc
Ladicek: Done.

spec/src/main/asciidoc/circuitbreaker.asciidoc
Ladicek: Reworded.

spec/src/main/asciidoc/bulkhead.asciidoc
Ladicek: Done.

Ladicek:
I did some rewording based on @Emily-Jiang's suggestions (thanks!).

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/circuitbreaker/lifecycle/BaseCircuitBreakerOnMethod.java
```diff
@@ -0,0 +1,45 @@
...
+package org.eclipse.microprofile.fault.tolerance.tck.circuitbreaker.lifecycle;
+
+import org.eclipse.microprofile.faulttolerance.CircuitBreaker;
+
+import javax.enterprise.context.Dependent;
+import java.io.IOException;
+import java.util.concurrent.atomic.AtomicInteger;
+
+@Dependent
+@CBLifecycle(CBLifecycleServiceType.BASE_M)
```
Azquelt, line 29: I really don't like using a qualifier like this to distinguish between the test beans, it makes it much harder to navigate through the test code to find out which test is using which bean.

I realise that we already have some tests which do it this way. I assume this is because otherwise when one class extends another you get multiple beans which have the superclass as a bean type. If so, could we use `@Typed` to restrict the bean types instead.

I'd need to check it still works but I have a commit which replaces the qualifiers with `@Typed` instead.

Azquelt: Here's the change I'd propose: https://github.com/Azquelt/microprofile-fault-tolerance/commit/828c2926be4ee51de216544ac242d1ad4f644eee

It does seem to work (in that all the tests fail in the expected way when run against open-liberty since we haven't implemented this yet)

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/lifecycle/AsyncInvoker.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import static java.util.concurrent.CompletableFuture.completedFuture;
+import static org.testng.Assert.fail;
+
+@ApplicationScoped
+public class AsyncInvoker {
```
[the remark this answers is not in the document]
Ladicek: I didn't know about `Exceptions.expect()`, it is indeed exactly the same and I'll use that.

When it comes to `AsyncCaller`, its `run` method (unlike `submit`) doesn't catch exceptins, and I'm not sure if that's intentional or mistake. So I created my own.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/lifecycle/Barrier.java
```diff
@@ -0,0 +1,43 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+package org.eclipse.microprofile.fault.tolerance.tck.bulkhead.lifecycle;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+
+public class Barrier {
```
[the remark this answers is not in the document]
Ladicek: Do you know if TestNG has something akin to JUnit 4 rules? That's what I'd do if this was JUnit 4.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/lifecycle/BulkheadLifecycleTest.java
```diff
@@ -0,0 +1,151 @@
...
+    private Instance<BulkheadLifecycleService1> service1;
+
+    @Inject
+    private Instance<BulkheadLifecycleService2> service2;
+
+    @Inject
+    private Instance<MutlipleMethodsBulkheadLifecycleService> multipleMethodsService;
+
+    @Test
+    public void noSharingBetweenClasses() throws InterruptedException, ExecutionException, TimeoutException {
```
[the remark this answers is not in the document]
Ladicek: Agree, will add the test.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/circuitbreaker/lifecycle/BaseCircuitBreakerOnMethod.java
Ladicek: This is interesting, I didn't know about `@Typed`. I'll check.

Ladicek: And yea, I copied what we do for the `@Retry` tests, even though I didn't like it. Your commit looks like a very nice simplification!

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/lifecycle/Barrier.java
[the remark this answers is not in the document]
Azquelt: So would I. I did look but TestNG doesn't seem to have anything equivalent. You could probably put together something with an `IInvokedMethodListener` which used reflection to find a field in the class but you'd have to check how it interacts with Arquillian.

The Arquillian extension mechanisms would be another option. I think you'd need to make a remote extension which observes before and after events, and then a local extension to make sure your remote extension gets packaged in each application. Again, you might need to check how that interacts with TestNG's lifecycle.

Lastly, we'd need some way of making what's going on clear to whoever is reading the tests.

A further thought: it might make sense to have an arquillian extension which fires CDI events at the start and end of each test method running on the server. Then we could implement any test fixtures as beans which observe those events.

I'm still on the fence a bit as to whether it's worth doing any of this rather than just having an `@AfterMethod` which adds a little more boilerplate in the test class, which is what we've done elsewhere.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/lifecycle/AsyncInvoker.java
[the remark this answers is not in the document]
Ladicek: Removed the assert method, but left `AsyncInvoker` for now.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/lifecycle/BulkheadLifecycleTest.java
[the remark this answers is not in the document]
Ladicek: Added.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/circuitbreaker/lifecycle/BaseCircuitBreakerOnMethod.java
Ladicek: Played with it a little and then just used your commit -- thanks!

Ladicek:
When #548 is merged, I will add release notes to this PR as well.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/lifecycle/AsyncInvoker.java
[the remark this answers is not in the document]
Azquelt: I had to look at this for a while before working out what I think is going on here.

I think `AsyncCaller.run()` doesn't catch exceptions because `@Asynchronous` will catch all thrown exceptions anyway.

`AsyncCaller.submit()` does catch exceptions because `Callable` throws checked exceptions so they have to be either caught or declared as thrown, even though `@Asynchronous` will always catch them.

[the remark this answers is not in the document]
Ladicek: Ah, yea! Got it -- sorry it takes so long :-)

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/lifecycle/Barrier.java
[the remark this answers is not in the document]
Ladicek: I'm thinking you not only need to open the barrier, you also need to wait for all the async operations to finish. That's why I did all the `try`/`catch`/`finally` dance. I'll try to think about this some more.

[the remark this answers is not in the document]
Azquelt: Yeah, that may be true. I'll think on it some more too. For this PR, it's fine as-is.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/bulkhead/lifecycle/MutlipleMethodsBulkheadLifecycleService.java
```diff
@@ -0,0 +1,38 @@
...
+    public void service2(Barrier barrier) {
+        barrier.await();
+    }
```
Azquelt, line 37: Would this test be improved if the `@Bulkhead` annotation was on the class rather than the method?

The behaviour should be the same (separate bulkhead for each method).

Same question for `MutlipleMethodsCircuitBreakerLifecycleService`

Azquelt: (Sorry, I missed this when I looked through it the first time).

Ladicek: Hmm, yea, good point. I'll take a look.

Ladicek: Done, thanks!

Ladicek:
Added release notes.

Issue #220, opened 2018-01-25
Add support for exponential backoff retry


cen1:
[Failsafe](https://github.com/jhalterman/failsafe) already has support for [this](http://jodah.net/failsafe/javadoc/net/jodah/failsafe/RetryPolicy.html#withBackoff-long-long-java.util.concurrent.TimeUnit-).

Preferred and less wasteful policy than fixed period when the target service could be under heavy load and unresponsive for longer periods of time.

Emily-Jiang:
sounds useful. +1 from me.

Emily-Jiang:
We can add one more method to configure the backoffFactor with the default value of 2

carlosdlr:
sfsf

brunobat:
+1

antoinesd:
Should be ok for our implementation. So +1 for me

jeanouii:
Very useful. We can start agressive at first and then get slower and slower until we reach a max value and decide it's dead :)

cen1:
I see the pull request is like 20 LOC, what is the holdup for the merge?

Emily-Jiang:
We talked this PR and had some questions. @antoinesd is going to update his PR accordingly. I will try to take this over, if Antoine cannot get around to do it.

lordofthejars:
@Emily-Jiang I am really interested in this feature as well, but instead of focusing on backoff retry. I'd create some kind of interface called `RetryPolicy` or something like that, so users can implement their own strategies, notice that you could try to use the exponential backoff policy but maybe you prefer a Fibonacci one or an incremental one. 

So the API should look like:

```
@Retry(retryPolicy=FibonacciRetryPolicy.class)
```

Ladicek:
When I was looking at this a while ago, I realized that even adding just exponential backoff results in adding not-exactly-intuitive parameters to `@Retry`. Perhaps externalizing the backoff policy into a separate class like @lordofthejars suggests would be an elegant way to solve that, but unfortunately it's less declarative/configuration.

[...]

PR #532, opened 2020-04-14
#220 backoff - draft

Emily-Jiang:


Ladicek:
I'd like to discuss https://github.com/eclipse/microprofile-fault-tolerance/issues/220#issuecomment-560269075, because I think adding exponential backoff directly to `@Retry` leads to overly complex configuration and prevents adding additional backoff strategies.

Ladicek:
Similarly to #565, I'd like to propose that we drop this from 3.0 as well. We have discussed this many times and didn't reach an agreement, so I believe there first needs to be some implementation experience.

PR #536, opened 2020-04-23
Update the spec with the new metrics for 3.0

### LINKED ISSUE #401: Improve MP FT metrics using Metrics tags
With the new functionality of Metrics 2.0 with multiple tag support, this massively simplifies the metrics name, which makes metrics query creation much easier.
We need to revisit our metrics name to see whether we can simplify the names.

### LINKED ISSUE #499: move FT metrics to base scope MetricsRegistry for MP 4.0
Aiming to have all metrics defined in the MP specs be under `base` to avoid potential for name collision we currently have when adding new metrics to `application` scope.

Azquelt:
Includes spec and TCKs.

We can still make tweaks to the spec if required.

Fixes #401 
Fixes #499

### DIFF
--- a/spec/src/main/asciidoc/metrics.asciidoc
+++ b/spec/src/main/asciidoc/metrics.asciidoc
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2018 Contributors to the Eclipse Foundation
+// Copyright (c) 2018-2020 Contributors to the Eclipse Foundation
 //
 // See the NOTICE file(s) distributed with this work for additional
 // information regarding copyright ownership.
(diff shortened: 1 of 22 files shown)


Azquelt:
I've now added new TCKs, though I don't have an implementation to run the TCKs against yet.

Mostly I've just adapted the existing tests to use the new metrics, though the new retry metrics resulted in several new test cases.

Azquelt:
I've now made a few fixes and got my implementation running against these tests.

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -26,164 +26,188 @@ the methods annotated with a `@Retry`, `@Timeout`, `@CircuitBreaker`, `@Bulkhead
...
-| The number of times the method was called and, after all Fault Tolerance actions had been processed, threw a `Throwable`
+| Name | `ft.invocations.total`
+
+| Type | `Counter`
+| Unit | None
+| Description | The number of times the method was called
+| Tags
+a| * `method` - the fully qualified method name
+ * `result` = `[valueReturned\|exceptionThrown]` - whether the invocation returned a value or threw an exception
+ * `fallbackUsed` = `[true\|false\|none]` - `true` if fallback was used, `false` if a fallback is configured but was not used, `none` if a fallback is not configured
```
Emily-Jiang, line 51: Had a further thought on this:
fallbackUsed with the choice of `true|false|none` is odd. How about fallbackStatus - `true|false|none`

Emily-Jiang: I reviewed the whole section and I think it is better to change `none` to `notApplicable`

`fallbackUsed` =`[ true|false|notApplicable]`

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -26,164 +26,188 @@ the methods annotated with a `@Retry`, `@Timeout`, `@CircuitBreaker`, `@Bulkhead
...
+[cols="1,5"]
+|===
+| Name | `ft.circuitbreaker.calls.total`
+
+| Type | `Counter`
+| Unit | None
+| Description | The number of times the timeout logic was run. This will usually be once per method call, but may be more than once if the method call is retried.
+| Tags
+a| * `method` - the fully qualified method name
+ * `circuitBreakerResult` = `[success\|failure\|circuitBreakerOpen]` - the result of the method call, as considered by the circuit breaker logic
```
Emily-Jiang, line 118: Need to add more explaintation on the success criteria.

In more details, the method call is considered to be successful when an method throws an exception but the exception is excluded from the `CircuitBreaker` based on its skipOn and 'failOn` criteria.

spec/src/main/asciidoc/metrics.asciidoc
Ladicek: A fallback that is not applicable is a fallback that has been defined, but configured to not apply for particular situation. It's an interesting case that we don't cover, but it's distinct from fallback that is not defined (which is what `none` means now).

What we have now is, essentially: `fallback=[applied|notApplied|notDefined]`.

We could expand that to `fallback=[applied|notAppliedBecauseNotNeeded|notAppliedBecauseOfConfiguration|notDefined]`. Whether we need to distinguish the two cases of why fallback wasn't applied, I don't know.

spec/src/main/asciidoc/metrics.asciidoc
Ladicek: IMHO, we don't need to repeat that here. That's clearly defined in the circuit breaker chapter.

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -212,43 +236,57 @@ public class MyClass {
...
+ft.timeout.calls.total{method="com.example.MyClass.doWork", timedOut="true"}
+ft.timeout.calls.total{method="com.example.MyClass.doWork", timedOut="false"}
+ft.timeout.executionDuration{method="com.example.MyClass.doWork"}
+```
 
 Now imagine the `doWork()` method is called and the invocation goes like this:
 
 * On the first attempt, the invocation takes more than 1000ms and times out
 * On the second attempt, something goes wrong and the method throws an `IOException`
 * On the third attempt, the method returns successfully and the result of this attempt is returned to the user
```
Emily-Jiang, line 260: Is this attempt from the Retry operation?

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/FallbackMetricBean.java
```diff
@@ -1,6 +1,6 @@
 /*
  *******************************************************************************
- * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ * Copyright (c) 2018-2020 Contributors to the Eclipse Foundation
```
Emily-Jiang, line 3: 2018, 2020

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/FallbackMetricHandler.java
```diff
@@ -1,6 +1,6 @@
 /*
  *******************************************************************************
- * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ * Copyright (c) 2018-2020 Contributors to the Eclipse Foundation
```
Emily-Jiang, line 3: 2018, 2020

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/MetricsDisabledTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ * Copyright (c) 2018-2020 Contributors to the Eclipse Foundation
```
Emily-Jiang, line 2: 2018, 2020

Emily-Jiang:
Can you also add a test scenario when configuring both failOn and skipOn on CircuitBreaker to measure the success counter for an exception excluded by `CircuitBreaker`?

spec/src/main/asciidoc/metrics.asciidoc
Azquelt: Maybe we could just reference back to the section where it is defined?

spec/src/main/asciidoc/metrics.asciidoc
Azquelt: We already know whether it was not applied because it wasn't needed or was not applied because the exception didn't match the configuration by the value of `result`

We currently have
* `result = valueReturned, fallbackUsed = false` - method returned successfully, fallback was not needed
* `result = valueReturned, fallbackUsed = true` - method threw an exception, fallback was invoked and also threw an exception
* `result = exceptionThrown, fallbackUsed = false` - method threw an exception, but fallback was not invoked because fallback configuration didn't include the thrown exception
* `result = exceptionThrown, fallbackUsed = true` - method threw an exception, fallback was invoked and also threw an exception

spec/src/main/asciidoc/metrics.asciidoc
Azquelt: Yes, I'll see if I can reword this to make it clearer that it's doing retries.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/FallbackMetricBean.java
Azquelt: Will do this under #545

Azquelt:
> Can you also add a test scenario when configuring both failOn and skipOn on CircuitBreaker to measure the success counter for an exception excluded by `CircuitBreaker`?

Yes, good spot. We test with an exception which isn't in `failOn` but we don't test with an exception which is in `skipOn`.

spec/src/main/asciidoc/metrics.asciidoc
Azquelt: I've added a cross reference.

spec/src/main/asciidoc/metrics.asciidoc
Azquelt: I've updated this example to make it explicit that it's doing retries.

Azquelt:
I think I've resolved all the comments on this PR, apart from the `fallbackUsed` tag. I'll have a further think on that.

spec/src/main/asciidoc/metrics.asciidoc
Ladicek: Ah, good point, thanks!

Ladicek: What I wanted to say is that `notApplicable` can't be used instead of `none`. It could be `notDefined` or perhaps `undefined`, though `none` is fine in my opinion.

Emily-Jiang: +1 on `notDefined`

Azquelt: I don't like `notDefined` - it sounds like there's something which could have a value but it hasn't been set, which isn't really the situation here.

`notApplicable` makes the most sense idiomatically - the whole question of whether fallback was used or not doesn't apply because there is no `@Fallback` annotation - but it has the downside of overlapping with the `applyOn` attribute and so could be misunderstood to mean "the fallback doesn't apply to this exception" rather than "the question of whether fallback was used or not does not apply".

`notEnabled` would be another option - this overlaps with the ability to disable and enable particular annotations via config, but maybe that's ok? A method where the annotation is disabled is functionally identical to one where the annotation is not present.

Emily-Jiang: `notEnabled` - does not mean the annotation does not exist. It could not the exception does not enable the fallback. Maybe `notSpecified` is better.

Ladicek: I propose we stay with `none`. We use some really excessively long names elsewhere, and I'm not really a big fan of that.

Emily-Jiang: I like the long names you specified @Ladicek , as it is self explainary 👍 .

Azquelt: > `notEnabled` - does not mean the annotation does not exist. It could not the exception does not enable the fallback.

I don't think we ever use the word "enabled" to talk about `applyOn` or `skipOn`, so I don't think it would be reasonable to think that it relates to that. We do use "enabled" to talk about enabling and disabling an annotation via config.

> EDIT: alternatively, I propose we go with long names here as well, and use something like fallback=[applied|notApplied|notDefined] :-)

I still don't really like `notSpecified` or `notDefined` as it sounds to me too much like a value which should exist is missing, but I can live with it. `fallback=notDefined` is at least better than `fallbackUsed=notDefined`.

Ladicek: I can see how `undefined` could sound like "something is supposed to be here, but some programmer made a mistake somewhere and so you're seeing this placeholder" (especially for people who have seen JavaScript), but `notDefined` is pretty unique IMHO.

Azquelt: I've changed it to `fallback=[applied|notApplied|notDefined]`

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -212,43 +236,57 @@ public class MyClass {
...
-* `ft.com.example.MyClass.doWork.retry.callsSucceededNotRetried.total`
-* `ft.com.example.MyClass.doWork.retry.callsSucceededRetried.total`
-* `ft.com.example.MyClass.doWork.retry.callsFailed.total`
-* `ft.com.example.MyClass.doWork.retry.retries.total`
-* `ft.com.example.MyClass.doWork.timeout.executionDuration`
-* `ft.com.example.MyClass.doWork.timeout.callsTimedOut.total`
-* `ft.com.example.MyClass.doWork.timeout.callsNotTimedOut.total`
+```
+ft.invocations.total{method="com.example.MyClass.doWork", result="valueReturned", fallbackUsed="none"}
+ft.invocations.total{method="com.example.MyClass.doWork", result="exceptionThrown", fallbackUsed="none"}
```
Ladicek, line 241: Old name `fallbackUsed`.

spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -212,43 +236,57 @@ public class MyClass {
...
+* The invocation is retried again and this time the method returns successfully and the result of this attempt is returned to the user
 
-`ft.com.example.MyClass.doWork.invocations.total = 1` +
-The method has been called once.
+After this sequence, the following metrics would have new values:
 
-`ft.com.example.MyClass.doWork.invocations.failed = 0` +
-No exceptions were propagated back to the caller.
+```
+ft.invocations.total{method="com.example.MyClass.doWork", result="valueReturned", fallbackUsed="none"} = 1
```
Ladicek, line 265: Old name `fallbackUsed`.

Ladicek:
Quickly skimmed through it and found a few occurences of the old `fallbackUser` name, otherwise LGTM.

spec/src/main/asciidoc/metrics.asciidoc
Azquelt: ah rats, thank you for spotting this, I completely forgot to update the example.

Azquelt:
Squashing all the review changes

PR #551, opened 2020-05-21
fix a couple of typos in the metrics specification

Ladicek:

### DIFF
--- a/spec/src/main/asciidoc/metrics.asciidoc
+++ b/spec/src/main/asciidoc/metrics.asciidoc
@@ -72,7 +72,7 @@ a| * `method` - the fully qualified method name
 
 | Type | `Counter`
 | Unit | None
-| Description | The number of times the method was retried 
+| Description | The number of times the method was retried
 | Tags
 a| * `method` - the fully qualified method name
 |===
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/metrics.asciidoc
```diff
@@ -104,21 +104,19 @@ a| * `method` - the fully qualified method name
 
 === Metrics added for `@CircuitBreaker`
 
-ft.circuitbreaker.calls.total{method=<name>, circuitBreakerResult=[success|failure|circuitBreakerOpen]}
```
Emily-Jiang, line 107: Why did you delete this line?

Ladicek: See how it looks in the rendered form. This line clearly isn't supposed to be there.

Azquelt: Yeah, suspect I copied and pasted directly from the design in the issue as a placeholder before translating it to have the right formatting for the spec. In this case I forgot to take out the original placeholder.

Ladicek:
Merging then, thanks.

PR #552, opened 2020-06-03
move the note about naming/security context propagation to an extra section

Ladicek:
This is because MicroProfile itself doesn't know anything about
naming or security context, and MicroProfile implementations
don't have to have anything like that. These terms only make sense
in a Java EE environment. Therefore, the requirement that naming
and security contexts are propagated should be present in an extra
section that only applies to Java EE container implementations.

### DIFF
--- a/spec/src/main/asciidoc/optional-container-integration.asciidoc
+++ b/spec/src/main/asciidoc/optional-container-integration.asciidoc
@@ -1,0 +1,29 @@
+//
+// Copyright (c) 2020 Contributors to the Eclipse Foundation
+//
+// See the NOTICE file(s) distributed with this work for additional
+// information regarding copyright ownership.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// You may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


spec/src/main/asciidoc/optional-container-integration.asciidoc
```diff
@@ -0,0 +1,29 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// Contributors:
+// Ladislav Thon
+
+[[optional-container-integration]]
+
+== Recommendations for Optional Container Integration
+
+This section describes the expected behaviors when the implementation runs in a Java EE container.
```
Emily-Jiang, line 25: Java EE or Jakarta EE

Ladicek: We refer to Java EE elsewhere in the spec, so I decided to stick to that.

rhusar: The 'Java EE or Jakarta EE' is tangential to this PR.

However, there appears to be only 3 mentions of 'Java EE' though my understanding is all occurrences should be ultimately renamed to Jakarta EE in a subsequent PR.

Ladicek: I agree this should be all renamed to Jakarta EE, but not in this PR.

Ladicek:
I'm not sure this was intended as optional when running in the Java EE environment, and if possible, we should make that more clear.

spec/src/main/asciidoc/optional-container-integration.asciidoc
Ladicek: Filed #556 for this.

PR #553, opened 2020-06-04
Test metrics are registered with correct units

### LINKED ISSUE #549: Add TCKs to assert that metrics have the correct type
The metric type is important because MP Metrics uses it to convert the metric to one of the [base types](https://download.eclipse.org/microprofile/microprofile-metrics-2.3/microprofile-metrics-spec-2.3.html#OpenMetrics_units) when exporting metrics in the OpenMetrics format.

If the types are not specified correctly, then metric queries may not be portable across implementations.

Azquelt:
Fixes #549

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/AllMetricsTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/AllMetricsTest.java
@@ -31,6 +31,8 @@ import static org.eclipse.microprofile.fault.tolerance.tck.metrics.util.MetricDe
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.greaterThan;
 import static org.hamcrest.Matchers.is;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNotNull;
 
 import java.util.concurrent.ExecutionException;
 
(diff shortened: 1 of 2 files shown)


tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/AllMetricsTest.java
```diff
@@ -128,4 +140,19 @@ public void testAllMetrics() throws InterruptedException, ExecutionException {
         assertThat("bulkhead queue wait time histogram present", m.getBulkheadWaitingDuration().isPresent(), is(true));
     }
     
+    @Test
+    public void testMetricUnits() throws InterruptedException, ExecutionException {
+        // Call the method to ensure that all metrics get registered
+        allMetricsBean.doWork().get();
```
Ladicek, line 146: If metrics are registered lazily, this isn't enough to trigger registering all metrics.

I'm not sure if metrics must be registered eagerly, I'll think about that a bit more.

Ladicek: Actually, perhaps my implementation is a bit _too_ lazy -- if no retry ever happens, it doesn't register the `ft.retry.retries.total` metric, even though `ft.retry.calls.total` is registered. That's certainly unexpected. I'll treat it as a bug on my side.

Ladicek: OK, good. Thanks for the test, found a few bugs :-)

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/AllMetricsTest.java
```diff
@@ -79,6 +87,10 @@ public static WebArchive deploy() {
     @Inject
     private AllMetricsBean allMetricsBean;
     
+    @Inject
+    @RegistryType(type = Type.BASE)
```
Emily-Jiang, line 91: This made me think: 
Do we have test to verify the metrics is under base instead of application?

Azquelt: We don't have an explicit test, but reading the Metrics API docs, I thought that the registries were separate and it wouldn't give you an aggregated view. When you inject a `MetricsRegistry` without specifying the type, you get the application registry.

In our tests we specify that we want the base registry, both here and in `MetricGetter`

Azquelt: The only other test we could add would be to ensure that metrics are _not_ registered under the application scope. I'm not sure we should add such a test since it wouldn't

PR #554, opened 2020-06-04
Test configuration of @Retry

Azquelt:
Add TCKs to test that each of the attributes of `@Retry` can be configured.

For #407

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/config/RetryConfigTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/config/RetryConfigTest.java
@@ -1,0 +1,186 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+
+package org.eclipse.microprofile.fault.tolerance.tck.config;
+
+import static org.eclipse.microprofile.fault.tolerance.tck.util.Exceptions.expect;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.lessThan;
+import static org.testng.Assert.assertEquals;
+
+import java.time.Duration;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import javax.inject.Inject;
+
+import org.eclipse.microprofile.fault.tolerance.tck.util.Packages;
+import org.eclipse.microprofile.fault.tolerance.tck.util.TestException;
+import org.eclipse.microprofile.faulttolerance.Retry;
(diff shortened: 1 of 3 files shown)


tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/config/RetryConfigTest.java
```diff
@@ -0,0 +1,186 @@
...
+                .set(RetryConfigBean.class, "serviceAbortOn", Retry.class, "abortOn",
+                     TestConfigExceptionA.class.getName() + "," + TestConfigExceptionB1.class.getName())
+                .set(RetryConfigBean.class, "serviceJitter", Retry.class, "jitter", "1")
+                .set(RetryConfigBean.class, "serviceJitter", Retry.class, "jitterDelayUnit", "SECONDS");
+        
+        JavaArchive jar = ShrinkWrap.create(JavaArchive.class, "ftRetryConfig.jar")
+                .addClasses(RetryConfigBean.class)
+                .addClasses(TestConfigExceptionA.class, TestConfigExceptionB.class, TestConfigExceptionB1.class)
+                .addPackage(Packages.UTILS)
+                .addAsManifestResource(config, "microprofile-config.properties");
```
Ladicek, line 67: Please add

```java
                .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml");
```

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/config/RetryConfigTest.java
```diff
@@ -0,0 +1,186 @@
...
+                .set(RetryConfigBean.class, "serviceMaxDuration", Retry.class, "maxDuration", "1")
+                .set(RetryConfigBean.class, "serviceMaxDuration", Retry.class, "durationUnit", "SECONDS")
+                .set(RetryConfigBean.class, "serviceDelay", Retry.class, "delay", "2000")
+                .set(RetryConfigBean.class, "serviceDelay", Retry.class, "delayUnit", "MICROS")
+                .set(RetryConfigBean.class, "serviceRetryOn", Retry.class, "retryOn",
+                     TestConfigExceptionA.class.getName() + "," + TestConfigExceptionB.class.getName())
+                .set(RetryConfigBean.class, "serviceAbortOn", Retry.class, "abortOn",
+                     TestConfigExceptionA.class.getName() + "," + TestConfigExceptionB1.class.getName())
+                .set(RetryConfigBean.class, "serviceJitter", Retry.class, "jitter", "1")
+                .set(RetryConfigBean.class, "serviceJitter", Retry.class, "jitterDelayUnit", "SECONDS");
```
Ladicek, line 61: I'm also wondering if we need to scale the values here with `TCKConfig`. Not sure about retry, but for timeout, I'm probably gonna do that.

Azquelt: I don't think it's necessary in this case.

At the moment the timed tests are set up so that if the config isn't applied they take 10 seconds and if it is applied, they take less than one second. The tests check that execution takes less than 8 seconds, so there's a fairly large margin of error there already.

The bigger issue is that if I wanted to increase the margin of error, then I need to increase the values on the annotation so that if config isn't applied then method execution takes longer than 10 seconds. The way we've done this before is to use config, but that would undermine what we're trying to test in this class.

Note that these tests are much looser than the general tests we have for `@Retry`. For example I'm not trying to determine whether `maxDuration` is implemented correctly, I'm assuming that is is implemented correctly and just checking whether or not it's picking up values from config correctly. So rather than trying to check that method execution takes approximately the expected time, I'm just trying to distinguish between it taking 1 second or taking 10 seconds.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/config/RetryConfigTest.java
Azquelt: Done, thanks for catching that.

PR #555, opened 2020-06-05
add TCK tests for configuring @Timeout attributes

Ladicek:

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/config/TimeoutConfigBean.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/config/TimeoutConfigBean.java
@@ -1,0 +1,53 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+
+package org.eclipse.microprofile.fault.tolerance.tck.config;
+
+import org.eclipse.microprofile.faulttolerance.Asynchronous;
+import org.eclipse.microprofile.faulttolerance.Timeout;
+
+import javax.enterprise.context.ApplicationScoped;
+import java.time.temporal.ChronoUnit;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Suite of methods for testing the various parameters of Timeout
+ */
+@ApplicationScoped
+public class TimeoutConfigBean {
+    @Timeout(value = 1, unit = ChronoUnit.MILLIS)
+    public void serviceValue() throws InterruptedException {
+        Thread.sleep(TimeUnit.MINUTES.toMillis(1));
+    }
+
+    @Timeout(value = 1000, unit = ChronoUnit.MICROS)
+    public void serviceUnit() throws InterruptedException {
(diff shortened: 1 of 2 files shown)


Ladicek:
Contributes to #407.

Ladicek:
Also I shamelessly copied #554 here, so thanks @Azquelt :-)

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/config/TimeoutConfigBean.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Timeout(value = 1000, unit = ChronoUnit.SECONDS)
+    public void serviceUnit() throws InterruptedException {
+        Thread.sleep(TimeUnit.DAYS.toMillis(1));
+    }
+
+    @Timeout(value = 1, unit = ChronoUnit.MINUTES)
+    public void serviceBoth() throws InterruptedException {
+        Thread.sleep(TimeUnit.DAYS.toMillis(1));
+    }
```
Azquelt, line 47: I don't like that if this test were to fail, it could hang for over 15 minutes if it wasn't reading the config correctly, or for several days if `@Timeout` wasn't implemented at all.

Given that you're targeting it timing out after 1s in the passing case (allowing times between 0.8 and 2.0 and applying the config multiplier), maybe it would make sense to set the attributes in the code to very short values (so that they finish too quickly if config isn't read) and then reduce the sleep timeout to something like 1 minute.

You could use:
* `serviceValue` - annotation `1ms`, config `1000ms`
* `serviceUnit` - annotation `1000µs`, config `1000ms` (use 1000ms rather than 1 second to allow the value to be scaled effectively by config)
* `serviceBoth` - annotation `10µs`, config `1000ms`

Ladicek: That would also work, and is much nicer. Thanks!

Ladicek: Done.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/config/TimeoutConfigBean.java
```diff
@@ -0,0 +1,48 @@
...
+import javax.enterprise.context.ApplicationScoped;
+import java.time.temporal.ChronoUnit;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Suite of methods for testing the various parameters of Timeout
+ */
+@ApplicationScoped
+public class TimeoutConfigBean {
+    @Timeout(value = 1, unit = ChronoUnit.MILLIS)
```
Emily-Jiang, line 34: Since `@Timeout` has a lot of advantages when used together with `@Asynchronous`, I am wondering add `@Asynchronous` on one of the endpoint might be good. Thoughts?

Ladicek: In the config test?

Azquelt: We're only testing timeout config here. Do you think we need a specific test for config when `@Timeout` and `@Asynchronous` are used together?

In general I prefer tests to remain focused. This class just tests Timeout + config while we have other tests for Timeout + async.

Emily-Jiang: Do we have Timeout + async + config? The reason is that timeout is better off to be used with `@Asynchronous`. We need to ensure the config works. I will be ok to skip `Timeout` with config and vote for `Timeout`+`Asynchronous` with config.

Ladicek: > timeout is better off to be used with @Asynchronous

Why is that?

> We need to ensure the config works

We do. This PR does that. `@Asynchronous` makes no difference whatsoever.

Ladicek: (Actually it might make a difference, depending on implementation, but so can putting the annotation on a class, or configuring an annotation parameter for all methods in a class, or configuring an annotation parameter globally. We decided to stick to this form of testing and I don't quite see how `@Asynchronous` is so very relevant.)

Ladicek: In any case, I have just changed one test from synchronous to `@Asynchronous`.

PR #558, opened 2020-06-10
Add tests for configuration of Bulkhead

Azquelt:
For #407


tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/config/BulkheadConfigBean.java
```diff
@@ -0,0 +1,76 @@
...
+    
+    @Bulkhead(value = 1, waitingTaskQueue = 5)
+    @Asynchronous
+    public Future<Void> serviceWaitingTaskQueue(Future<?> waitingFuture) {
+        tasksRunning.incrementAndGet();
+        try {
+            waitingFuture.get(1, TimeUnit.MINUTES);
+            return CompletableFuture.completedFuture(null);
+        }
+        catch (InterruptedException | ExecutionException | TimeoutException e) {
```
Ladicek, line 62: Any reason to not just let these exceptions bubble up, especially since the method is `@Asynchronous`?

PR #562, opened 2020-06-18
Improve the bulkhead tests

### LINKED ISSUE #271: Improve speed and perhaps reliability of the Bulkhead TCK tests.

Azquelt:
Fixes #271 

This PR rethinks how we do the main `Bulkhead` tests

Previously we made lots of calls to the bulkhead method and had each one sleep for a while. We would then verify that the bulkhead did not exceed its limits.

In the new tests, we use `Barrier` wherever possible to allow us to fill up the bulkhead so that we can test its behaviour when we try to exceed its limits. This means we can spend less time waiting for tasks to run and complete.

However, I do note that the old tests used `Bulkhead` in a way which was fairly similar to the way we expect it to be used by users, whereas the new tests use very artificial scenarios to more easily exercise corner cases. To remedy this, I've added `BulkheadPressureTest` which just does two tests (sync and async) in the old style, throwing lots of work at a bulkhead and checking it doesn't exceed its limits.

Rewriting the bulkhead tests allowed me to remove lots of framework code which was written to support them. However, there were a few other tests which also used some of this framework code so they needed to be updated too. This creates quite long and noisy file diff, so I've broken it down into several different commits to make it easier to read.

Some local testing suggests that this does speed up running the bulkhead tests significantly (reducing the TCK runtime from about 5 mins to under 3 mins on the smallrye implementation).

The slowest parts of the test are now where we assert that something does _not_ happen (e.g. a task does not await a barrier, a Future does not complete). Currently the tests wait 0.5 seconds to check that the thing doesn't happen (with this value scaled by the timeout multiplier). We could reduce this value at the risk of having tests occasionally pass when they should not.


Azquelt:
@Ladicek When running this against smallrye, I saw some of the bulkhead tests using barriers fail sometimes. In one case it was failing to await on the barrier within 3s, so I'm not sure if that's a fundamental issue with the test, a problem with the impl or if I should just increase the timeouts on the tests (or use the multiplier).

I was running locally on my laptop so I was surprised to hit issues with things not happening as quickly as expected.

Ladicek:
I'll take a look.

Ladicek:
I mean, I _will_ take a look, but not today, too many things going on already :-) Thanks, though!

Ladicek:
So I found 2 issues when running this:

1. Our (SmallRye Fault Tolerance) implementation of retry is suboptimal in when we delay and when we don't (specifically, we sometimes delay even if it isn't necessary). This made a couple of tests from this PR fail. I recognize this as something to be improved on our side, but the MicroProfile Fault Tolerance specification doesn't really define how delays between retry attempts should work (neither does the TCK). So this is a bit of a grey area IMHO.

2. The `BulkheadAsynchTest.testBulkhead` method has a little bug in how it handles bulkhead queues. Specifically, it has this:

    ```java
    List<BarrierTask<?>> queuedTasks = new ArrayList<>();
    for (int i = 0; i < maxQueued; i++) {
        BarrierTask<?> task = taskManager.runAsyncBarrierTask(bulkheadMethod);
        queuedTasks.add(task);
    }
    ```

    And subsequent code then expects that the `queuedTasks` has the same order as the bulkhead queue (`queuedTasks.get(0).assertAwaits();`), which doesn't always have to be the case. We would have to have a way how to await until the task is put into the bulkhead queue, and only then submit another task.

When I fixed these 2 issues, the entire TCK passes for me (tried several times). Specifically, I fixed the 1st issue on our side, and for the 2nd, I replaced

```java
// Check the first queued task now starts
queuedTasks.get(0).assertAwaits();
```

with this:

```java
// Check that one queued task now starts
await().atMost(3, TimeUnit.SECONDS).untilAsserted(() -> {
    int count = 0;
    for (BarrierTask<?> task : queuedTasks) {
        if (task.waitingOnBarrier()) { // had to add this method, but it's pretty simple
            count++;
        }
    }
    assertEquals(count, 1, "Exactly 1 queued task was expected to start running");
});
```

which I believe is more correct.

WDYT?

Ladicek:
Just to be sure, I've made this modification:

```java
// Fill the queue
List<BarrierTask<?>> queuedTasks = new ArrayList<>();
for (int i = 0; i < maxQueued; i++) {
    BarrierTask<?> task = taskManager.runAsyncBarrierTask(bulkheadMethod);
    queuedTasks.add(task);
    try {
        Thread.sleep(500);
    } catch (InterruptedException e) {
        e.printStackTrace();
    }
}
```

With this change, all the tests pass in the original form (i.e., using `queuedTasks.get(0).assertAwaits();`).

Obviously this makes the tests run much longer, and the proper way would be to somehow wait for tasks entering the bulkhead queue. The only way how to do that that I can think of is metrics.

I tried to use them by adding a `GaugeMetric bulkheadWaitingCount` parameter to the `testBulkhead` method and modifying it like this:

```java
List<BarrierTask<?>> queuedTasks = new ArrayList<>();
for (int i = 0; i < maxQueued; i++) {
    BarrierTask<?> task = taskManager.runAsyncBarrierTask(bulkheadMethod);
    queuedTasks.add(task);

    int expectedTasksInQueue = i + 1;
    await().until(() -> bulkheadWaitingCount.value() == expectedTasksInQueue);
}
```

With this, the tests also pass for me.

Ideas?

Azquelt:
So, just checking I have this right, the issue is that when we add tasks in a tight loop, they don't necessarily join the bulkhead queue in the same order, so it's not always the case that the first task added to the queue is the first one to run?

Ladicek:
Exactly.

Ladicek:
I mean -- "the issue is that when we add tasks in a tight loop, they don't necessarily join the bulkhead queue in the same order", yes, that's the cause of the issue here. Moreover, it is always the case that "the first task added to the queue is the first one to run", but the order of tasks being added to the queue is not the same as the order of tasks being thrown at the bulkhead -- if they are thrown at it concurrently. I've proposed an approach how to serialize that.

Azquelt:
Hmm, the test isn't calling the test methods concurrently, `taskManager.runAsyncBarrierTask` doesn't use `AsyncCaller` or anything like that so if there's any concurrency to the way that tasks are thrown at the bulkhead, it's coming from your implementation, though I don't think the spec forbids that.

I don't really like using the metrics here since it makes it impossible to test that you have your bulkhead working until you've also added metrics which isn't very nice, though I also don't like having a forced wait there either.

I'll have a think further on this, maybe I can redo some of these tests so that they don't rely on the exact ordering. E.g. if the queue size is `5`, we could say "check that if I try to run six tasks, exactly one is rejected" without requiring that it's the last one that's rejected.

It might need a little more framework code to make constraints like that easy to express.

Azquelt:
Re-reading again, I think this is close to your first suggested change.

Ladicek:
In the tests that I've seen failing, the tasks themselves were `@Asynchronous` methods, so there's definitely some concurrency involved.

Azquelt:
Right, that's what I meant by your implementation providing any concurrency, since it implements `@Asynchronous`

The test calls the method multiple times on a single thread and I don't think there's anything in the spec that _prevents_ you from ensuring that tasks are queued onto the bulkhead in the order that their method calls are made. There's nothing that explicitly requires that either, so I'm happy to relax the test.

I've gone with your first suggested approach (asserting that any queued execution starts). We don't give the user any visibility over when tasks are added to the queue and it doesn't seem like there's anything they can do to reliably ensure that one task is added to the bulkhead before another, so I don't think it makes sense to try to test it.

Ladicek:
Ah, right. Indeed we put tasks into an executor pretty early in some cases, and I'm pretty sure the spec allows that (as you also say). The test change looks good to me.

Azquelt:
super unhelpfully, github decided to change the commit author when I clicked squash-and-merge on Ladicek's helpful PR Azquelt/microprofile-fault-tolerance#1 which is why the eca bot is unhappy.

I'll clean it up by cherry-picking the original commit from that PR.

Azquelt:
All better!

Ladicek:
Sorry for that :-)

PR #571, opened 2020-08-12
Support running against both MP Metrics 2.3 and 3.0

### LINKED ISSUE #570: TCK cannot be run using MP Metrics 3.0
MP Metrics 3.0 has several breaking changes which prevent the TCK from running:
* `MetricsRegistry` changed from an abstract class to an interface (this breaks all method calls to `MetricRegistry` since the bytecode for calling an interface method is different to calling a class method)
* The method for getting the unit as an `Optional<String>` from `Metadata` has been renamed from `getUnit` to `unit`

Since we expect Fault Tolerance 3.0 to be included with Metrics 3.0 in Microprofile 4.0, this is not ideal.

I think we need to do one of the following:
* specify that only integration with MP Metrics 3.0 is required and update the tests for the Metrics 3.0 API
* split off the metrics tests into two separate modules, one for MP Metrics 2.x and another for MP Metrics 3.0
* use reflection so that the same test code can run against either MP Metrics 2.x or MP Metrics 3.0

Azquelt:
This PR uses reflection to work around the differences in API so that the TCK can run with either MP Metrics 2.3 or 3.0.

Most of the reflective nastiness is encapsulated in `MetricRegistryProxy` and `MetricRegistryProxyHandler` leaving the test code fairly clean.

Fixes #570

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/AllMetricsTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/AllMetricsTest.java
@@ -39,15 +39,15 @@ import java.util.concurrent.ExecutionException;
 import javax.inject.Inject;
 
 import org.eclipse.microprofile.fault.tolerance.tck.config.ConfigAnnotationAsset;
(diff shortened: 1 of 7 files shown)


Azquelt:
I've tested this locally against OpenLiberty integrating with Metrics 2.3 and Metrics 3.0.

Issue #574, opened 2020-11-17
Release MicroProfile Fault Tolerance under MPWG guideline


Emily-Jiang:
Some required artifacts and copyright need to be produced and packaged. See more required details [here](https://github.com/microprofile/microprofile-wg/issues/6).

PR #575, opened 2020-11-17
Prepare for MP FT 2.0 release under MPWG

Emily-Jiang:
Fixes: #574 


Ladicek (approved):
I don't really understand what this change does, but I trust it's OK :-)

PR #578, opened 2020-11-24
improve the way to handle copyright year

Emily-Jiang:

### DIFF
--- a/spec/pom.xml
+++ b/spec/pom.xml
@@ -84,6 +84,8 @@
                         <revnumber>${project.version}</revnumber>
                         <revremark>${revremark}</revremark>
                         <revdate>${revisiondate}</revdate>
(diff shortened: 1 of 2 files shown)


Ladicek (approved):

Azquelt:
What does this do and why does the pom.xml have `<license>Apache License v2.0</license>` whereas the placeholders are in a file containing the Eclipse Foundation Specification License?

I notice there's also an aslv2 license file in this directory which is a bit confusing - it seems to just be referenced from the site.yaml in the root?

Emily-Jiang:
@Azquelt only the generated pdf/html uses Eclipse Foundation license. The source is still Apache license. For javadoc, only the final released version uses EF license while the RCx uses Apache license. The jars are still using Apache Licenses.

PR #582, opened 2020-12-03
add test that all possible metric combinations are registered

Ladicek:


Ladicek:
CC @jbee

jbee (approved):

Ladicek:
This clearly needs more thinking (e.g. it expects a metric with `fallback=notDefined` to be present for methods that _do_ define `@Fallback`), and I don't think we should rush it into 3.0. Converted this PR to a draft for now.

PR #596, opened 2021-09-22
drop one TCK test that uses beans.xml

Ladicek:
I believe this test should be dropped for the following reasons.

1. It doesn't test anything related to MicroProfile Fault Tolerance,
   it only verifies CDI behavior. Such test belongs to the CDI TCK,
   not to MicroProfile Fault Tolerance TCK. (And I'm sure the CDI TCK
   tests interceptor enablement and ordering pretty extensively.)
2. The upcoming CDI Lite specification only includes interceptor
   enablement using `@Priority` and ignores the `beans.xml` content
   completely (with the exception of `bean-discovery-mode`).
   In my opinion, MicroProfile Fault Tolerance implementations
   should be able to pass the TCK even when running in a container
   that only implements CDI Lite.


Emily-Jiang (approved):
LGTM

PR #599, opened 2021-10-08
Corrected Interceptor Specification URL

hbelmiro:
The Interceptor spec URL was incorrect.


Ladicek:
I think the `master` branch is now aligned with Jakarta EE 9, so maybe we should point to https://jakarta.ee/specifications/interceptors/2.0/interceptors-spec-2.0.html?

PR #611, opened 2022-10-03
Tolerate both MP Metrics 4.0 and 5.0 APIs

Azquelt:
Tolerate the following breaking changes in MP Metrics 5.0, while retaining support for MP Metrics 4.0:

- Methods removed from `Snapshot`
- `Gauge<T>` changed to `Gauge<T extends Number>`
  - Breaks binary compatibility because method signatures are different
    after type erasure
- `MetricsRegistry` injected with `@RegistryScope` instead of `@RegistryType`


Azquelt:
For some reason the github action builds are not using the latest commit. I'll force push again to see if that fixes it.

Azquelt:
MP Metrics 5.0 is going to deprecate and retain `@RegistryType`, so the changes in #612 are sufficient.

PR #612, opened 2022-10-03
Tolerate MP Metrics 5.0 API changes, except MetricsRegistry injection

### LINKED ISSUE #609: TCK does not work with MP Metrics 5.0
MP Metrics 5.0 is bringing in [breaking changes][breaking-spec-50] which require changes in the Fault Tolerance TCK.

The ones that I think affect the Fault Tolerance TCK are:
* `@RegistryType` is removed, replaced with `@RegistryScope`
  * `@RegistryScope` is not a qualifier which means we can't look it up with `CDI.current().select()`
  * The only way to get hold of the `MetricsRegistry` for the base scope is to inject it into a bean with the `@RegistryScope` annotation
* Some methods have been removed from `Snapshot`, particularly `getValues()` and `getMedian()`
  * The median can likely be obtained from the new `getPercentileValues()` array
  * There's no replacement for `getValues()` - several of our tests used this to assert that the correct data had been added to the histogram
* `Gauge<T>` has been changed to `Gauge<T extends Number>` which breaks binary compatibility because it changes the signatures of methods which return `T`

I don't think there are any changes needed to the spec or API to accomodate MP Metrics 5.0, just to the TCK.

[breaking-spec-50]: https://download.eclipse.org/microprofile/microprofile-metrics-5.0.0-RC2/microprofile-metrics-spec-5.0.0-RC2.html#_breaking_changes

Azquelt:
Tolerate the following breaking changes in MP Metrics 5.0, while retaining support for MP Metrics 4.0:

- Methods removed from `Snapshot`
- `Gauge<T>` changed to `Gauge<T extends Number>`
  - Breaks binary compatibility because method signatures are different after type erasure

This does not include support for tolerating the removal of `@RegistryType`. See #117

Fixes #609

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/BulkheadMetricTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/BulkheadMetricTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018-2020 Contributors to the Eclipse Foundation
+ * Copyright (c) 2018-2022 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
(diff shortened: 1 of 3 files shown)


Azquelt:
With the merge of https://github.com/eclipse/microprofile-metrics/pull/748, these are the only changes which are needed to make the TCK compatible with MP Metrics 5.0.

I've now run this locally on my workstation with openliberty, incorporating this smallrye-metrics change: https://github.com/smallrye/smallrye-metrics/pull/558

Emily-Jiang (approved):
LGTM! Thanks @Azquelt

PR #614, opened 2023-01-24
Weld cannot find beans because of missing annotation

jbescos:
Relates to this https://github.com/helidon-io/helidon/pull/5830

Weld requires some annotations to know the beans

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/config/FallbackHandlerA.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/config/FallbackHandlerA.java
@@ -1,6 +1,6 @@
 /*
  *******************************************************************************
- * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ * Copyright (c) 2020, 2023 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
@@ -23,6 +23,9 @@ package org.eclipse.microprofile.fault.tolerance.tck.config;
 import org.eclipse.microprofile.faulttolerance.ExecutionContext;
 import org.eclipse.microprofile.faulttolerance.FallbackHandler;
 
+import jakarta.enterprise.context.Dependent;
+
+@Dependent
 public class FallbackHandlerA implements FallbackHandler<String> {
 
     @Override
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/config/FallbackHandlerB.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/config/FallbackHandlerB.java
@@ -1,6 +1,6 @@
 /*
  *******************************************************************************
- * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ * Copyright (c) 2020, 2023 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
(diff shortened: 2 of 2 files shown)


Ladicek:
I vaguely recall this was intentional. It is not directly captured in the specification text that fallback handlers do not have to be beans, but it is mentioned in the API (though the intent of the text is not very clear to me):

- https://github.com/eclipse/microprofile-fault-tolerance/blob/master/api/src/main/java/org/eclipse/microprofile/faulttolerance/Fallback.java#L79-L80
- https://github.com/eclipse/microprofile-fault-tolerance/blob/master/api/src/main/java/org/eclipse/microprofile/faulttolerance/FallbackHandler.java#L21-L22

In CDI Full, this is easy to achieve using the `Unmanaged` API, similarly to: https://github.com/smallrye/smallrye-fault-tolerance/blob/main/implementation/fault-tolerance/src/main/java/io/smallrye/faulttolerance/DefaultFallbackHandlerProvider.java. In CDI Lite, a container-specific API to make the bean `@Dependent` is probably best.

jbescos:
To be honest I don't know much details about this, but Weld 5.x is failing to find the bean because it is not annotated.

There was a similar issue in microprofile-config: https://github.com/eclipse/microprofile-config/issues/742

It looks this is required for CDI 4 (EE10).

Ladicek:
Perhaps there's an empty `beans.xml` in the test deployment in this case? That has indeed changed meaning: it previously meant bean discovery of `all`, while it now means bean discovery of `annotated`. Sounds like your implementation relies on fallback handlers being beans. My understanding is that the MP Fault Tolerance specification doesn't require that -- but I agree the specification is not very clear on that.

jbescos:
There is no beans.xml at all. In a first view it looks Weld 5 is behaving as an empty beans.xml when this file is not provided.

But adding the next beans.xml in the classpath of the tests seems to have no effect:
```
<beans xmlns="https://jakarta.ee/xml/ns/jakartaee"
       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
       xsi:schemaLocation="https://jakarta.ee/xml/ns/jakartaee
                           https://jakarta.ee/xml/ns/jakartaee/beans_4_0.xsd"
       version="4.0"
       bean-discovery-mode="all">
</beans>
```

I am going to investigate it further.

Ladicek:
I can only repeat what I already said: Sounds like your implementation relies on fallback handlers being beans. My understanding is that the MP Fault Tolerance specification doesn't require that -- but I agree the specification is not very clear on that.

Azquelt:
> According to this, the test is going to fail always with CDI 4.0, because the discovery is going to be annotated, instead of all.

This does not fail on Open Liberty with CDI-4.0 and our Fault Tolerance implementation.

I think @Ladicek is correct, a fallback handler does not need to be a bean and you have a bug in your Fault Tolerance implementation.

That said I don't think it was our intention to change the meaning of this TCK test when run on CDI 4.0, we just didn't notice because the test is still within spec.

I'm minded to accept this change to retain the original intent of this test, but raise an issue to add a new test where the fallback handler is not a bean, since that's a testing gap we have in the TCK. We could perhaps add a note in the spec to be clearer that the fallback handler does not have to be a bean. What do you think @Ladicek?

However, whether this change is accepted or not, you still have a bug. Any user who runs into this can point to the spec and show that your behaviour is incorrect, regardless of whether or not it's sufficient to pass the TCK.

Ladicek:
I agree we should have tests for both situations: when the fallback handler is a bean, and when it is not. I also agree the specification text should be more clear on this. The javadoc mentions I linked above are the only places I could locate that speak about this topic.

jbescos:
Okay, I will review it.

But this rises some questions:
How is that handler supposed to be instanced?, with reflection using an empty constructor?. And which scope?, singleton?.

Ladicek:
As I mentioned above, if using Weld, which is a CDI Full implementation, you can use the `Unmanaged` API. Here's how we do it in SmallRye Fault Tolerance: https://github.com/smallrye/smallrye-fault-tolerance/blob/main/implementation/fault-tolerance/src/main/java/io/smallrye/faulttolerance/DefaultFallbackHandlerProvider.java

jbescos:
Feel free to close this if it is not required.

PR #619, opened 2023-11-13
This PR replaces all expected RuntimeExceptions with TestException. T…

### LINKED ISSUE #620: The use of expected RuntimeExceptions in the TCK can cause false positives
In our build system I noticed some tests failing with a RuntimeException. This is concerning because if our build system creates unwanted runtime exceptions it might create them inside the try blocks where a RuntimeException is expected.

To fix this all expected RuntimeExceptions should be replaced with TestException.

benjamin-confino:
…his is because I saw some genuine RuntimeExceptions causing failures in our build system and if runtime exceptions can appear where they are not expected the wrong runtime exception may appear where a runtime exception is expected

Where a runtime exception is not expected it also updates the error message to include the details of that exception

fixes #620

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsyncTimeoutTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsyncTimeoutTest.java
@@ -34,6 +34,7 @@ import org.eclipse.microprofile.fault.tolerance.tck.asynctimeout.clientserver.As
 import org.eclipse.microprofile.fault.tolerance.tck.asynctimeout.clientserver.AsyncTimeoutClient;
 import org.eclipse.microprofile.fault.tolerance.tck.config.ConfigAnnotationAsset;
 import org.eclipse.microprofile.fault.tolerance.tck.util.Connection;
+import org.eclipse.microprofile.fault.tolerance.tck.util.TestException;
 import org.eclipse.microprofile.faulttolerance.Timeout;
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.testng.Arquillian;
@@ -78,7 +79,8 @@ public class AsyncTimeoutTest extends Arquillian {
 
         JavaArchive testJar = ShrinkWrap
                 .create(JavaArchive.class, "ftAsyncTimeout.jar")
-                .addClasses(AsyncTimeoutClient.class, AsyncClassLevelTimeoutClient.class, Connection.class)
+                .addClasses(AsyncTimeoutClient.class, AsyncClassLevelTimeoutClient.class, Connection.class,
+                        TestException.class)
                 .addAsManifestResource(config, "microprofile-config.properties")
                 .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml")
                 .as(JavaArchive.class);
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsyncTimeoutTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsyncTimeoutTest.java
@@ -34,7 +34,6 @@ import org.eclipse.microprofile.fault.tolerance.tck.asynctimeout.clientserver.As
 import org.eclipse.microprofile.fault.tolerance.tck.asynctimeout.clientserver.AsyncTimeoutClient;
 import org.eclipse.microprofile.fault.tolerance.tck.config.ConfigAnnotationAsset;
 import org.eclipse.microprofile.fault.tolerance.tck.util.Connection;
-import org.eclipse.microprofile.fault.tolerance.tck.util.TestException;
 import org.eclipse.microprofile.faulttolerance.Timeout;
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.testng.Arquillian;
@@ -79,8 +78,7 @@ public class AsyncTimeoutTest extends Arquillian {
 
         JavaArchive testJar = ShrinkWrap
                 .create(JavaArchive.class, "ftAsyncTimeout.jar")
-                .addClasses(AsyncTimeoutClient.class, AsyncClassLevelTimeoutClient.class, Connection.class,
-                        TestException.class)
+                .addClasses(AsyncTimeoutClient.class, AsyncClassLevelTimeoutClient.class, Connection.class)
                 .addAsManifestResource(config, "microprofile-config.properties")
                 .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml")
                 .as(JavaArchive.class);
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/FallbackTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/FallbackTest.java
@@ -83,7 +83,7 @@ public class FallbackTest extends Arquillian {
             Assert.assertTrue(result.contains("serviceA"),
                     "The message should be \"fallback for serviceA\"");
         } catch (TestException ex) {
-            Assert.fail("serviceA should not throw a RuntimeException in testFallbackSuccess", ex);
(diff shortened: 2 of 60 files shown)


Azquelt:
I think instances like this should also be updated: https://github.com/eclipse/microprofile-fault-tolerance/pull/619/files#diff-75e65464c0fe1fb4456a2b1d045d03b43dbc7ed637603883d5f8292dccdd5b07L102

This test should expect a TestException on iteration 1,2,3,7,8 and 9. It should fail with a different error (in the surrounding catch block) if another exception is thrown.

Emily-Jiang (approved):
LGTM

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/AsyncTimeoutTest.java
```diff
@@ -78,7 +79,8 @@ public static WebArchive deploy() {
 
         JavaArchive testJar = ShrinkWrap
                 .create(JavaArchive.class, "ftAsyncTimeout.jar")
-                .addClasses(AsyncTimeoutClient.class, AsyncClassLevelTimeoutClient.class, Connection.class)
+                .addClasses(AsyncTimeoutClient.class, AsyncClassLevelTimeoutClient.class, Connection.class,
+                        TestException.class)
```
Azquelt, line 83: Unsure why `TestException` is needed in this app, none of the other classes in the app appear to reference it.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/FallbackTest.java
```diff
@@ -81,17 +82,17 @@ public void testFallbackSuccess() {
             String result = fallbackClient.serviceA();
             Assert.assertTrue(result.contains("serviceA"),
                     "The message should be \"fallback for serviceA\"");
-        } catch (RuntimeException ex) {
-            Assert.fail("serviceA should not throw a RuntimeException in testFallbackSuccess");
+        } catch (TestException ex) {
+            Assert.fail("serviceA should not throw a RuntimeException in testFallbackSuccess", ex);
```
Azquelt, line 86:
```suggestion
        } catch (TestException ex) {
            Assert.fail("serviceA should not throw a TestException in testFallbackSuccess", ex);
```

Azquelt: Similar messages throughout this class should also change.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/FallbackTest.java
```diff
@@ -81,17 +82,17 @@ public void testFallbackSuccess() {
...
         Assert.assertEquals(fallbackClient.getCounterForInvokingServiceA(), 2,
                 "The execution count should be 2 (1 retry + 1)");
         try {
             String result = fallbackClient.serviceB();
             Assert.assertTrue(result.contains("serviceB"),
                     "The message should be \"fallback for serviceB\"");
-        } catch (RuntimeException ex) {
-            Assert.fail("serviceB should not throw a RuntimeException in testFallbackSuccess");
+        } catch (TestException ex) {
+            Assert.fail("serviceB should not throw a RuntimeException in testFallbackSuccess", ex);
```
Azquelt, line 95: The changes to the `catch` blocks in this test are good, but the following methods in other classes should be updated to match:

- `FallbackClient.nameService()`
- `FallbackOnlyClient.nameService()`
- `FallbackWithBeanClient.nameService()`

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/retry/clientserver/RetryClassLevelClientAbortOn.java
```diff
@@ -63,7 +64,7 @@ private void writingService() {
         counterForInvokingWritingService++;
         try {
             Thread.sleep(100);
-            throw new RuntimeException("WritingService failed");
+            throw new TestException("WritingService failed");
```
Azquelt, line 67: `serviceB` (just above this change in the file), should be updated to abortOn a `TestException` rather than a `RuntimeException`.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/RetryConditionTest.java
```diff
@@ -62,6 +63,9 @@
  */
 public class RetryConditionTest extends Arquillian {
 
+    private final static String SIMULATED_EXCEPTION_MESSAGE = "Simulated error";
+    private final static String SIMULATED_RUNTIME_EXCEPTION_MESSAGE = "Test Exception - Simulated error";
```
Azquelt, line 67: Changing these strings to constants is fine, but these strings are also used in `AsyncRetryClient`.

Please either leave the strings in the tests, or update `AsyncRetryClient` to use the same constants. Using the constants in one place but not the other gives you none of the benefits of using a constant and makes it harder to see that both parts of the test are using the same string.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/visibility/retry/RetryVisibilityTest.java
```diff
@@ -272,11 +273,11 @@ private void checkServiceCall(int nbExpectedRetries, RetryService service, Strin
                             RetryVisibilityTest.class.getSimpleName(),
                             testName,
                             expectedNbCalls));
-        } catch (RuntimeException ex) {
+        } catch (TestException ex) {
```
Azquelt, line 276: Most changes in this class should be undone.

This class is using `IOException` as its expected exception - similar to the way we're using `TestException` in other classes.

If anything, this catch block could change from `RuntimeException` to just `Exception` - it's just adding more information about unexpected thrown exceptions.

You _could_ change `IOException` to `TestException` in this class and in `BaseRetryOnClassService`.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/visibility/retry/RetryVisibilityTest.java
```diff
@@ -272,11 +273,11 @@ private void checkServiceCall(int nbExpectedRetries, RetryService service, Strin
...
             Assert.fail(String.format("no %s exception should have been thrown in %s#%s",
                     ex.getClass().getName(),
                     RetryVisibilityTest.class.getSimpleName(),
-                    testName));
+                    testName), ex);
```
Azquelt, line 280: This change can stay.

Azquelt (changes requested):
See comments above.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/visibility/retry/RetryVisibilityTest.java
```diff
@@ -264,19 +263,19 @@ private void checkServiceCall(int nbExpectedRetries, RetryService service, Strin
...
+        } catch (TestException re) {
             Assert.assertEquals(
                     service.getNumberOfServiceCalls(),
                     expectedNbCalls,
                     String.format("in %s#%s service() should have been called exactly %d times",
                             RetryVisibilityTest.class.getSimpleName(),
                             testName,
                             expectedNbCalls));
-        } catch (RuntimeException ex) {
+        } catch (TestException ex) {
```
Azquelt, line 274: You've changed this test to use `TestException` rather than `IOException` which is fine, but this second catch block still needs to change to catch either `Exception` or `RuntimeException`, since this catch block is dealing with _unexpected_ exceptions.

Azquelt: I'm actually surprised that having two catch blocks for the same exception type isn't a compile error.

Azquelt:

Azquelt:

Azquelt:

PR #626, opened 2024-04-18
Remove broken Java 8 GH action build

Azquelt:
The newest version of maven does not work with Java 8. The jenkins build uses an older version of maven, but github actions have no intention of allowing the confiugration of the maven version actions/setup-java#457


Azquelt:
I would like to merge this because having the Java 8 build always failing means that no other github actions checks run for pull requests.

The Jenkins build will still cover building with Java 8.

PR #633, opened 2024-06-26
Make config disabled tests use `AsyncTaskManager`

### LINKED ISSUE #634: Config Disablement TCK tests use unmanaged threads
**Describe the bug**

All of the config disablement tests (e.g. `DisableAnnotationGloballyEnableOnClassTest`) use `Executors.newFixedThreadPool` to create threads to test whether the bulkhead is enabled.

It hasn't been a problem before, but when running in JakartaEE, some things may break when unmanaged threads like these are used so we should avoid them.

Azquelt:
Previously these tests were using a fixed thread pool executor which will create unmanaged threads on a Jakarta EE server.

We already have `AsyncCaller` which is designed to avoid this, and `AsyncTaskManager` which provides additional framework for exactly these type of tests so we should use them here.

Unfortunately these test classes are also testing the disablement of the `@Asynchronous` annotation which `AsyncCaller` and `AsyncTaskManager` use. We could split these tests up, but it's easier to add additional config to enable the annotation for that class specifically.

Fixes #634

### DIFF
--- a/spec/src/main/asciidoc/release_notes.asciidoc
+++ b/spec/src/main/asciidoc/release_notes.asciidoc
@@ -32,6 +32,9 @@ No.
 === Specification changes
 - Work with MicroProfile Telemetry Metrics (link:https://github.com/eclipse/microprofile-fault-tolerance/issues/622[#622])
 
+=== Other Changes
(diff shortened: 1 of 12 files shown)


Emily-Jiang (approved):

Emily-Jiang:
@Azquelt can you also add this to the release note in the spec doc?

Azquelt:
I've updated the release notes.

PR #637, opened 2024-07-04
Final tidyup

Emily-Jiang:
Specify the dependency on MP Telemetry instead of directly on OTel APIs

### DIFF
--- a/tck/pom.xml
+++ b/tck/pom.xml
@@ -31,6 +31,7 @@
         <checkstyle.methodNameFormat>^_?[a-z][a-zA-Z0-9_]*$</checkstyle.methodNameFormat>
         <microprofile-config-api.version>3.1</microprofile-config-api.version>
         <microprofile-metrics-api.version>4.0</microprofile-metrics-api.version>
+        <microprofile-telemetry-api.version>2.0-RC2</microprofile-telemetry-api.version>
     </properties>
 
     <dependencyManagement>
@@ -65,11 +66,7 @@
             <scope>provided</scope>
         </dependency>
 
-        <dependency>
-            <groupId>io.opentelemetry</groupId>
-            <artifactId>opentelemetry-api</artifactId>
-            <version>1.39.0</version>
-        </dependency>
+        
 
         <dependency>
             <groupId>jakarta.enterprise</groupId>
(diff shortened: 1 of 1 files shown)


tck/pom.xml
```diff
@@ -92,11 +89,13 @@
         </dependency>
 
         <dependency>
-            <groupId>io.opentelemetry</groupId>
-            <artifactId>opentelemetry-sdk-metrics</artifactId>
-            <version>1.39.0</version>
```
Azquelt, line 97: Is this dependency really not needed?

I'm fairly sure it's not a transitive dependency of `microprofile-telemetry-api`, but I guess it may be a dependency of `opentelemetry-sdk-extension-autoconfigure`?

Emily-Jiang: The tests compile okay without it.

Azquelt: Looks like we get `opentelemetry-sdk-extension-autoconfigure` -> `opentelemetry-sdk` -> `opentelemetry-sdk-metrics` :+1:

benjamin-confino:
I ran this locally and this change caused no issues, approving.

benjamin-confino (approved):

rhusar:
> I ran this locally and this change caused no issues, approving.

Sounds like you have `microprofile-telemetry-api:pom:2.0` installed manually? Becuase it's not publicly available on central yet. Are we waiting for this to be synced?

```
[ERROR] Failed to execute goal on project microprofile-fault-tolerance-tck: Could not resolve dependencies for project org.eclipse.microprofile.fault-tolerance:microprofile-fault-tolerance-tck:jar:4.2-SNAPSHOT
[ERROR] dependency: org.eclipse.microprofile.telemetry:microprofile-telemetry-api:pom:2.0 (provided)
```

Also

Azquelt:
@rhusar Yes, that's correct. `microprofile-telemetry-api:2.0` has been staged in preparation for release. You can either build it locally from the 2.0 tag, or temporarily add the staging repository to the pom (https://oss.sonatype.org/content/repositories/staging/).

Be aware that if you take either of these options, you'll want to manually remove the staged artifacts from your local maven repository in case they change as maven may not re-download them if it thinks you already have them locally.

The safer option for testing changes locally at the moment is to temporarily change the dependency on telemetry back to the RC version available on maven central.

Issue #639, opened 2024-07-24
Problem with metric types and units


donbourne:
Apologies for the late issue, but I noticed problems with the FaultTolerance OpenTelemetry metrics spec (which I had previously looked at, but which I now realize was missing important criteria in my review).

The following MP FT metrics should switch from being OTel Gauges to being UpDownCounters (either `LongUpDownCounter` or `ObservableLongUpDownCounter`), the determining factor in choosing between Gauges and UpDownCounters is whether the metric in question is [addititve](https://opentelemetry.io/docs/specs/otel/metrics/supplementary-guidelines/#:~:text=If%20the%20measurement%20values%20are,increasing%20%2D%20use%20an%20Asynchronous%20UpDownCounter).  Looking at the metrics currently listed as using Gauges in the MP FT spec:
- `ft.bulkhead.executionsRunning` - this metric is additive (I can sum the value across all methods and the result would have meaning)
- `ft.bulkhead.executionsWaiting` - same. this seems additive.

I would also recommend that the `Histograms` be changed to `DoubleHistograms` with units of `(s)` (instead of nanoseconds).  Rationale is to match general guidance from prometheus to use seconds for durations, and precedent from [HTTP semantic convention](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/) which times HTTP requests in seconds. This applies to:
- `ft.timeout.executionDuration`
- `ft.bulkhead.runningDuration`
- `ft.bulkhead.waitingDuration`

Azquelt:
I think both of these make sense and I can see how we ended up here.

1. MP Metrics has neither an up-down counter or a Histogram that can process `double`
2. MP Metrics automatically converts to the standard units when outputting in prometheus format, so with that API using nanoseconds as the unit made sense
3. When we added support for Open Telemetry, we mostly did a like-for-like conversion of the metrics
    - We did catch that `ft.circuitbreaker.state.total` could now be an `ObservableLongCounter` rather than a `Gauge`, but I wasn't aware of the additive property being a difference between an up-down counter and a gauge.

donbourne:
opened PR to address spec change #640

donbourne:
[...]

PR #641, opened 2024-07-27
Update TCK tests for OTel metrics types and units

Azquelt:
For #639, matching the spec changes proposed in #640


tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/telemetryMetrics/util/InMemoryMetricReader.java
```diff
@@ -71,61 +87,139 @@ public CompletableResultCode shutdown() {
...
+     *
+     * @param attributes
+     *            the attributes to check for
+     * @return a predicate which returns {@code true} if {@link PointData#getAttributes()} returns a superset of
+     *         {@code attributes}, otherwise {@code false}
+     */
+    private static Predicate<PointData> hasAllAttributes(Attributes attributes) {
+        return pointData -> {
+            for (Entry<AttributeKey<?>, Object> e : attributes.asMap().entrySet()) {
+                if (!pointData.getAttributes().get(e.getKey()).equals(e.getValue())) {
```
benjamin-confino, line 190: This will NPE if `e.getKey()` isn't in the attributes.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/telemetryMetrics/util/TelemetryMetricID.java
```diff
@@ -30,59 +30,48 @@ public class TelemetryMetricID {
...
-        int namesHash = Arrays.deepHashCode(names);
-        int valuesHash = Arrays.deepHashCode(value);
-
-        int hash = 17;
-        hash = hash * 31 + namesHash;
-        hash = hash * 31 + valuesHash;
-        hash = hash * 31 + name.hashCode();
-
-        return hash;
+        return Objects.hash(attributes, name, type);
```
benjamin-confino, line 74: I wasn't aware of this bit of syntactic sweetness. Thanks for making me aware of it :)

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/telemetryMetrics/util/InMemoryMetricReader.java
Azquelt: Good point, thank you

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/telemetryMetrics/util/TelemetryMetricID.java
Azquelt: I'll be honest, Eclipse's "Generate hashCode and equals" that made me aware of it.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/telemetryMetrics/util/TelemetryHistogramMetric.java
```diff
@@ -32,31 +42,126 @@
...
+     * @param expectedResultsMillis
+     *            the list of values which the histogram is expected to have recorded
+     */
+    public void assertBucketCounts(long... expectedResultsMillis) {
+        HistogramPointData pointData =
+                getHistogramPoint().orElseThrow(() -> new AssertionError("No data point for " + metricId.name));
+
+        // Check we have the right number of results in the histogram:
+        assertEquals(pointData.getCount(), expectedResultsMillis.length,
+                "Wrong number of results in histogram getCounts for " + metricId.name);
```
benjamin-confino, line 97: Please can we have metricId.toString() for these three asserts.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/telemetryMetrics/util/InMemoryMetricReader.java
Azquelt: Pushed a commit to fix this.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/telemetryMetrics/util/TelemetryHistogramMetric.java
```diff
@@ -32,31 +42,126 @@
...
+     * Assert that the bucket counts of the histogram are consistent with the expected set of recorded results.
+     * <p>
+     * The results are adjusted using {@link TCKConfig#getTimeoutInMillis(long)}.
+     * <p>
+     * A 20% deviation from the expected time is permitted and the upper limit is extended to 200ms.
+     *
+     * @param expectedResultsMillis
+     *            the list of values which the histogram is expected to have recorded
+     */
+    public void assertBucketCounts(long... expectedResultsMillis) {
```
benjamin-confino, line 91: This is a really nice method!

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/telemetryMetrics/BulkheadTelemetryTest.java
```diff
@@ -206,24 +203,24 @@ public void bulkheadMetricHistogramTest() throws InterruptedException, Execution
...
 
-        long count = executionTimesPoints.stream()
-                .mapToLong(points -> points.getCount())
-                .sum();
+        long count = executionTimesPoint
+                .map(HistogramPointData::getCount)
+                .orElse(0L);
 
-        assertThat("mean", Math.round(time / count), approxMillis(1000)); // histogram
+        assertThat("mean", time / count, approxMillisFromSeconds(1000)); // histogram
```
benjamin-confino, line 215: This could cause a divide by zero exception if `ft.bulkhead.runningDuration` isn't exported. Not the biggest issue since that's a fail, and it didn't cause problems in the old code, but since we have an Optional anyway we could assert isPresent for a nicer error message.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/telemetryMetrics/BulkheadTelemetryTest.java
```diff
@@ -206,24 +203,24 @@ public void bulkheadMetricHistogramTest() throws InterruptedException, Execution
...
 
         // Now let's put some quick results through the bulkhead
         bulkheadBean.waitForHistogram(CompletableFuture.completedFuture(null));
         bulkheadBean.waitForHistogram(CompletableFuture.completedFuture(null));
 
         // Should have 4 results, ~0ms * 2 and ~1000ms * 2
-        executionTimesCount = m.getBulkheadRunningDuration().getHistogramCount().get();
-        assertThat("histogram count", executionTimesCount, is(4L));
+        m.getBulkheadRunningDuration().assertBucketCounts(1000, 1000, 0, 0);
```
benjamin-confino, line 222: I wonder if we should go back to the old metric test and also make that use buckets? Not urgent of course.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/telemetryMetrics/util/TelemetryHistogramMetric.java
```diff
@@ -32,31 +42,126 @@
...
+                .map(HistogramPointData::getCount);
+    }
+
+    public boolean isPresent() {
+        return getHistogramPoint().isPresent();
+    }
+
+    public void assertBoundaries() {
+        HistogramPointData pointData =
+                getHistogramPoint().orElseThrow(() -> new AssertionError("No data point for " + metricId.name));
```
benjamin-confino, line 73: Can this output metricId.toString() please.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/telemetryMetrics/util/TelemetryMetricID.java
Azquelt: I should also note that this is only valid because implementations of `Attributes` are required to implement `equals` and `hashCode` correctly, otherwise the code that you had which was doing its own hashing of `attributes` would have been necessary.

Often interfaces assert this by redeclaring `equals` and `hashCode` (e.g. `java.util.Map` does this), but in the case of `Attributes`, it's only stated [in the class description](https://www.javadoc.io/doc/io.opentelemetry/opentelemetry-api/1.39.0/io/opentelemetry/api/common/Attributes.html).

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/telemetryMetrics/BulkheadTelemetryTest.java
Azquelt: Good point, we should stop the test if there's no data for the metric.

However, I think we still need to assert `count > 0` before computing the mean, since the metric could be present but still have no data in it.

Azquelt: Hmm, in fact we're already asserting `count == 2` above but just getting it in a different way... let's tidy this up.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/telemetryMetrics/BulkheadTelemetryTest.java
Azquelt: That would be a good idea in the future, but at the moment the TCK compiles against MP Metrics 4.0 which doesn't have explicit buckets, it just has percentiles (which we can't reliably test, see message on 28d305).

If we set a minimum version of MP Metrics in a future release, we could add an assertion like this to `BulkheadMetricTest`.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/telemetryMetrics/BulkheadTelemetryTest.java
Azquelt: Pushed a commit to tidy up this method.

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/telemetryMetrics/util/TelemetryHistogramMetric.java
Azquelt: Done

tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/telemetryMetrics/util/TelemetryHistogramMetric.java
Azquelt: Done

Azquelt:
- Squashed review changes
- Reverted the dependency on MP Telemetry to RC2, so that the build checks will pass and we can release a new FT RC
- rebased on `main`

PR #649, opened 2024-09-11
Avoid using test groups for test ordering

Ladicek:
Test groups are global, not just for single test. This means that multiple test classes may be "in progress" at the same time, which leads to issues in at least one implementation which has global state.

Instead of test groups and `dependsOnGroups`, this commit uses simple `dependsOnMethods`, which is local to the test class.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/telemetryMetrics/BulkheadTelemetryTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/telemetryMetrics/BulkheadTelemetryTest.java
@@ -112,7 +112,7 @@ public class BulkheadTelemetryTest extends Arquillian {
         return result;
     }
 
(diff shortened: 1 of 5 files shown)


Ladicek:
I'd love this to be released in 4.1.1, which should be OK I think?

Azquelt:
I can confirm that I can see the same behaviour running on OpenLiberty - the chronological results view shows the `testMetricUnits` from each class running at the end after all the others. The relevant test applications remain deployed while other tests are running.

Azquelt:

Ladicek:
Bump, any updates here?

Emily-Jiang (approved):
LGTM

Ladicek:
Thanks!

Ladicek:
Can we please get a MP FT 4.1.1 release with this? The `main` branch doesn't contain anything else, so I think that should be safe? Or is there an organizational issue?

PR #655, opened 2025-03-05
baseline metrics in CircuitBreakerTelemetryTest

### LINKED ISSUE #656: TCK Challenge: CircuitBreakerTelemetryTest may fail because metrics are not baselined correctly
The `CircuitBreakerMetricTest` does this:

```java
MetricGetter m = new MetricGetter(CircuitBreakerMetricBean.class, "doWork");
m.baselineMetrics();
```

and it works correctly.

The `CircuitBreakerTelemetryTest` does just

```java
TelemetryMetricGetter m = new TelemetryMetricGetter(CircuitBreakerMetricBean.class, "doWork");
```

and while it may work correctly in some cases, it may easily fail due to the `@BeforeTest` method running requests against the circuit breaker. This is what `baselineMetrics()` is supposed to fix, but it has to be used. The `CircuitBreakerTelemetryTest` doesn't.

Ladicek:
The call to `baselineMetrics()` after creating the `TelemetryMetricGetter` was missing in this test. Funnily enough, the MP Metrics counterpart of this test (`CircuitBreakerMetricTest`) baselines metrics correctly.

Fixes #656

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/CircuitBreakerMetricTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/metrics/CircuitBreakerMetricTest.java
@@ -44,7 +44,7 @@ import org.jboss.shrinkwrap.api.ShrinkWrap;
 import org.jboss.shrinkwrap.api.asset.EmptyAsset;
 import org.jboss.shrinkwrap.api.spec.JavaArchive;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
-import org.testng.annotations.BeforeTest;
+import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
 import jakarta.inject.Inject;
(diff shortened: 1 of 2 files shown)


Emily-Jiang (approved):

Azquelt:

Azquelt (approved):

Ladicek:
Added an extra commit that replaces `@BeforeTest` with `@BeforeMethod`, which I feel is a lot more appropriate (after, once again, reading upon the difference).

Azquelt (approved):

Azquelt:
Thanks, I think this makes sense. I agree the cleanup in this method probably isn't necessary since the test class only has one test method, but the current behaviour (where the `@BeforeTest` method seems to run for you but doesn't run for me on Open Liberty) is somewhat disconcerting so I think it makes sense to change it to `@BeforeMethod` (which was the intended behaviour all along).

Azquelt:
I'll merge this on Monday unless any objections are raised and then we can do a service release.

brunobat (approved):

